PR #4, opened 2019-09-24
Container jfr controller

andrewazores:
WIP for implementation of controller for container-jfr services within cluster


andrewazores:
This is now working at a minimal baseline level and I think is worth merging. From there we can figure out what other enhancements are needed.

Makefile
```diff
@@ -20,15 +20,31 @@ deploy: undeploy
...
+	- oc delete deployment exposecontroller
+	- oc delete configmap exposecontroller
+	- oc delete all -l project=exposecontroller
+	- oc delete routes -l generator=exposecontroller
 	- oc delete deployment container-jfr-operator
 	- oc delete containerjfr --all
 	- oc delete flightrecorder --all
 	- oc delete all -l name=container-jfr-operator
+	- oc delete all -l app=containerjfr
+	- oc delete all -l app=containerjfr-pod
```
ebaron: I can't seem to find where this label is used, I think the pod is also under the `app=containerjfr` label.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -0,0 +1,276 @@
...
+				"app": "containerjfr",
+			},
+			Annotations: map[string]string{
+				"fabric8.io/expose": "true",
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Type: corev1.ServiceTypeClusterIP,
+			Selector: map[string]string{
+				"app": "containerjfr",
```
ebaron: Perhaps to make this more flexible, we should use the name of the ContainerJFR object, like the pod does for its app label.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -0,0 +1,276 @@
...
```
ebaron: Same as above.

Makefile
```diff
@@ -20,15 +20,31 @@ deploy: undeploy
...
+	- oc delete configmaps -l app=containerjfr
 	- oc delete role container-jfr-operator
 	- oc delete rolebinding container-jfr-operator
 	- oc delete serviceaccount container-jfr-operator
 	- oc delete crd flightrecorders.rhjmc.redhat.com
+	- oc delete crd containerjfrs.rhjmc.redhat.com
+	- oc adm policy remove-cluster-role-from-user cluster-admin system:serviceaccount:default:exposecontroller
+	- oc adm policy remove-cluster-role-from-group cluster-reader system:serviceaccounts
```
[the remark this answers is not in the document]
andrewazores: Ah, no. The quickstart for exposecontroller suggested to add those (very generous) policies to allow the exposecontroller to work, but after narrowing the granted roles and policies I forgot to remove those from the undeploy step there.

deploy/exposecontroller_config_map.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    provider: fabric8
+    project: exposecontroller
+  name: exposecontroller
+data:
+  config.yml: |-
+    watch-namespaces: myproject
```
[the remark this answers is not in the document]
andrewazores: Yes, I'll remove it here. I couldn't get the config here to actually take effect in the controller for some reason, but setting it via an arg in the deployment works just as well (for now anyway).

Makefile
andrewazores: Good call, probably a holdover from some in-between iteration.

pkg/controller/containerjfr/containerjfr_controller.go
andrewazores: Done.

pkg/controller/containerjfr/containerjfr_controller.go
andrewazores: Done.

PR #6, opened 2019-10-03
Initial FlightRecorder API, Service controller

ebaron:
This PR adds some fields to the FlightRecorder spec/status resource. This custom resource mirrors compatible services (and potentially pods) on a 1-to-1 basis. The spec contains a toggle to request recording to start/stop for the target service. The status contains the current recording status, an object reference to the target service, and a list of JFR recordings for that service. 

The target field currently has the same reference as the ownerReference. One advantage of having this information in the status subresource, is that we can easily prevent the user from modifying it by restricting PUT access to `flightrecorders/status`. Modifications to the object as a whole do not also update the status fields [1].

In addition to this API, there is a new Service controller that watches for new services and if they are compatible with Container JFR (currently checking if they listen on port 9091), the controller creates a FlightRecorder object for that service. The service is listed as the owner of the FlightRecorder object, so the garbage collector will also delete the FlightRecorder when the service is deleted.

After this, I plan to fill in the FlightRecorder controller to watch for changes to FlightRecorder objects and act accordingly.

Example object:
```json
{
    "apiVersion": "rhjmc.redhat.com/v1alpha1",
    "kind": "FlightRecorder",
    "metadata": {
        "creationTimestamp": "2019-10-03T18:51:02Z",
        "generation": 1,
        "labels": {
            "app": "hot-methods"
        },
        "name": "hot-methods",
        "namespace": "myproject",
        "ownerReferences": [
            {
                "apiVersion": "v1",
                "blockOwnerDeletion": true,
                "controller": true,
                "kind": "Service",
                "name": "hot-methods",
                "uid": "a2f4f499-e0a9-11e9-8b59-52540051bad9"
            }
        ],
        "resourceVersion": "138243",
        "selfLink": "/apis/rhjmc.redhat.com/v1alpha1/namespaces/myproject/flightrecorders/hot-methods",
        "uid": "bf3167cd-e60e-11e9-9d97-52540051bad9"
    },
    "spec": {
        "recordingActive": false
    },
    "status": {
        "recordingActive": false,
        "recordings": [],
        "target": {
            "apiVersion": "v1",
            "kind": "Service",
            "name": "hot-methods",
            "namespace": "myproject",
            "resourceVersion": "2971",
            "uid": "a2f4f499-e0a9-11e9-8b59-52540051bad9"
        }
    }
}
```
[1] https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource


pkg/controller/service/service_controller.go
```diff
@@ -0,0 +1,178 @@
+package service
+
+import (
+	"context"
+
+	cjfrapi "github.com/rh-jmc-team/container-jfr-operator/pkg/apis/rhjmc/v1alpha1"
```
lberk: why `cjfrapi` instead of `v1alpha1` ?  By referencing the apis by explicitly by version, you can make sure your controller/reconciliation can upgrade versions smoothly (by changing to `v1beta1` and `v1` explicitly and adjusting to the change in struct types).  Leaving it as `cjfrapi` means, internally, you'll have to add a variety of conditional logic based on the api version you're referencing, which can get unwieldy.

pkg/controller/service/service_controller.go
```diff
@@ -0,0 +1,178 @@
...
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a Service object and makes changes based on the state read
+// and what is in the Service.Spec
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileService) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
lberk: it might make sense to set the logging config you want in the controller itself, and the pass via context so you'd be able to share the same logger settings everywhere and set them via something akin to `logger := logging.FromContext(ctx).Infof(...)`

pkg/controller/service/service_controller.go
ebaron: Sounds good. I have renamed the import to rhjmcv1alpha1, which is already used elsewhere.

pkg/controller/service/service_controller.go
ebaron: I'm a bit conflicted here. The doc for `context` says it shouldn't be stored in a struct and to pass contexts as arguments to each function that needs it [1], but Reconcile is an interface method, so I can't add a context parameter there. What do you think?

[1] https://godoc.org/context

andrewazores:
@ebaron do you intend to add any more commits to this set? If not, do you mind if I squash before merging?

ebaron:
Nope, this is good to go.

PR #8, opened 2019-10-16
Clean up exposecontroller configuration

andrewazores:
Configure exposecontroller to watch services only in current namespace
as its own serviceaccount, removing setup-time configuration for which
namespace will be watched. This is preparation for bundling the operator
for operator catalogs, so that yaml definitions do not contain any
"REPLACE_IMAGE" or "REPLACE_PROJECT" type variables to be run through
sed.


deploy/operator.yaml
```diff
@@ -16,8 +16,7 @@ spec:
       serviceAccountName: container-jfr-operator
       containers:
         - name: container-jfr-operator
-          # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: quay.io/rh-jmc-team/container-jfr:0.4.7
```
ebaron: Should this be container-jfr-operator?

pkg/controller/containerjfr/resource_definitions/resource_definitions.go
```diff
@@ -63,7 +63,7 @@ func NewPodForCR(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Pod {
 func NewCoreContainer(cr *rhjmcv1alpha1.ContainerJFR) corev1.Container {
 	return corev1.Container{
 		Name:  cr.Name,
-		Image: "quay.io/rh-jmc-team/container-jfr:0.4.6-debug",
+		Image: "quay.io/andrewazores/container-jfr:latest",
```
ebaron: Was this change intentional, or just for local testing?

andrewazores: Local testing, fixed.

deploy/operator.yaml
andrewazores: Whoops, yes. Thanks for the catch!

PR #10, opened 2019-10-25
Replace exposecontroller with direct Routes

andrewazores:
Remove exposecontroller in favour of directly generated OpenShift
Routes. This potentially removes compatibility with other Kubernetes
distributions, but allows the Operator to rely on the OpenShift
configured Router setup much more robustly without any user intervention
to configure the exposecontroller to match the Router.

This also adds the "bundle" subdirectory, which contains some Operator
catalog bundle definitions so that the Operator can be deployed to a
cluster by graphically installing it through the OpenShift Console, and
eventually can be published to OperatorHub so that the custom Operator
Source is not required.


andrewazores:
Additional benefit of creating the Routes directly rather than using an external controller: the ingress URLs for these routes can be retained and used directly when creating the application pod to set the required environment variables. This is much quicker than relying on exposecontroller's configmap updates, since the pod only needs to be started once rather than restarted each time a service is exposed to the outside and its URL changes.

And of course, we drop a whole extra pod's worth of memory usage and a little bit of CPU time as well since we don't need to run exposecontroller. The permissions model is also *vastly* simplified compared to what I was working on to get exposecontroller running correctly in the new environment.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -72,8 +76,9 @@ var _ reconcile.Reconciler = &ReconcileContainerJFR{}
 type ReconcileContainerJFR struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client client.Client
-	scheme *runtime.Scheme
+	client      client.Client
+	routeClient routeClient.RouteV1Client
```
ebaron: I think we might be able to use the existing client by adding Route to our Scheme, but this is fine for now.

pkg/controller/containerjfr/resource_definitions/resource_definitions.go
```diff
@@ -30,42 +37,17 @@ func NewPersistentVolumeClaimForCR(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Persi
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      cr.Name + "-pod",
 			Namespace: cr.Namespace,
 			Labels: map[string]string{
-				"app": cr.Name,
+				"app":  cr.Name,
+				"kind": "containerjfr",
+			},
+			Annotations: map[string]string{
+				"redhat.com/containerJfrUrl": specs.CoreAddress,
```
ebaron: Just curious, what do you have in mind for this new label and annotation?

pkg/controller/containerjfr/containerjfr_controller.go
andrewazores: Good to know. I figured there must be a better way to do it, but I couldn't figure it out.

bundle/containerjfrs.rhjmc.redhat.com.crd.yaml
```diff
@@ -0,0 +1,42 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: containerjfrs.rhjmc.redhat.com
+spec:
+  group: rhjmc.redhat.com
+  names:
+    kind: ContainerJFR
+    listKind: ContainerJFRList
+    plural: containerjfrs
+    singular: containerjfr
```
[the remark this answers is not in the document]
andrewazores: Good idea, I'll commit something for this in a minute. They're not generated so I'll do something with copying them in in a make target.

pkg/controller/containerjfr/resource_definitions/resource_definitions.go
andrewazores: They're not particularly important, but I added them just so that the pod and the general URL for the frontend can be easily selected and discovered. Basically, the `oc get` line that I modified in the README makes use of them.

Makefile
```diff
@@ -10,31 +13,44 @@ compile:
...
 	oc create -f deploy/operator_role.yaml
-	oc create -f deploy/exposecontroller_role.yaml
 	oc create -f deploy/operator_role_binding.yaml
-	oc create -f deploy/exposecontroller_role_binding.yaml
-	oc create -f deploy/exposecontroller_role_binding_cluster_admin.yaml
-	oc create -f deploy/exposecontroller_role_binding_cluster_reader.yaml
-	oc create -f deploy/crds/rhjmc_v1alpha1_flightrecorder_crd.yaml
-	oc create -f deploy/crds/rhjmc_v1alpha1_containerjfr_crd.yaml
+	oc create -f deploy/crds/flightrecorders.rhjmc.redhat.com.crd.yaml
+	oc create -f deploy/crds/containerjfrs.rhjmc.redhat.com.crd.yaml
```
ebaron: After the rename of the CRD files, `operator-sdk generate openapi` does not seem to update them in response to changes in *_types.go. It might be best just copy each CRD manually, and leave them named as whatever the tooling expects.

Makefile
```diff
@@ -10,31 +13,44 @@ compile:
 image: compile
 	operator-sdk build $(IMAGE_TAG)
 
+.PHONY: test
+test:
+	@echo $(CRDS)
```
ebaron: Not sure if you meant to leave this target in.

pkg/controller/containerjfr/containerjfr_controller.go
ebaron: I think I have a rough idea how to do it. I'll play around with it.

Makefile
andrewazores: Gone.

PR #14, opened 2019-10-30
Modify scorecard config for sdk v0.10.1 compatibility

andrewazores:
The current config was written with an operator-sdk v0.11.0 client binary in use. This simpler config is compatible with the v0.10.x currently specified in the project go.mod.

### DIFF
--- a/.osdk-scorecard.yaml
+++ b/.osdk-scorecard.yaml
@@ -1,14 +1,3 @@
 ---
-scorecard:
-    output: text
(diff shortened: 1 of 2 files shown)


ebaron:
Hey Andrew, is this still desired after the Operator SDK upgrade?

andrewazores:
Nope, doesn't make sense anymore. Better to keep the 0.11.0 config.

PR #23, opened 2019-11-04
Update go.sum

andrewazores:
It seems like maybe the released binary for operator-sdk v0.11.0 was changed after we did our update, since the checksum seems to no longer match. Elliott, could you verify that you see this too?


ebaron:
Weird, I'm not seeing any issues when I try to build. Any suggestions on how to reproduce it?

andrewazores:
Huh. Even just a `go mod tidy` fails for me because of a checksum mismatch.

andrewazores:
Okay, nevermind I guess. I nuked the operator-sdk contents in my $GOPATH and it's fine now, so I guess somehow I ended up with a bad copy in there...

PR #29, opened 2019-11-15
Add "Minimal" field to ContainerJFR spec

### LINKED ISSUE #27: Operator should have a minimal/"headless" container-jfr deployment option
See rh-jmc-team/container-jfr#68

The Operator should provide some way to deploy container-jfr in the minimal/headless image variant, and without Grafana/jfr-datasource deployed within the pod. This would minimize the application footprint within the cluster while still serving the requirements for starting/stopping/retrieving recordings via the Operator CRD APIs.

andrewazores:
If "Minimal" is set to true then container-jfr will be deployed with the
-minimal image variant, containing no web-client UI resources, and the
pod deployment will also not contain Grafana or jfr-datasource containers, services, or routes.

The Operator also creates/cleans up the additional services and routes if the Spec.Minimal field changes on an existing ContainerJFR. The Pod will be destroyed and recreated to match - it should be checked in the future that this can't be done more efficiently and less disruptively by updating the set of containers running within the Pod.

Fixes #27

### DIFF
--- a/pkg/controller/containerjfr/containerjfr_controller.go
+++ b/pkg/controller/containerjfr/containerjfr_controller.go
@@ -3,6 +3,7 @@ package containerjfr
 import (
 	"context"
 	"fmt"
+	"time"
 
 	openshiftv1 "github.com/openshift/api/route/v1"
 	routeClient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
@@ -24,11 +25,6 @@ import (
 )
 
 var log = logf.Log.WithName("controller_containerjfr")
-
-/**
-* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
-* business logic.  Delete these comments after modifying this file.*
- */
 
 // Add creates a new ContainerJFR Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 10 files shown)


pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -111,17 +109,67 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
...
+		serviceSpecs.DatasourceAddress = fmt.Sprintf("http://%s", url)
+
+		// check for existing minimal pod and delete if found
+		pod := resources.NewPodForCR(instance, serviceSpecs)
+		err = r.client.Get(context.Background(), types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, pod)
+		if err == nil && len(pod.Spec.Containers) == 1 {
+			reqLogger.Info("Deleting existing minimal pod")
+			err = r.client.Delete(context.Background(), pod)
+			if err != nil && !errors.IsNotFound(err) {
+				return reconcile.Result{Requeue: true, RequeueAfter: time.Second * 10}, err
```
ebaron: Just curious, what's the reason for overriding the defaults in Result here?

andrewazores: Basically that creating and deleting Pods tends to be a slow process compared to everything else, so if an unexpected error results when deleting it I figured it makes sense to add a bit of delay before requeuing the re-attempt, just in case it's some transient error and the pod might actually be in the process of deletion. Without a delay then there would likely be a lot of failed re-attempts here until the pod does eventually go away.

ebaron: Understood, thanks!

Issue #39, opened 2019-12-05
TLS edge termination should be enabled


andrewazores:
ContainerJFR has been updated to work behind SSL proxies since rh-jmc-team/container-jfr#82 / rh-jmc-team/container-jfr#83 . This should allow the Operator to configure TLS edge termination for the container-jfr routes

PR #41, opened 2019-12-05
Tls edge

andrewazores:
Addresses #39 . Enables TLS edge termination for container-jfr, command channel, and Grafana routes, and removes the external routing to the jfr-datasource service (so that it is inaccessible from the outside, which is more suitable anyway but also has the side effect of reducing the need for encrypted connections)

### DIFF
--- a/pkg/controller/containerjfr/containerjfr_controller.go
+++ b/pkg/controller/containerjfr/containerjfr_controller.go
@@ -116,7 +116,7 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-		serviceSpecs.GrafanaAddress = fmt.Sprintf("http://%s", url)
+		serviceSpecs.GrafanaAddress = fmt.Sprintf("https://%s", url)
 
 		datasourceSvc := resources.NewJfrDatasourceService(instance)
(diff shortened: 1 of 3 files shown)


pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -200,15 +201,28 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
...
+	if err := r.client.Get(context.Background(), types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}, svc); err != nil {
+		return "", err
+	}
+	if svc.Spec.ClusterIP == "" {
+		return "", errors.NewInternalError(goerrors.New(fmt.Sprintf("Expected service %s to have ClusterIP, but got empty string", svc.Name)))
+	}
+	if len(svc.Spec.Ports) != 1 {
+		return "", errors.NewInternalError(goerrors.New(fmt.Sprintf("Expected service %s to have one Port, but got %d", svc.Name, len(svc.Spec.Ports))))
+	}
+	return fmt.Sprintf("%s:%d", svc.Spec.ClusterIP, svc.Spec.Ports[0].TargetPort.IntVal), nil
```
ebaron: I think we might want `Ports[0].Port` instead of `TargetPort`, although in our case they should be the same number.

PR #42, opened 2019-12-09
Add liveness probes to all container definitions

### LINKED ISSUE #40: Services should have health checks
Liveness/readiness probes should be defined for containers

andrewazores:
Fixes #40

Readiness probes aren't added since none of the containers do anything particularly heavy at initialization time, so liveness is equivalent to readiness.

### DIFF
--- a/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
+++ b/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
@@ -151,6 +151,14 @@ func NewCoreContainer(cr *rhjmcv1alpha1.ContainerJFR, specs *ServiceSpecs) corev
 			},
 		},
 		Env: envs,
+		LivenessProbe: &corev1.Probe{
+			Handler: corev1.Handler{
+				HTTPGet: &corev1.HTTPGetAction{
+					Port: intstr.IntOrString{IntVal: 8181},
+					Path: "/clienturl",
+				},
+			},
+		},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


andrewazores:
> Looks good. If I understand the defaults correctly, this should allow for 30 seconds (3 attempts * 10s period) before killing the pod. That seems reasonable.
> 
> As an RFE, it would be nice if this also tested if the command server is ready. I think I've encountered cases where querying `clienturl` is successful, but I can't yet connect via WebSocket to the command server. I could be wrong though.

That's probably possible. I don't know exactly how the vertx server is handling that, but I know that the startup is async, though container-jfr does what it can to handle startup/shutdown synchronously. There's still probably some timing window where the handler for the static URLs like /clienturl is ready and the websocket handler is not. I did look at implementing a readiness probe on the websocket connection but I didn't really find a way to do that. Instead, maybe it makes sense to add a /health or something like that to container-jfr, and have it keep track of its own startup state including vertx status, so that the response there is based on a more complete service status?

ebaron:
> > Looks good. If I understand the defaults correctly, this should allow for 30 seconds (3 attempts * 10s period) before killing the pod. That seems reasonable.
> > As an RFE, it would be nice if this also tested if the command server is ready. I think I've encountered cases where querying `clienturl` is successful, but I can't yet connect via WebSocket to the command server. I could be wrong though.
> 
> That's probably possible. I don't know exactly how the vertx server is handling that, but I know that the startup is async, though container-jfr does what it can to handle startup/shutdown synchronously. There's still probably some timing window where the handler for the static URLs like /clienturl is ready and the websocket handler is not. I did look at implementing a readiness probe on the websocket connection but I didn't really find a way to do that. Instead, maybe it makes sense to add a /health or something like that to container-jfr, and have it keep track of its own startup state including vertx status, so that the response there is based on a more complete service status?

That would be ideal. It's not a big concern right now, more of a nice-to-have feature.

Issue #25, opened 2019-11-05
OpenShift 4 "topology" graphic shows operator, but no apparent link to ContainerJFR instance(s)


andrewazores:
(As Developer role)

![image](https://user-images.githubusercontent.com/3787464/68216034-3a5bd500-ffae-11e9-953c-051cee8e1361.png)

![image](https://user-images.githubusercontent.com/3787464/68216051-42b41000-ffae-11e9-8285-0c4758c27136.png)

After an Administrator installs the Operator, they can create a ContainerJFR CR which triggers the creation of a ContainerJFR Pod and its associated resources (Services+Routes, PersistentVolumeClaim). The exposed URL for one of these Routes is the ContainerJFR application URL.

There seems to be no easy way for a Developer user to find this application URL from their view within the OpenShift Console. They can see the Operator in the workspace topology, but cannot see the ContainerJFR Pod or its associated resources.

Deploying some other application (ex. `quay.io/andrewazores/container-jmx-docker-listener`) into the workspace produces a nice node in the topology graph, with a link to the application URL.

![image](https://user-images.githubusercontent.com/3787464/68216569-10ef7900-ffaf-11e9-93a6-af44e038ed23.png)

![image](https://user-images.githubusercontent.com/3787464/68216595-1a78e100-ffaf-11e9-8308-3605f988791f.png)

andrewazores:
https://docs.openshift.com/container-platform/4.2/applications/application-life-cycle-management/odc-viewing-application-composition-using-topology-view.html#odc-labels-and-annotations-used-for-topology-view_viewing-application-composition-using-topology-view

PR #43, opened 2019-12-09
Topology view

andrewazores:
See #25

Screenshot_2019-12-09 Topology · Red Hat OpenShift Container Platform

With this PR, the container-jfr Pod is placed into a Deployment, which allows it to be visible on the Topology view for developer accounts. Labels/annotations are also added to this Pod, which provide the arrow linking the container-jfr instance back up to its Operator parent. I spent some time trying to get the application grouping working, but this requires a corresponding label to be present on the Operator Deployment, which it does not seem is currently possible (https://godoc.org/github.com/operator-framework/operator-lifecycle-manager/pkg/controller/install contains no information about attaching labels or other metadata to the Deployment itself, only to its owned Pods). I did some manual testing and everything seems to still deploy and run correctly, both with the Makefile and via OperatorHub (with custom source). Scorecard testing is also unimpacted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,14 @@ clean-bundle:
 # "Local" (ex. minishift/crc) testing targets #
 #########################################
 
+.PHONY: catalog
+catalog: remove_catalog
+	oc create -f bundle/openshift/operator-source.yaml
+
+.PHONY: remove_catalog
+remove_catalog:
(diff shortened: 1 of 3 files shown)


andrewazores:
Also, the little "open link" icon off of the container-jfr deployment opens the container-jfr web UI. This seems to be automatically picked up by the first defined route, or some other automatic mechanism.

PR #44, opened 2019-12-13
FlightRecorder API for managing recordings

ebaron:
This PR is a more complete replacement for #24. This adds new fields to the FlightRecorder CRD to allow creating, listing and downloading of recordings. The controller has been updated to reconcile these new fields with Container JFR. Communication with Container JFR is done using a WebSocket connection. The WebSocket client and its Container JFR-specific code can be found in the new `client` package.

### Creating a new recording
The user adds a new item to `recordingRequests` under the FlightRecorder spec. This request includes the name of the new recording, the duration as a `time.Duration` string, and a string array of event options. These event options correspond to the syntax used by Container JFR. When the reconciler processes the updated spec, it will forward this request to Container JFR. If successful, it will remove the request from the spec and the recording will be listed under the FlightRecorder's status.

### Listing and downloading recordings
After creating a new recording and while there are any still-running recordings for a given FlightRecorder, the reconciler will periodically update the status of these recordings with information from Container JFR. This information is found in the `recordings` array in the FlightRecorder's status. Each recording lists its name, start time (as an RFC3339 string), duration (as a time.Duration string), whether the recording is still active (non-stopped), and a download URL for the JFR file. In order to allow this JFR file to be downloaded while Container JFR isn't connected to its associated JVM, the reconciler automatically saves recordings as they finish to persistent storage. We can likely improve upon this in the future.

Example FlightRecorder CR:
```json
{
    "apiVersion": "rhjmc.redhat.com/v1alpha1",
    "kind": "FlightRecorder",
    "metadata": {
        "creationTimestamp": "2019-12-12T23:03:30Z",
        "generation": 1,
        "labels": {
            "app": "my-service"
        },
        "name": "my-service",
        "namespace": "default",
        "ownerReferences": [
            {
                "apiVersion": "v1",
                "blockOwnerDeletion": true,
                "controller": true,
                "kind": "Service",
                "name": "my-service",
                "uid": "3a378b6e-10a0-11ea-9566-fa163e644a90"
            }
        ],
        "resourceVersion": "305867",
        "selfLink": "/apis/rhjmc.redhat.com/v1alpha1/namespaces/default/flightrecorders/my-service",
        "uid": "f6c043c3-10a0-11ea-9566-fa163e644a90"
    },
    "spec": {
        "recordingRequests": [
	    {
		"name": "new-recording",
		"duration": "2m",
		"eventOptions": [
		    "jdk.ExecutionSample:enabled=true",
		    "jdk.ExecutionSample:period=200ms"
		]
	    }
	]
    },
    "status": {
        "recordings": [
            {
                "active": false,
                "downloadUrl": "http://containerjfr.example.com:80/recordings/172-30-30-231_my-recording_20191212T230321Z.jfr",
                "duration": "5s",
                "name": "my-recording",
                "startTime": "2019-12-12T23:03:16Z"
            }
        ],
        "target": {
            "apiVersion": "v1",
            "kind": "Service",
            "name": "my-service",
            "namespace": "default",
            "resourceVersion": "302913",
            "uid": "3a378b6e-10a0-11ea-9566-fa163e644a90"
        }
    }
}
```


andrewazores:
The example CR looks great and the description makes good sense. Hang on while I review the implementation.

PR #45, opened 2019-12-24
Auth

andrewazores:
This ~~WIP~~ PR brings the operator up to speed with the auth changes in container-jfr since https://github.com/rh-jmc-team/container-jfr/pull/86 .

The operator now includes its serviceaccount token on the command channel connection that the flightrecorder controller uses. Without this the updated container-jfr container will reject the websocket handshake, since the user issuing the request cannot be authenticated and the request authorized.

~~WIP because there's still a TODO regarding certificate verification for TLS on the websocket connection. In CRC at least, the certificate used by the cluster is somehow invalid (probably self-signed, haven't checked the exact reason yet), so the operator's websocket client refuses the connection on its end before getting to the handshake rejection due to lack of authz. The PR currently unconditionally disables cert verification.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,8 @@ undeploy: undeploy_sample_app
 
 .PHONY: sample_app
 sample_app:
-	oc new-app andrewazores/container-jmx-docker-listener:latest --name=jmx-listener
+	oc new-app quay.io/andrewazores/vertx-fib-demo
 
 .PHONY: undeploy_sample_app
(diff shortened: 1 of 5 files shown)


andrewazores:
A `TLS_VERIFY` env var is now used to control whether certs are verified. If that env var is set to "false" then the operator will skip certificate verification when configuring the Grafana instance and when connecting to the container-jfr command channel websocket.

PR #46, opened 2020-01-07
Podman image builder

andrewazores:
~~This is based on top of the currently open Auth PR #45 . Only the last commit is new for this branch.~~

This simply adds the image-builder arg to the operator-sdk invocation, with `podman` specified as the image builder rather than the default Docker. This is motivated by `podman` requiring less setup, being rootless/daemonless, and along that same line, with Docker being incompatible with cgroupsv2 and therefore not currently working by default in Fedora 31. Since I'm using Fedora 31 this is a convenience win for me, but it also seems like a net benefit in general.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ openapi:
 
 .PHONY: image
 image: generate
-	operator-sdk build $(IMAGE_TAG)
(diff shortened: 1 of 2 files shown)


andrewazores:
Rebased.

PR #47, opened 2020-01-07
Target jmx named port

andrewazores:
This is based on top of the currently open ~~Auth #45 and~~ podman builder #46 PRs.

The last ~~two~~ three commits in this branch replace the hardcoded exporter port number 8181 and JMX port number 9091 with named ports.

For the JMX target port of 9091, this is now used as a fallback rather than a hardcoded default. The service_controller still watches for new services to appear and for each new service it checks all of the defined service ports. If a port named "jmx" is found then the service is marked as a compatible target. If no "jmx" port is found but the 9091 port is found then it is also marked as compatible. If neither are found then the service is ignored as a potential target. Then, after the service_controller creates the FlightRecorder resource including this port number, the flightrecorder_controller (watching for new FlightRecorder CRs) reconcile is triggered, which uses this port number for establishing the target connection via ContainerJFR.

~~This part is still WIP since the default 9091 fallback port number constant is hardcoded in both controllers. I intend to resolve this by adding the port number to the FlightRecorder CRD, so that the service_controller can include the port number in the CR upon creation, and this way pass the port number to the flightrecorder_controller and avoid the duplicate fallback logic.~~

For the exporter 8181 port, the flightrecorder_controller assumes that the owned ContainerJFR instance will have a port named "export", which is simply the same existing port definition on 8181 as before, but renamed from "8181-tcp" to "export". This allows the controller to connect to the ContainerJFR instance even if the webserver port is changed in the future without requiring an update to the controller code.

### DIFF
--- a/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
+++ b/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
@@ -254,12 +254,12 @@ func NewExporterService(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Service {
 			},
 			Ports: []corev1.ServicePort{
 				{
-					Name:       "8181-tcp",
+					Name:       "export",
 					Port:       8181,
 					TargetPort: intstr.IntOrString{IntVal: 8181},
 				},
 				{
-					Name:       "9091-tcp",
+					Name:       "jmx",
 					Port:       9091,
 					TargetPort: intstr.IntOrString{IntVal: 9091},
 				},
@@ -285,7 +285,7 @@ func NewCommandChannelService(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Service {
 			},
 			Ports: []corev1.ServicePort{
 				{
-					Name:       "9090-tcp",
+					Name:       "cmdchan",
 					Port:       9090,
 					TargetPort: intstr.IntOrString{IntVal: 9090},
 				},
--- a/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
+++ b/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
@@ -259,7 +259,7 @@ func NewExporterService(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Service {
 					TargetPort: intstr.IntOrString{IntVal: 8181},
 				},
 				{
-					Name:       "jmx",
+					Name:       "jfr-jmx",
 					Port:       9091,
 					TargetPort: intstr.IntOrString{IntVal: 9091},
 				},
--- a/pkg/controller/service/service_controller.go
+++ b/pkg/controller/service/service_controller.go
@@ -139,11 +139,14 @@ func (r *ReconcileService) Reconcile(request reconcile.Request) (reconcile.Resul
 	return reconcile.Result{}, nil
 }
 
-const containerJFRPort int = 9091 // TODO make a property in ContainerJFR CRD?
+const defaultContainerJFRPort int = 9091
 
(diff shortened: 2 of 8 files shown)


andrewazores:
Rebased.

pkg/controller/service/service_controller.go
```diff
@@ -139,15 +145,23 @@ func (r *ReconcileService) Reconcile(request reconcile.Request) (reconcile.Resul
 	return reconcile.Result{}, nil
 }
 
-const containerJFRPort int = 9091 // TODO make a property in ContainerJFR CRD?
-
 func isJFRAwareService(svc *corev1.Service) bool {
+	_, err := getServiceJMXPort(svc)
+	return err == nil
+}
```
ebaron: Perhaps we should get rid of this function and just retain the JMX port number (if any), then we wouldn't have to call `getServiceJMXPort` again when creating the FlightRecorder object.

pkg/controller/service/service_controller.go
```diff
@@ -115,8 +116,13 @@ func (r *ReconcileService) Reconcile(request reconcile.Request) (reconcile.Resul
...
 	found := &rhjmcv1alpha1.FlightRecorder{}
 	err = r.client.Get(ctx, types.NamespacedName{Name: jfr.Name, Namespace: jfr.Namespace}, found)
-	if err != nil && errors.IsNotFound(err) {
+	if err != nil && kerrors.IsNotFound(err) {
 		reqLogger.Info("Creating a new FlightRecorder", "Namespace", jfr.Namespace, "Name", jfr.Name)
+		jmxPort, err := getServiceJMXPort(svc)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		jfr.Spec.Port = jmxPort
```
ebaron: We probably also need to handle the case where the port number is changed in the service after creation, and update the FlightRecorder accordingly.

pkg/controller/service/service_controller.go
andrewazores: Makes sense, I'll do that

pkg/controller/service/service_controller.go
andrewazores: Good catch, I'll give that a shot too

andrewazores:
Sure, I'll rename the port to something less likely to conflict.

Fair point on looking up the service again anyway, but now I'm also thinking maybe it still makes sense to keep the port number in the FlightRecorderSpec. This way the user can override it themselves if they need to. I can't really think of a good reason why that might be the case, but the flexibility is probably good to have. Maybe the user for some reason doesn't want to modify their configuration to name the port, and has something else already bound to port 9091 and JMX on some other port number?

pkg/controller/containerjfr/resource_definitions/resource_definitions.go
```diff
@@ -254,12 +254,12 @@ func NewExporterService(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Service {
...
 			Ports: []corev1.ServicePort{
 				{
-					Name:       "8181-tcp",
+					Name:       "export",
 					Port:       8181,
 					TargetPort: intstr.IntOrString{IntVal: 8181},
 				},
 				{
-					Name:       "9091-tcp",
+					Name:       "jmx",
```
ebaron: I think this also needs to be changed to the `jfr-jmx` name.

andrewazores: Ah right. I'll correct that, thanks.

andrewazores:
@ebaron I think the latest round of patches addresses your comments from last time. How does it look now?

PR #56, opened 2020-01-21
Websocket subprotocol token

### LINKED ISSUE #52: WebSocket auth token should be sent via subprotocol
This is a pre-emptive bug against #45 . That PR updates a controller to supply the service account auth token via a query parameter. However, there is planned work at rh-jmc-team/container-jfr#100 to switch from a query parameter to a websocket subprotocol, so the implementation here will need to be updated to mirror this to correspond to that future work item.

andrewazores:
Fixes #52

### DIFF
--- a/pkg/client/containerjfr_client.go
+++ b/pkg/client/containerjfr_client.go
@@ -52,18 +52,15 @@ func (client *ContainerJfrClient) Close() error {
 }
 
 func newWebSocketConn(server *url.URL, token *string, tlsVerify bool) (*websocket.Conn, error) {
-	q := server.Query()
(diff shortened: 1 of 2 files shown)


andrewazores:
That would probably make sense to do, especially given the prefix I chose (ahem, copied from what OpenShift does). Currently on the receiving end, container-jfr isn't going to decode the token, so the sending side here should also just be the raw token string. I'm going to need to work on that anyway when I get around to rh-jmc-team/container-jfr#103 , so I'll add a note there to make sure that the base64 encode/decode is also handled better then. Thanks for the catch!

PR #69, opened 2020-02-03
Add documentation on how to use the API

ebaron:
This PR adds a new documentation file with a brief introduction to the FlightRecorder API provided by our operator, along with instructions on how to use it to list, create and download recordings.

### DIFF
--- a/API.md
+++ b/API.md
@@ -77,11 +77,11 @@ oc edit -o json flightrecorders/containerjfr
     "spec": {
         "port": 9091,
         "recordingRequests": [
-                {
-                        "name": "my-recording",
-                        "eventOptions": [ "jdk.SocketRead:enabled=true", "jdk.SocketWrite:enabled=true" ],
-                        "duration": "30s"
-                }
+            {
+                "name": "my-recording",
+                "eventOptions": [ "jdk.SocketRead:enabled=true", "jdk.SocketWrite:enabled=true" ],
+                "duration": "30s"
+            }
         ]
     },
     "status": {
--- a/API.md
+++ b/API.md
@@ -86,6 +86,15 @@ oc edit -o json flightrecorders/containerjfr
     },
     "status": {
         "recordings": [],
+        "target": {
+            "apiVersion": "v1",
+            "kind": "Service",
+            "name": "containerjfr",
+            "namespace": "default",
+            "resourceVersion": "298899",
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -0,0 +1,141 @@
...
+        "recordingRequests": [
+                {
+                        "name": "my-recording",
+                        "eventOptions": [ "jdk.SocketRead:enabled=true", "jdk.SocketWrite:enabled=true" ],
+                        "duration": "30s"
+                }
+        ]
+    },
+    "status": {
+        "recordings": [],
```
andrewazores: Looks like the JSON lost some braces and such here

API.md
```diff
@@ -0,0 +1,141 @@
...
+    },
+    "status": {
+        "recordings": [
+            {
+                "active": true,
+                "duration": "30s",
+                "name": "my-recording",
+                "startTime": "2020-01-23T15:55:04Z"
+            }
+        ],
```
andrewazores: Same here, looks like some missing braces

API.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+You'll need to pass your bearer token with the curl request. (You may also need -k if your test cluster uses a self-signed certificate)
+```shell
+$ curl -k -H "Authorization: Bearer $(oc whoami -t)" https://containerjfr-default.apps-crc.testing:443/recordings/172-30-157-32_my-recording_20200123T155535Z.jfr my-recording.jfr
```
andrewazores: `oc whoami -t | base64` I think

API.md
ebaron: This was actually cropped to only show the relevant parts of the JSON. I'll add the missing closing braces at least.

API.md
ebaron: Is Base64 encoding for the web server as well or just for commands over WebSockets? Using version 0.13.0, I only succeed without the encoding.

andrewazores: Ah, you're right - whatever charset is used by some auth system's tokens is probably a permissible subset of what's allowed in HTTP headers, so I didn't add base64 encode/decode there. But now I'm questioning if that was a strong enough assumption. Anyway, your example reflects the actual current state of things, so it's good.

API.md
ebaron: This should be fixed now.

API.md
ebaron: Should also be fixed.

PR #76, opened 2020-02-25
Use a UUID when communicating with command server

ebaron:
This PR takes advantage of the ID field added to command messages with https://github.com/rh-jmc-team/container-jfr/pull/117. We generate a UUID upon creating a CommandMessage using apimachinery's UUID util package.

When reading a response, we set this expected ID in the response struct that our decoder will attempt to unmarshal into. This allows our custom JSON decoder to verify the read ID is what we're expected before attempting to unmarshal the rest of the response. This prevents us from unmarshalling an unexpected payload type.

If the ID in the response is not what we're expecting, we return a special error indicating that the message was meant for another client. Our client will read in a loop until we receive the correct response, or some other error occurs. This seems to work well, but I'm a bit concerned something goes wrong and our client loops endlessly. Perhaps we should set some kind of timeout?

### DIFF
--- a/pkg/client/containerjfr_client.go
+++ b/pkg/client/containerjfr_client.go
@@ -8,12 +8,16 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/gorilla/websocket"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
 var log = logf.Log.WithName("containerjfr_client")
+var debugLog = log.V(1)
+
+const ioTimeout = 30 * time.Second
 
 // Config stores configuration options to connect to Container JFR's
 // command server
@@ -166,6 +170,7 @@ func (client *ContainerJfrClient) ListSavedRecordings() ([]SavedRecording, error
 }
 
 func (client *ContainerJfrClient) syncMessage(msg *CommandMessage, responsePayload interface{}) error {
+	client.conn.SetWriteDeadline(time.Now().Add(ioTimeout))
 	err := client.conn.WriteJSON(msg)
 	if err != nil {
 		log.Error(err, "could not write message", "message", msg)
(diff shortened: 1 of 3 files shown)


pkg/client/containerjfr_client.go
```diff
@@ -194,3 +191,20 @@ func (client *ContainerJfrClient) syncMessage(msg *CommandMessage, responsePaylo
...
+	// Set the original message ID so our decoder can verify it.
+	// By setting the output argument in the struct, the decoder knows what type
+	// to unmarshall the payload into and also stores it for returning to the caller.
+	resp := &ResponseMessage{ID: msg.ID, Payload: responsePayload}
+
+	// Keep reading messages until we find a response to this message,
+	// or encounter an error other than ErrWrongID
+	for {
+		err := client.conn.ReadJSON(resp)
+		if err == nil || err != ErrWrongID {
```
andrewazores: Would it make sense to log the `ErrWrongID`s at some lower log level? It seems right now that if somehow the operator does end up stuck in this loop that it'll be difficult to diagnose it as being in that state.

I do agree that some form of timeout would also make sense here. Maybe 10 seconds? 30? Most commands should have very short response latencies, but it also depends on various external factors - since we're talking about an Operator here then there is most likely an authentication system enabled, so message response times are also subject to authz latencies, for example.

ebaron: I've changed this to a debug-level message, and also added an argument to the dev_operator setting the log level to debug.

I added a 30 second timeout for WebSocket reads and writes, and also used a 30 second timeout to exit the read loop. We could still get a 29 second read with a wrong ID and a 30 second one that times out, but the upper bound should be < 60 seconds. Still, this prevents it from hanging indefinitely.

ebaron: I should point out that the logging happens in UnmarshalJSON within `command_types.go`, right before `ErrWrongID` is returned.

Issue #51, opened 2020-01-08
Grafana dashboard container uses default admin auth credentials


andrewazores:
The Grafana dashboard container comes pre-configured with a default admin account. This should be disabled and some sort of secret configuration added so that the outside world does not have admin access to the Grafana container. Relating to #50 , there should be some access token that the operator service account can use for configuring the Grafana instance, and this (or a similar) token, or some non-default credentials, should be somehow attached to the Grafana instance in a way that the cluster admin is still able to get admin access into the Grafana instance as well, in case there is a need for additional configuration.

andrewazores:
https://grafana.com/docs/grafana/latest/installation/docker/

https://grafana.com/docs/grafana/latest/auth/overview/

https://grafana.com/docs/grafana/latest/auth/generic-oauth/

Grafana has what looks to be a pretty easy way to hook in to external OAuth. We should be able to set this up to work with OpenShift OAuth, possibly by also running OpenShift oauth-proxy in the container-jfr pod.

andrewazores:
Ideally we don't add the oauth-proxy container, since this carries again some more overhead that can probably be avoided, and which we are really trying to avoid carrying for the minimal deployment case in particular. And we definitely want to avoid any scenario where we conditionally deploy the proxy for full setups and use direct auth (as in the current implementation) for minimal setups.

https://docs.openshift.com/container-platform/4.1/authentication/understanding-authentication.html

https://docs.openshift.com/container-platform/4.1/authentication/configuring-internal-oauth.html

andrewazores:
https://community.grafana.com/t/openshift-generic-oauth/2148

https://blog.sakuragawa.moe/deploy-granafa-with-prometheus-and-oauth2-on-openshift/

andrewazores:
Ran into a stumbling block:

https://github.com/openshift/origin/issues/18013

andrewazores:
Possible workaround:

https://community.grafana.com/t/how-to-config-api-url-when-there-is-no-userinfo-endpoint-provided-by-my-oauth-server/719/4

andrewazores:
The credentials for the default basic auth admin user can be configured by setting the environment variables `GF_SECURITY_ADMIN_USER` and `GF_SECURITY_ADMIN_PASSWORD`. If the operator sets these (or at least the password) in the Grafana container env vars then the credentials can at least be non-default and customized by the admin for each installed operator instance.

[...]

Issue #50, opened 2020-01-08
Grafana controller uses basic auth for configuration


andrewazores:
Currently in grafana_controller.go , the controller uses basic auth with default credentials to access the Grafana dashboard instance and add configurations for the jfr-datasource and dashboard panels. This should be replaced with some form of access token provided by the Grafana API and added in request headers.

andrewazores:
This situation has improved in that the credentials are now auto-generated (well, the password is - the default username is still used), however the auth method is still simply HTTP Basic.

andrewazores:
Closing, generated credentials for Basic behind TLS termination seems satisfactory (for now anyway).

PR #77, opened 2020-03-03
Use generated Grafana credentials, not default

andrewazores:
See #50 and #51 . This PR results in Grafana being configured to use a generated
password rather than the insecure default credentials for its normal HTTP Basic
auth scheme. The generated password is stored in a Secret, with the contents
written into the Grafana container's environment variables. These generated
credentials are also retrieved and used by the Operator itself when it performs
its startup configuration of the Grafana instance (creating the datasource
and dashboard).

### DIFF
--- a/pkg/controller/grafana/grafana_controller.go
+++ b/pkg/controller/grafana/grafana_controller.go
@@ -91,6 +91,18 @@ func (r *ReconcileGrafana) Reconcile(request reconcile.Request) (reconcile.Resul
 	reqLogger.Info("Found Grafana service", "Namespace",
 		svc.Namespace, "Name", svc.Name)
 
+	secret := &corev1.Secret{}
+	err = r.client.Get(ctx, types.NamespacedName{Name: svc.Name + "-basic", Namespace: request.Namespace}, secret)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			reqLogger.Info("Could not find Grafana credentials secret")
+			return reconcile.Result{Requeue: true}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	reqLogger.Info("Found Grafana credentials secret", "Namespace",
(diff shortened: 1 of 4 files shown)


pkg/controller/grafana/grafana_controller.go
```diff
@@ -240,6 +252,12 @@ func (r *ReconcileGrafana) configureGrafanaDashboard(route *openshiftv1.Route) e
 	return nil
 }
 
+func GetCredentialedHostPathUrl(secret *corev1.Secret, host string, path string) string {
+	user := secret.Data["GF_SECURITY_ADMIN_USER"]
+	pass := secret.Data["GF_SECURITY_ADMIN_PASSWORD"]
+	return fmt.Sprintf("https://%s:%s@%s%s", strings.TrimSpace(string(user)), strings.TrimSpace(string(pass)), host, path)
```
ebaron: Quick question. Does the secret `Data` slice have Base64 encoded values in it? If so, would we need to decode them before inserting them into the URL?

andrewazores: Yea, I was a bit confused about that after reading the docs, but when I tested it manually I found that when reading/writing from the `Secrets` object, the data seemed to be automatically encoded/decoded. However when you view the contents of the secret with ex. `oc get -o json secret foo`, it's still encoded. I'll double check all that though.

andrewazores: Just checked by adding in a log statement when the grafana_controller reads the credentials here and they are indeed already decoded at this point, but when you do an `oc get secret` it's still encoded.

ebaron: Thanks for checking. The documentation was a bit unclear about that, and I couldn't find where the decoding happens. Seems good to go from me.

PR #80, opened 2020-04-23
Add UPL license and headers

ebaron:
This PR adds a LICENSE file to the root of the project, along with license headers to all Go source files. UPL was chosen to match the JMC project.


andrewazores:
Was there some script you used to apply the license to the Go sources, or did you do it by hand? I think adding a script for adding/updating the licenses automatically would be a nice addition.

ebaron:
I just did it with a simple bash loop. It should be easy enough to add a script to do this and also check for existence. I opened https://github.com/rh-jmc-team/container-jfr-operator/issues/81 to keep track of this suggestion.

PR #82, opened 2020-04-27
Update API doc with changes from API revision

ebaron:
This PR updates the API usage documentation to reflect changes introduced by https://github.com/rh-jmc-team/container-jfr-operator/pull/72.

Fixes https://github.com/rh-jmc-team/container-jfr-operator/issues/78

### DIFF
--- a/API.md
+++ b/API.md
@@ -121,67 +121,89 @@ $ oc get -o json flightrecorders/jmx-listener
 
 ## Creating a new Flight Recording
 
-To start a new recording, you will need to add a new recording request to the `FlightRecorder` in the `spec.recordingRequests` array. You can do this nicely on the command line with `oc edit`. The recording request must include the following:
+To start a new recording, you will need to create a new `Recording` custom resource. The `Recording` must include the following:
 
-1. `name`: a string uniquely identifying the recording within that service
+1. `name`: a string uniquely identifying the recording within that service.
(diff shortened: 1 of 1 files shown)


API.md
```diff
@@ -1,168 +1,345 @@
...
+(Event listing abbreviated for readability)
 
 ## Creating a new Flight Recording
 
-To start a new recording, you will need to add a new recording request to the `FlightRecorder` in the `spec.recordingRequests` array. You can do this nicely on the command line with `oc edit`. The recording request must include the following:
+To start a new recording, you will need to create a new `Recording` custom resource. The `Recording` must include the following:
 
-1. `name`: a string uniquely identifying the recording within that service
+1. `name`: a string uniquely identifying the recording within that service.
 2. `eventOptions`: an array of string options passed to Container JFR. The `"ALL"` special string can be used to enable all available events.
```
andrewazores, line 127: I think this needs to be updated - since ContainerJFR gained support for remote target event templates, the event specifier string syntax for that is "template=Foo". The `ALL` special string was made into a pseudo-template, so it should be selected with `template=ALL`.

andrewazores:
Looks great other than the minor "template" note above.

API.md
ebaron: Understood, I've changed this now. I'll have to add some support for templates to the API in the future as well.

PR #85, opened 2020-06-01
Add controller to clean up recordings when operator is deleted

### LINKED ISSUE #84: Can't delete recordings after operator
Since we use a finalizer to clean up recording resources, recording deletion is prevented until the operator removes the finalizer. If the operator pod is deleted before this can happen, trying to delete recordings will hang.

To fix this, we could add an OwnerReference to Recordings, listing either the operator's pod or the ContainerJFR custom resource as the owner, with BlockOwnerDeletion set to true. If I understand correctly, an attempt to delete the owner should attempt to delete the dependents, while keeping the owner alive to handle the finalizer.

ebaron:
The goal of this PR is to keep the operator alive long enough to finalize any created recordings when the operator is being deleted. I've tried a few approaches to make this work, but this is the first that suits our needs. This new cleanup controller watches only the operator's deployment and adds a finalizer to it. When the operator is being deleted, the controller deletes all recordings. The recordings in turn have their own finalization logic, which cleans up external JFR/PVC resources. Once the recordings have been deleted, the finalizer is removed from the operator deployment.

Fixes #84

### DIFF
--- a/pkg/controller/cleanup/cleanup_controller.go
+++ b/pkg/controller/cleanup/cleanup_controller.go
@@ -2,11 +2,13 @@ package cleanup
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	"github.com/rh-jmc-team/container-jfr-operator/pkg/apis/rhjmc/v1alpha2"
 	rhjmcv1alpha2 "github.com/rh-jmc-team/container-jfr-operator/pkg/apis/rhjmc/v1alpha2"
 	appsv1 "k8s.io/api/apps/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -112,7 +114,7 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 	instance := &appsv1.Deployment{}
 	err := r.client.Get(ctx, request.NamespacedName, instance)
 	if err != nil {
-		if errors.IsNotFound(err) {
+		if kerrors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
@@ -130,6 +132,7 @@ func (r *ReconcileDeployment) Reconcile(request reconcile.Request) (reconcile.Re
 			if err != nil {
 				log.Error(err, "failed to delete recordings", "namespace", instance.Namespace,
 					"name", instance.Name)
+				return reconcile.Result{}, err
 			}
 
 			// Remove our finalizer only once our cleanup logic has succeeded
@@ -166,6 +169,16 @@ func (r *ReconcileDeployment) deleteRecordings(ctx context.Context, namespace st
 		},
 	})
 	// TODO Do we need finalizer on ContainerJFR CR as well?
+
+	// Check if any recordings still exist
+	recordings := &v1alpha2.RecordingList{}
+	err = r.client.List(ctx, recordings)
+	if err != nil {
+		return err
+	}
+	if len(recordings.Items) > 0 {
+		return fmt.Errorf("%d recordings still exist in %s", len(recordings.Items), namespace)
+	}
 	return err
 }
 
--- a/pkg/controller/cleanup/cleanup_controller.go
+++ b/pkg/controller/cleanup/cleanup_controller.go
@@ -1,0 +1,217 @@
+package cleanup
+
+import (
+	"context"
+
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	rhjmcv1alpha2 "github.com/rh-jmc-team/container-jfr-operator/pkg/apis/rhjmc/v1alpha2"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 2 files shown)


pkg/controller/cleanup/cleanup_controller.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	// Check if this operator deployment is being deleted
+	if instance.GetDeletionTimestamp() != nil {
+		if hasCleanupFinalizer(instance) {
+			// Cleanly delete all recordings
+			err := r.deleteRecordings(ctx, r.watchNs)
+			if err != nil {
+				log.Error(err, "failed to delete recordings", "namespace", instance.Namespace,
+					"name", instance.Name)
```
andrewazores, line 132: Should this return the `err` and restart the reconcile loop or is it meant to always fall through to removing the finalizer from here?

andrewazores:
Hmm. I checked out this branch and did the following (abbreviated):

```
$ make clean image && podman push $IMAGE_STREAM:$IMAGE_VERSION && make deploy
$ oc create -f example-recording.yaml # after waiting for everything to be deployed and running
$ oc get recordings
NAME      AGE
example   2m7s
$ oc delete deployment container-jfr-operator
$ oc get recordings
NAME      AGE
example   3m16s
$ # oc get all shows that the operator deployment and pod have been deleted
```

I may just be misunderstanding what this does or how to test it... the code looked good from a read-over, but I'm not seeing the behaviour I expected to see.

pkg/controller/cleanup/cleanup_controller.go
ebaron: Yes it should. Good catch!

ebaron:
I'm seeing this a lot when deleting the operator deployment:
```
{"level":"error","ts":1591129749.7243004,"logger":"controller_recording",
"msg":"failed to remove finalizer from recording","namespace":"default","name":"my-recording",
"error":"Operation cannot be fulfilled on recordings.rhjmc.redhat.com \"my-recording\": 
the object has been modified; please apply your changes to the latest version and try again"
```

Not sure why this keeps coming up, but I've added a check to requeue if any recordings still exist after the delete operation.

andrewazores:
That last change makes sense. Do you have a specific way to trigger that behaviour that I can test? I think it's still fine if not.

ebaron:
This doesn't seem to work when the operator is managed by OLM. If you install using the CSV and delete it, the cleanup controller fails to delete the recordings. This appears to be because the necessary RBAC objects have already been deleted.

```
{"level":"info","ts":1591389841.3301594,"logger":"controller_cleanup","msg":"Reconciling Operator Deployment","Request.Namespace":"default","Request.Name":"container-jfr-operator"}
{"level":"error","ts":1591389841.3466809,"logger":"controller_cleanup","msg":"failed to delete recordings","namespace":"default","name":"container-jfr-operator","error":"Unauthorized","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/ebaron/go/pkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/rh-jmc-team/container-jfr-operator/pkg/controller/cleanup.(*ReconcileDeployment).Reconcile\n\tcontainer-jfr-operator/pkg/controller/cleanup/cleanup_controller.go:133\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/home/ebaron/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:256\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/ebaron/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:232\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/home/ebaron/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:211\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/ebaron/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/ebaron/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/ebaron/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/util/wait/wait.go:88"}
```

We could place finalizers on those as well perhaps, or reconsider the whole approach. You can point your CatalogSource to the following to try it out:
```
image: quay.io/ebaron/container-jfr-operator-index@sha256:81a0d5abc56cbfd3ddd5286c3c9082fb78240a1f2b64c897174f4106234b822b
```

andrewazores:
Ugh, that's annoying. Did you say that there was an issue open somewhere in one of the Operator-related upstream repos for how to properly handle deletion cascading like this? It seems like the approach you have here should be enough, so long as the infrastructure deletes resources in the right order (and consistently so). It really seems like it shouldn't be this hard for an operator implementation to clean itself up. If we were to publish this as-is and a user didn't properly delete the ContainerJFR CR(D), don't they end up with the deletion hanging? Does this eventually time out or are they just going to end up with their cluster in a bad state?

ebaron:
Agreed, this turned out to be much more difficult than I expected. Here's the thread that I based this implementation off of: https://github.com/operator-framework/operator-sdk/issues/2706. AFAIK, it does not time out and the objects will stay around until the user manually removes the finalizers from the affected objects.

andrewazores:
Hmm. I'm not familiar enough with this aspect to have any better or different ideas on how else we could implement the cleanup. I guess for now maybe the best approach really would be to add finalizers to everything? That doesn't seem like the most maintainable way to go, but I don't think that requiring users to do manual cleanup steps before uninstalling from the marketplace would go over very well either.

I'm okay with this PR being merged as-is for now since it does seem to solve the problem when the operator is installed manually at least, and then a follow-up PR later on can try to tackle solving the problem for marketplace installations.

ebaron:
@andrewazores I posted a proposal https://github.com/rh-jmc-team/container-jfr-operator/issues/84#issuecomment-639864580 that might be a better way to handle this. The user would be required to delete the ContainerJFR CR to cleanly remove everything created by the operator before uninstalling it. Could be seen as doing the opposite of the installation procedure.

andrewazores:
Yea, that's not that bad, although it still seems unfortunate that manual intervention is required to get a clean deletion. At least it's a single (well, per namespace), easy step.

ebaron:
After discussing with the Operator SDK team, it's best practice to leave the CRs in the hands of the user upon uninstall. If they choose to uninstall the operator before deleting recordings, they have to be responsible for removing the finalizers from the recordings.

PR #86, opened 2020-06-16
Update container-jfr image version

andrewazores:
Also update controllers for updated connectionless API, currently
emulating a connection-oriented interface by holding internal state.
Finally, also replaces "ALL" event string with "template=ALL".

### DIFF
--- a/deploy/crds/rhjmc.redhat.com_v1alpha2_recording_cr.yaml
+++ b/deploy/crds/rhjmc.redhat.com_v1alpha2_recording_cr.yaml
@@ -4,7 +4,7 @@ metadata:
   name: example-recording
 spec:
   name: example-recording
-  eventOptions: [ ALL ]
+  eventOptions: [ "template=ALL" ]
   duration: 30s
   archive: true
   flightRecorder:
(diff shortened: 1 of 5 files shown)


andrewazores:
The (dis)connect methods on the `JfrClient` aren't really necessary anymore and therefore the `ClientLock` mutex shouldn't be, either, but I've left things as similar as possible here intentionally to minimize the risk of any regressions.

deploy/olm-catalog/container-jfr-operator-bundle/0.4.0/container-jfr-operator-bundle.v0.4.0.clusterserviceversion.yaml
```diff
@@ -38,7 +38,7 @@ metadata:
             "archive": true,
             "duration": "30s",
             "eventOptions": [
-              "ALL"
+              "template=ALL"
```
ebaron, line 41: Perhaps we should leave the 0.4.0 CSV as-is, and do a new "release" with version bump and updated CSV?

andrewazores: Sure, I'll back out that part of the change. I think you did the last CSV update - do you want to do this next one too?

ebaron: Sure, I can do that.

PR #90, opened 2020-07-14
Remove unneeded state and synchronization for Container JFR client

ebaron:
With Container JFR no longer requiring a stateful connection for its commands, we can simplify the WebSocket client used by controllers in the operator. Synchronization is no longer necessary since interleaving of commands is not an issue anymore. Commands that require a target take a new TargetAddress struct as a parameter. This is an abstraction of a host/port pair. I also added some missing comments in this PR.

### DIFF
--- a/pkg/controller/common/common_reconciler.go
+++ b/pkg/controller/common/common_reconciler.go
@@ -57,6 +57,8 @@ import (
 
 var log = logf.Log.WithName("common_reconciler")
 
+// CommonReconciler contains helpful methods to communicate with Container JFR.
+// It is meant to be embedded within other Reconcilers.
 type CommonReconciler struct {
 	// This client, initialized using mgr.Client(), is a split client
 	// that reads objects from the cache and writes to the apiserver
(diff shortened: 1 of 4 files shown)


pkg/controller/common/common_reconciler.go
```diff
@@ -103,33 +107,20 @@ func (r *CommonReconciler) ConnectToContainerJFR(ctx context.Context, namespace
 	return jfrClient, nil
 }
 
-func (r *CommonReconciler) ConnectToPod(targetPod *corev1.Pod, jmxPort int32) error {
+// GetPodTarget returns a TargetAddress for a particular pod and port number
+func (r *CommonReconciler) GetPodTarget(targetPod *corev1.Pod, jmxPort int32) (*jfrclient.TargetAddress, error) {
 	// Have Container JFR connect to the target JVM
```
andrewazores, line 112: This comment looks like it's outdated too

pkg/controller/recording/recording_controller.go
```diff
@@ -189,19 +184,16 @@ func (r *ReconcileRecording) Reconcile(request reconcile.Request) (reconcile.Res
 	}
 
 	// Tell Container JFR to connect to the target pod
```
andrewazores, line 186: Same here

andrewazores:
Nice, changes look good. 'll give this a quick smoketest tomorrow just to be sure.

pkg/controller/common/common_reconciler.go
ebaron: Good catch, fixed this.

pkg/controller/recording/recording_controller.go
ebaron: Fixed this too.

PR #92, opened 2020-07-20
Reorder undeploy target to better handle recordings

ebaron:
This PR simply reorders the `oc delete` commands in `undeploy`. This avoids deleting the operator deployment before any created recordings, and requires you to manually remove the finalizer from those.


andrewazores:
How does one manually remove the finalizers? Do we have that documented somewhere?

ebaron:
I use `oc edit` and remove the finalizers property altogether from the objects. It works the same as any other object update.

Issue #97, opened 2020-08-04
Operator should be able to set ContainerJFR's RJMX credentials


andrewazores:
See https://github.com/rh-jmc-team/container-jfr/issues/221

ebaron:
For now, this should involve generating a password (using "containerjfr" for the username is probably fine), and passing the credentials to Container JFR in the deployment's pod template with the environment variables added in https://github.com/rh-jmc-team/container-jfr/pull/230.

For generating the password, perhaps we can make use of this existing function:
https://github.com/rh-jmc-team/container-jfr-operator/blob/2f6dce3ae7918c117c55bfd0fbbc821544591ff3/pkg/controller/containerjfr/resource_definitions/resource_definitions.go#L248-L256

ebaron:
Fixed with #103

PR #103, opened 2020-08-07
set ContainerJFR's RJMX password

Alexjsenn:


Alexjsenn:
I ran it in crc and it seems to work.

Alexjsenn:
#97

ebaron:
Cool, nice and simple. Just to make it explicit, could we also define `CONTAINER_JFR_RJMX_USER` with "containerjfr" before this?

andrewazores:
> #97

Btw, if you'd like issues to be auto-closed when your PRs are merged, include text like "Fixes #97" or "Closes #97" in your commit messages. Linking it in the PR discussion like this is good because at least it automatically creates links between the Issue and PR, but then someone still has to go and manually close the issue afterward.

Alexjsenn:
Fixes #97

Alexjsenn:
> Cool, nice and simple. Just to make it explicit, could we also define `CONTAINER_JFR_RJMX_USER` with "containerjfr" before this?

Yeah, makes sense!

PR #108, opened 2020-09-11
Replace WebSocket client with HTTP client

ebaron:
This PR changes the way the FlightRecorder and Recording controllers communicate with Container JFR. Before this was done using the now deprecated WebSockets API, now it is done using the HTTP REST API. The HTTP API will enable us to use newly added security features in Container JFR.

The work is mostly a drop-in replacement of the underlying client code, but I've also done some refactoring to simplify how reconcilers obtain the client. This is partially because it no longer maintains a connection to Container JFR.

One important note is that this PR has a temporary commit to use a custom version of the Container JFR container image that supports the HTTP REST API. We should push a new version of the image before merging this PR. Since this version defaults to using JMX authentication, which is not yet supported in the operator, I've disabled this for now using `CONTAINER_JFR_DISABLE_JMX_AUTH`.

I've noticed some errors connecting to Container JFR while testing locally with CRC. The requests work eventually, but they seem to fail somewhat frequently and not just during startup. I'm not sure why it's happening, but perhaps it's an issue on my end. Maybe I need to extend the timeout, although I would have expected 30 seconds to be enough. I've seen the following failures:

`"Get https://containerjfr.default.svc:8181/api/v1/targets/10.116.0.58:9093/recordings: dial tcp 172.25.38.147:8181: connect: connection refused"`

`"Get https://containerjfr.default.svc:8181/api/v1/targets/10.116.0.58:9093/recordings: dial tcp 172.25.38.147:8181: i/o timeout"`

`"Get https://containerjfr.default.svc:8181/api/v1/targets/10.116.0.58:9093/recordings: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)"`


andrewazores:
Looks great from a reading, I'll build and test it shortly. In the meantime I've opened a PR ( https://github.com/rh-jmc-team/container-jfr/pull/265 ) to release a new ContainerJFR image version from current `main`, which includes all of the relevant HTTP API changes for this PR.

> I've noticed some errors connecting to Container JFR while testing locally with CRC. The requests work eventually, but they seem to fail somewhat frequently and not just during startup. I'm not sure why it's happening, but perhaps it's an issue on my end. Maybe I need to extend the timeout, although I would have expected 30 seconds to be enough. I've seen the following failures:

Hmm, that's not encouraging. I wonder if it's a resource constraints problem with CRC. If/when we can get access to a more high-powered OpenShift instance this should be one of the top priorities for us to check out. I occasionally see `connection refused` failures running ContainerJFR's integration tests where there's no apparent reason for the connection to be refused - the container's logs show it seemingly running normally, but its WebServer logs don't reflect it receiving any requests at all. This is intermittent and I haven't found any way to reproduce it, but whatever the cause is could be the same as what you're seeing.

andrewazores:
`v0.20.0` images are now published on Quay

ebaron:
> `v0.20.0` images are now published on Quay

Thanks! I've updated the PR to use it.

andrewazores:
When I run this the Operator logs look fairly normal from what I can tell, and `oc get all` shows everything deployed as expected, but after a few minutes `oc status` indicates that the ContainerJFR deployment has failed and the pod has been restarted several times. Opening `https://containerjfr-default.apps-crc.testing` in a browser also results in a `502 Bad Gateway`. I made this small patch that seems to fix it:

```
diff --git a/pkg/controller/containerjfr/resource_definitions/resource_definitions.go b/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
index 9f7971c..60bfa22 100644
--- a/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
+++ b/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
@@ -160,6 +160,10 @@ func NewCoreContainer(cr *rhjmcv1alpha1.ContainerJFR, specs *ServiceSpecs) corev
                        Name:  "CONTAINER_JFR_ALLOW_UNTRUSTED_SSL",
                        Value: "true",
                },
+               {
+                       Name:  "CONTAINER_JFR_DISABLE_SSL",
+                       Value: "true",
+               },
                {
                        Name:  "CONTAINER_JFR_WEB_PORT",
                        Value: "8181",
diff --git a/pkg/controller/flightrecorder/flightrecorder_controller.go b/pkg/controller/flightrecorder/flightrecorder_controller.go
index bc9c7c2..c2f5efd 100644
--- a/pkg/controller/flightrecorder/flightrecorder_controller.go
+++ b/pkg/controller/flightrecorder/flightrecorder_controller.go
@@ -67,7 +67,8 @@ func Add(mgr manager.Manager) error {
 func newReconciler(mgr manager.Manager) *ReconcileFlightRecorder {
        return &ReconcileFlightRecorder{Client: mgr.GetClient(), Scheme: mgr.GetScheme(),
                Reconciler: common.NewReconciler(&common.ReconcilerConfig{
-                       Client: mgr.GetClient(),
+                       Client:     mgr.GetClient(),
+                       DisableTLS: true,
                }),
        }
 }
diff --git a/pkg/controller/recording/recording_controller.go b/pkg/controller/recording/recording_controller.go
index d09e8aa..8ae12a2 100644
--- a/pkg/controller/recording/recording_controller.go
+++ b/pkg/controller/recording/recording_controller.go
@@ -72,7 +72,8 @@ func Add(mgr manager.Manager) error {
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
        return &ReconcileRecording{Scheme: mgr.GetScheme(), Client: mgr.GetClient(),
                Reconciler: common.NewReconciler(&common.ReconcilerConfig{
-                       Client: mgr.GetClient(),
+                       Client:     mgr.GetClient(),
+                       DisableTLS: true,
                }),
        }
 }

```

This patch should more or less retain the status quo as before this PR. With this, I don't see the pod being restarted and I'm able to visit the container's web-client in my browser normally, and a quick test of all of the graphical functionality looks like everything is working. I suspect that the new autogenerated self-signed SSL cert normally used by ContainerJFR breaks incoming connections from outside of the cluster due to the existing TLS edge termination configuration being stacked on top of the self-signed cert. I believe the cause of the restarting pod is that the liveness probe defined for the container is a simple `GET HTTP /api/v1/clienturl` - in `resource_definitions.go` line ~247, this could be configured to use the `HTTPS` scheme and I believe this would fix that half of the issue, but requests coming from outside the cluster would still fail due to the TLS setup.

ebaron:
Good catch! That probably explains why I was seeing those HTTP request errors as well. I've done as you suggested and disabled SSL for now.

andrewazores (approved):

andrewazores:
I repeated my test as above as well as tested creating a simple recording via CR and everything looks good.

PR #109, opened 2020-09-21
Deploy Container JFR with TLS using service serving certificates

ebaron:
This PR aims to support TLS in Container JFR when deployed in OpenShift from the operator. The idea is to use the default router wildcard certificate for the Container JFR route, but with TLS termination set to re-encrypt instead of edge. The internal service traffic is encrypted using a key-pair/cert created by the OpenShift service serving certificates mechanism [1]. The created certificate is signed by OpenShift's service CA, which is already trusted by the router. These keys/certs get converted from PEM format keys/certs into PKCS12 keystores/truststores using an init container and are passed to Container JFR using a mounted emptyDir volume [2].

This seems to work, except for enabling TLS for the remote JMX connection. After passing `-Djavax.net.debug=ssl,handshake` to Container JFR, I get the following error:
```
javax.net.ssl|ERROR|1C|RMI TCP Connection(16)-10.116.0.54|2020-09-18 20:33:23.111 GMT|TransportContext.java:318|Fatal (CERTIFICATE_UNKNOWN): Extended key usage does not permit use for TLS client authentication (
"throwable" : {
  sun.security.validator.ValidatorException: Extended key usage does not permit use for TLS client authentication
      at java.base/sun.security.validator.EndEntityChecker.checkTLSClient(EndEntityChecker.java:245)
      at java.base/sun.security.validator.EndEntityChecker.check(EndEntityChecker.java:146)
      at java.base/sun.security.validator.Validator.validate(Validator.java:277)
      at java.base/sun.security.ssl.X509TrustManagerImpl.validate(X509TrustManagerImpl.java:313)
      at java.base/sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:222)
      at java.base/sun.security.ssl.X509TrustManagerImpl.checkClientTrusted(X509TrustManagerImpl.java:123)
      at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.checkClientCerts(CertificateMessage.java:1267)
      at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.onConsumeCertificate(CertificateMessage.java:1186)
      at java.base/sun.security.ssl.CertificateMessage$T13CertificateConsumer.consume(CertificateMessage.java:1163)
      at java.base/sun.security.ssl.SSLHandshake.consume(SSLHandshake.java:392)
      at java.base/sun.security.ssl.HandshakeContext.dispatch(HandshakeContext.java:444)
[...]

Issue #101, opened 2020-08-04
API Enhancements


ebaron:
Some useful functionality in Container JFR is not exposed via the operator's API. Some of this is existing in Container JFR today and some still a work-in-progress. Let's track what sorts of changes should go into the next version of our CRDs.

Confirmed:
- [x] Listing event templates in FlightRecorder (https://github.com/rh-jmc-team/container-jfr-operator/issues/100)
- [x] Report URL in Recordings (https://github.com/rh-jmc-team/container-jfr-operator/issues/89)
- [x] JMX authentication credentials in FlightRecorder/Recording (https://github.com/rh-jmc-team/container-jfr-operator/issues/98)
- [x] Add TrustedCertSecrets property to ContainerJFR CRD (#142)
- [x] Add application URL to Container JFR CRD (#187)
- [x] Options to control creation of Ingress objects on Kubernetes (#120)
- [x] Document new API additions (#134)

Declined:
- ~container-jfr log level env var should be controllable via CR spec (#63)~

Some other ideas:
- Use Conditions to inform the user about errors in spec that require fixing. (e.g. authentication failures)
- Enable/Disable security features ~~TLS hostname verification parameter in ContainerJFR spec~~
- Allow user to archive running recordings and archive a recording more than once

ebaron:
With the documentation done, this is complete.

PR #115, opened 2020-10-09
Use cert-manager for TLS between services

ebaron:
This PR makes use of cert-manager, another operator that provides flexible issuing of certificates and integrating those certificates within Kubernetes [1]. The key advantages of cert-manager, compared to the OpenShift service cert feature used in #109, is the ability to issue certificates that support client authentication (needed for JMX), and not depending on OpenShift-specific functionality. It also has support for generating PKCS12 or JKS keystores when issuing certificates. This allows us to much more easily supply the certificate/key-pair to Container JFR.

The setup works as follows:
1. Bootstrap CA certificate using self-signed issuer
2. Create new certificate issuer using created CA certificate
3. Issue certificates for Container JFR and Grafana using the created CA

Once issued, cert-manager creates Secrets containing the key/certificate/CA/keystore for the certificate. For Container JFR, we mount the keystore and pass the file path in `KEYSTORE_PATH` and mount the CA certificate under `/truststore` so Container JFR will import it upon startup. For Grafana, we mount the key and certificate from the secret and set its configuration using the appropriate environment variables. The Container JFR and Grafana routes are set to use TLS re-encryption where outside traffic is presented with the router's default certificate and then forwarded to the services over HTTPS, trusting Container JFR's CA.

The Container JFR and Grafana clients have been set up to trust the created Container JFR CA, thus eliminating the need for the TLS_VERIFY environment variable and usage of InsecureSkipVerify in the TLS client configuration. I've removed this environment variable, and perhaps we can replace it with a more general option to disable TLS for services.

I've also taken the opportunity to convert the unit tests we have to use an HTTPS server. It required surprisingly few changes. The main change is obtaining the `httptest` certificate used to start the server and then creating a TLS secret for it, as cert-manager would.

For now, this PR assumes cert-manager is installed as a precondition. This can be done as follows:
`kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v1.0.2/cert-manager.yaml`

TODO:
- [ ] Lock down jfr-datasource, only need it to be available over localhost
- [ ] Detect certificate renewal and redeploy/update routes
- [ ] Update README with cert-manager recommended requirement
- [ ] Make TLS using cert-manager configurable, using HTTP for internal traffic when disabled
- [ ] Create new OLM bundle, listing cert-manager CRDs as required CRDs to trigger dependency resolution
- [ ] Improve documentation

[1] https://cert-manager.io


andrewazores:
Very cool, seems to be working as expected so far. I think some of the TODOs can be done as follow-up enhancements (ex. redeploy on renewal, make jfr-datasource only accessible locally, even making cert-manager optional/configurable).

ebaron:
> Very cool, seems to be working as expected so far. I think some of the TODOs can be done as follow-up enhancements (ex. redeploy on renewal, make jfr-datasource only accessible locally, even making cert-manager optional/configurable).

Thanks! I agree on those TODOs. I added an environment variable to disable cert-manager for now, since it wasn't too difficult. I'm going to add some more comments and file issues for the remaining tasks. Then I think this PR will be complete enough.

andrewazores:
Sounds good. Do you have any other features/fixes planned before we tag our v1.0 releases across the repos?

ebaron:
I'd like to do a CRD version bump adding some new fields. I've been compiling a list here: #101. I think the most important one of these would be JMX auth credentials, as that's still missing. Adding a way for the user to add trusted certs to Container JFR would be useful as well. Although with this PR, they can at least issue certs for their applications using the Container JFR CA, which is already trusted.

ebaron:
I think this PR is complete enough for review.

andrewazores:
Looks good. When I visit the web-client in my browser I get a Grafana configuration error, and `curl -vk https://containerjfr-default.apps-crc.testing/health` gets a `500` response, but the ContainerJFR logs suggest the problem is probably fixed in https://github.com/rh-jmc-team/container-jfr/pull/278 , which I assume is not yet present in the ContainerJFR image version deployed.

andrewazores (approved):

ebaron:
That's right, I ended up using Container JFR head for my own testing. We can make a release and update the version in the operator in a followup patch.

PR #116, opened 2020-10-14
Update Grafana container to 7.2.1

andrewazores:
See also https://github.com/rh-jmc-team/container-jfr/pull/292 for a quicker smoketest of the new version. It seems to be fully compatible with our dashboard definition and the jfr-datasource with no apparent regressions or other adjustments needed.


ebaron (approved):
The new version works fine for me. I tested it with the fixes from #115 rebased on top of this.

Issue #117, opened 2020-10-14
Lock down jfr-datasource to localhost traffic only


ebaron:
Since jfr-datasource is deployed within the same pod as Container JFR and Grafana, and does not need to be consumed directly by the user, we can bind over localhost and remove its service. This can be done with the `quarkus.http.host` system property. Currently this is hard-coded to 0.0.0.0 in the jfr-datasource Dockerfile [1]. We could make this configurable using an environment variable.

[1] https://github.com/rh-jmc-team/jfr-datasource/blob/8e4c0b595ee21af13d259ccd8c12ab5d1290a591/src/main/docker/Dockerfile.native#L6

PR #121, opened 2020-10-19
Restrict jfr-datasource to pod-internal traffic

ebaron:
This PR depends on https://github.com/rh-jmc-team/jfr-datasource/pull/54. Since communication to/from jfr-datasource is only between containers in the same pod, we can restrict it to loopback traffic only. This is particularly important since this communication is unencrypted.

The fix is fairly straightforward, aside from the liveness probe. HTTP/TCP probes appear to require the port be accessible outside of the pod. I was able to use a command probe instead where the `curl` command is done within the container itself.

Fixes: #117

### DIFF
--- a/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
+++ b/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
@@ -529,32 +529,6 @@ func NewGrafanaService(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Service {
 	}
 }
 
-func NewJfrDatasourceService(cr *rhjmcv1alpha1.ContainerJFR) *corev1.Service {
-	return &corev1.Service{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      cr.Name + "-jfr-datasource",
-			Namespace: cr.Namespace,
-			Labels: map[string]string{
-				"app":       cr.Name,
-				"component": "jfr-datasource",
-			},
-		},
-		Spec: corev1.ServiceSpec{
-			Type: corev1.ServiceTypeClusterIP,
-			Selector: map[string]string{
-				"app": cr.Name,
-			},
-			Ports: []corev1.ServicePort{
-				{
-					Name:       "8080-tcp",
-					Port:       8080,
-					TargetPort: intstr.IntOrString{IntVal: 8080},
-				},
-			},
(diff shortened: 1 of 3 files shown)


ebaron:
The PR currently includes temporary commits to point to a container-jfr image with Grafana fixes added after 0.20, and a jfr-datasource image built with https://github.com/rh-jmc-team/jfr-datasource/pull/54.

andrewazores (approved):

andrewazores:
Haven't tested it, but changes (other than image tag needing to be fixed) look simple enough.

andrewazores:
> The fix is fairly straightforward, aside from the liveness probe. HTTP/TCP probes appear to require the port be accessible outside of the pod. I was able to use a command probe instead where the curl command is done within the container itself.

Aha! That makes complete sense and I bet it's the pitfall I ran into in trying to get https://github.com/rh-jmc-team/container-jfr/pull/256 images working with the Operator's probes. I'll have to try that when I circle back.

ebaron:
Great, I hope that helps! I've removed the temporary commits now. The updated jfr-datasource image is available at `quay.io/rh-jmc-team/jfr-datasource:0.0.2`.

andrewazores:
Actually, this `127.0.0.1:8080` is repeated in a few different places - does it make sense to extract it to a constant somewhere?

ebaron:
Done! Also fixed a compile error (oops)

pkg/controller/containerjfr/resource_definitions/resource_definitions.go
```diff
@@ -420,6 +419,12 @@ func NewGrafanaContainer(cr *rhjmcv1alpha1.ContainerJFR, tls *TLSConfig) corev1.
 	}
 }
 
+const datasourceHost = "127.0.0.1"
+const datasourcePort = "8080"
+
+// DatasourceURL contains the fixed URL to jfr-datasource's web server
+const DatasourceURL = "http://" + datasourceHost + datasourcePort
```
andrewazores, line 426: I think this is missing a `:` to separate the host and port, isn't it?

ebaron: So there is. Thanks for catching it. Fixed now!

andrewazores (approved):

PR #125, opened 2020-10-29
Support JMX authentication in API

### LINKED ISSUE #96: WebSocket API calls should be replaced with HTTP REST calls
See:
- https://github.com/rh-jmc-team/container-jfr/pull/222
- https://github.com/rh-jmc-team/container-jfr/pull/225

JMX connections may be secured and require user authentication. For various reasons, this authentication layer will not be implemented across the WebSocket Command Channel, so any such API calls are being re-implemented with HTTP request handlers. The JMX auth credentials will then be supplied using an optional HTTP header. Connection failures will result in a `4xx` status code (currently `407`) from the ContainerJFR server to indicate to clients that the request was refused due to target connection authentication, as opposed to the `401` used to signify lacking client auth from ex. an OpenShift account token.

### LINKED ISSUE #98: FlightRecorder CRD should include target auth credentials to include with API calls
As a follow-on to https://github.com/rh-jmc-team/container-jfr-operator/issues/96 , there should be some way for users to specify target auth credentials for ContainerJFR to use when establishing a connection to any given target JVM. The most natural way to do this to me seems to be to add the credentials somehow to the FlightRecorder CRD, or perhaps link that CRD to a Secrets object containing the credentials, which the Operator can then read and include in an HTTP header when performing various API actions against the ContainerJFR server.

### LINKED ISSUE #114: Automatically apply FlightRecorder label to Recordings
The `rhjmc.redhat.com/flightrecorder` label is meant to allow for potential clients to query all recordings for a pod using a label selector [1]. The reference to this FlightRecorder is already present in the Recording object, so we could populate this label automatically in the Recording controller. This would save the user (or client) from applying this label themselves.

[1] https://github.com/rh-jmc-team/container-jfr-operator/blob/30c3791442501abffcf0b6273fc37dfe049a6c2c/API.md#creating-a-new-flight-recording

ebaron:
This PR begins the next API version for the operator, v1beta1. The only change so far is to add credentials for JMX authentication to the FlightRecorder CRD. These credentials are specified using the name of a secret, which should be in the same namespace as the FlightRecorder. There are also optional properties so specify alternate username and password keys to index the credentials within the secret. If not specified, these are "username" and "password", which are the defaults for Basic Auth type secrets.

When reconciling a FlightRecorder, these credentials are read from the secret and passed to the Container JFR HTTP client. The client then sends the appropriate `X-JMX-Authorization` header with all requests it sends. When reconciling a Recording, the controller finds the credentials in the linked FlightRecorder. In order to reconcile Recordings when these credentials are updated in the FlightRecorder, I've added a new Watch in the Recording controller to find and reconcile all recordings related to a FlightRecorder when it changes. In order to do this efficiently, I implemented #114 to ensure the FlightRecorder label is available for list queries.

For Container JFR itself, we generate its JMX credentials in the operator. Rather than requiring the user to fill in the credentials themselves, we detect that a FlightRecorder we create is for Container JFR itself and fill in the generated secret automatically. We test for the Container JFR service by checking if it is owned by the ContainerJFR CR in the given namespace, and also contains a matching JMX port.

I think it would be good to add Conditions in the FlightRecorder and Recording status to indicate to the user when there is missing or invalid JMX credentials, but this is probably for another PR.

While testing, I noticed that there was nothing present to verify that all expected HTTP requests were sent to the test Container JFR server. I've now added a check to verify the number of requests received by the server.

Fixes #114, Fixes #96, Fixes #98


andrewazores:
I took a quick skim and this looks good so far. Looking forward to testing it out.

ebaron:
I think this is in a finished state and ready for review now.

andrewazores:
Very cool. So far I've deployed it and tested that a creation/retrieval workflow with curl still works as a sanity check, and then performed the same workflow using `oc`, both on the container-jfr pod itself, and it worked like a charm.

Maybe this is better suited for a follow-up PR, but could the `vertx-fib-demo` sample app be updated to `0.4.0` and some example CRs created to go along with that too? That version of the sample app has SSL and auth (`admin:adminpass123`) enabled as well and it's what I've been generally using for testing with the `smoketest.sh` in container-jfr itself. I'll try to piece together this setup myself to be sure I understand how to properly use it - it doesn't really seem all that complicated, but I'm always unsure of whether I'm writing a new CR properly by hand.

ebaron:
> Very cool. So far I've deployed it and tested that a creation/retrieval workflow with curl still works as a sanity check, and then performed the same workflow using `oc`, both on the container-jfr pod itself, and it worked like a charm.
> 
Awesome!

> Maybe this is better suited for a follow-up PR, but could the `vertx-fib-demo` sample app be updated to `0.4.0` and some example CRs created to go along with that too? That version of the sample app has SSL and auth (`admin:adminpass123`) enabled as well and it's what I've been generally using for testing with the `smoketest.sh` in container-jfr itself. I'll try to piece together this setup myself to be sure I understand how to properly use it - it doesn't really seem all that complicated, but I'm always unsure of whether I'm writing a new CR properly by hand.

That sounds like a good idea. We would need some way to add its certificate to the truststore though, I think. I was thinking of adding a field in this revision of the ContainerJFR CRD to allow the user to specify certs to add to the truststore when deploying.

andrewazores:
Ah, right. Hmm... how would the truststore make it into the cluster for the CRD to reference it?

ebaron:
> Ah, right. Hmm... how would the truststore make it into the cluster for the CRD to reference it?
> 
> Using `oc` I mean, if that's possible - the user can just POST the cert to the ContainerJFR instance with curl or whatever and then create the CR referencing it after that. But I wonder if there might be a more streamlined way.

The field I had in mind would be something like a list of certificate secrets. Before the controller deploys Container JFR, it mounts these secrets as files under `/truststore` in the container.

Triggering automatic redeployment when the list/certs change is a bit more complicated. We could track the currently configured secrets in the ContainerJFR CR's Status, and redeploy when the user modifies the list of secrets in the Spec. This still would not capture the case where the content of the secret changes, e.g. cert renewal.

andrewazores (approved):

PR #126, opened 2020-11-06
Watch own deployment

### LINKED ISSUE #123: Container JFR controller should watch owned deployment
Right now the Container JFR controller watches ContainerJFR CRs and pods owned by a ContainerJFR CR. Since Container JFR is now created in a Deployment, the controller should watch Deployments owned by ContainerJFR CRs and not pods (which are owned by their replica set). This could be extended to watch all objects the controller creates: services, secrets, routes, etc., to re-create them if they are deleted.

Alexjsenn:
Fixes #123

### DIFF
--- a/pkg/controller/containerjfr/containerjfr_controller.go
+++ b/pkg/controller/containerjfr/containerjfr_controller.go
@@ -95,38 +95,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to secondary resource Deployements and requeue the owner ContainerJFR
-	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &rhjmcv1beta1.ContainerJFR{},
-	})
-	if err != nil {
-		return err
-	}
+	resources := []runtime.Object{&appsv1.Deployment{}, &corev1.Service{}, &corev1.Secret{}, &openshiftv1.Route{}, &corev1.PersistentVolumeClaim{}}
 
-	err = c.Watch(&source.Kind{Type: &corev1.Service{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &rhjmcv1beta1.ContainerJFR{},
-	})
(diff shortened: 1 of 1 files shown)


pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -96,7 +96,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 	// TODO(user): Modify this to be the types you create that are owned by the primary resource
 	// Watch for changes to secondary resource Pods and requeue the owner ContainerJFR
```
ebaron, line 98: Could you change this comment to read Deployments instead of Pods, and remove the TODO above it?

ebaron (changes requested):
Hey Alex, looks good so far.

While you're at it, could you also add Watches for other resources that the controller creates and owns (e.g. Services)? If you search the code for `SetControllerReference`, you should find every kind of resource the controller creates and owns.

These Watches would be similar to what you've added already (`IsController: true, OwnerType: &rhjmcv1beta1.ContainerJFR{}`).

Alexjsenn:
Ok, I added one for services, secrets and routes, should I also add persistent volume claim? I'm not quite sure if that needs to be watched?

ebaron:
> Ok, I added one for services, secrets and routes, should I also add persistent volume claim? I'm not quite sure if that needs to be watched?

I would do PVC as well. Perhaps we could do this in a loop to cut down on the repetition? Just create a slice of `runtime.Object` with the different types objects, and iterate through them.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -95,38 +95,17 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	}
 
 	// Watch for changes to secondary resource Deployements and requeue the owner ContainerJFR
```
ebaron, line 97: I suppose now we should change this comment to just `secondary resources` now that it's more than just Deployments.

Alexjsenn: Done.

ebaron (approved):
Looks good to me! Nice work!

PR #128, opened 2020-11-12
Add report url to crd

Alexjsenn:
It seems to be working with crc, I was able to create a recording, see the report URL, and download it with CURL. There are some things that might need changing, like the way I repurposed `findDownloadURL` in `pkg/controller/recording/recording_controller.go`. I am not sure if returning three objects like that is recommended. I am also not sure about certain corner cases regarding if one of the URLs exists and the other doesn't.


pkg/controller/recording/recording_controller.go
```diff
@@ -305,48 +310,48 @@ func (r *ReconcileRecording) getFlightRecorder(ctx context.Context, recording *r
 	return jfr, nil
 }
 
-func findDownloadURL(cjfr jfrclient.ContainerJfrClient, filename string) (*string, error) {
+func findURLs(cjfr jfrclient.ContainerJfrClient, filename string) (*string, *string, error) {
```
ebaron, line 313: To make this more readable, let's change the signature to something like this:
```suggestion
func findSavedRecording(cjfr jfrclient.ContainerJfrClient, filename string) (*jfrclient.SavedRecording, error) {
```
Then just return the whole `SavedRecording` instead of the separate URLs. It could lead to bugs if we mix them up int the future.

pkg/controller/recording/recording_controller.go
```diff
@@ -305,48 +310,48 @@ func (r *ReconcileRecording) getFlightRecorder(ctx context.Context, recording *r
...
+			return &savedRecordings[idx].DownloadURL, &savedRecordings[idx].ReportURL, nil
 		}
 	}
-	return nil, nil
+	return nil, nil, nil
 }
 
 func archiveStoppedRecording(cjfr jfrclient.ContainerJfrClient, recording *rhjmcv1beta1.Recording,
-	target *jfrclient.TargetAddress) (*string, error) {
+	target *jfrclient.TargetAddress) (*string, *string, error) {
```
ebaron, line 329: We could return the whole `*jfrclient.SavedRecording` here too.

ebaron (commented):
Looks very good so far! I'll try it out tomorrow. In the meantime I have a couple of small suggestions below. Also it would be good to modify our tests to verify report URLs. You can run them with `make test-unit`.

Alexjsenn:
Ok, I modified the functions.

Alexjsenn:
Ill work on the tests now.

Alexjsenn:
Ok, I had some trouble figuring out what was the issue with the tests, but it seems to be working now!

ebaron (approved):
Looks good to me. Excellent work!

PR #129, opened 2020-11-19
Add event templates

### LINKED ISSUE #100: FlightRecorder should include event templates
Along with the list of event types, the FlightRecorder CR should include event templates provided by Container JFR's API.

Alexjsenn:
Fixes #100 
It seems to populate the template list correctly. There are just a few small details I wanted to ask about below.


pkg/apis/rhjmc/v1beta1/flightrecorder_types.go
```diff
@@ -58,6 +58,9 @@ type FlightRecorderStatus struct {
 	// Listing of events available in the target JVM
 	// +listType=set
 	Events []EventInfo `json:"events"`
+	// Listing of templates available in the target JVM
+	// +listType=set
```
Alexjsenn, line 62: What does +listType=set specify exactly?

pkg/controller/flightrecorder/flightrecorder_controller.go
```diff
@@ -171,6 +171,21 @@ func (r *ReconcileFlightRecorder) Reconcile(request reconcile.Request) (reconcil
...
+	log.Info("Listing templates for pod", "name", targetPod.Name, "namespace", targetPod.Namespace)
+	templates, err := cjfr.ListTemplates(targetAddr)
+	if err != nil {
+		log.Error(err, "failed to list templates")
+		return reconcile.Result{}, err
+	}
+
+	// Update Status with templates
+	instance.Status.Templates = templates
+	err = r.Client.Status().Update(ctx, instance)
```
Alexjsenn, line 184: Is there an issue with calling Status().Update twice (done above for event types)? It could be joined with the code above and then a single Update could be called, im not sure which is better.

pkg/apis/rhjmc/v1beta1/flightrecorder_types.go
andrewazores: https://github.com/kubernetes/kube-openapi/blob/master/pkg/idl/doc.go#L21

deploy/crds/rhjmc.redhat.com_flightrecorders_crd.yaml
```diff
@@ -368,10 +368,36 @@ spec:
...
                     type: string
                 type: object
+              templates:
+                description: Listing of templates available in the target JVM
+                items:
+                  description: TemplateInfo contains metadata for a JFR template
+                  properties:
+                    description:
+                      description: A description of the template and its performance
+                        impact
```
andrewazores, line 378: I would remove the "and its performance impact" here. The two templates included with recent OpenJDKs might happen to include this as part of the template description, but that isn't in any way guaranteed.

deploy/crds/rhjmc.redhat.com_flightrecorders_crd.yaml
```diff
@@ -368,10 +368,36 @@ spec:
...
+                      type: string
+                    name:
+                      description: The name of the template
+                      type: string
+                    provider:
+                      description: The organization which has provided the template
+                      type: string
+                    type:
+                      description: The type of template
```
andrewazores, line 387: The `type` field only has two possible values, `TARGET` and `CUSTOM`. These denote that the template either comes from the target JVM as a built-in preset template or that it was added to ContainerJFR (probably be the end user). I think this is worth documenting here, since the values are briefly enumerable and the meanings are well-defined.

pkg/apis/rhjmc/v1beta1/flightrecorder_types.go
ebaron: Thanks for bringing this up, actually. I recall us having trouble with a newer Operator SDK due to Events being a Set. I believe this was due to not having a key defined for comparison.

```
The CustomResourceDefinition "flightrecorders.rhjmc.redhat.com" is invalid: spec.validation.openAPIV3Schema.properties[status].properties[events].items.x-kubernetes-map-type: Invalid value: "null": must be atomic as item of a list with x-kubernetes-list-type=set 
```

I think both Events and Templates should be Atomic type instead. They aren't ever updated on a per-element basis, just the entire list updated at once. This would avoid the trouble of not providing a key field. I think it's fine to change both now, while you're making related changes.

pkg/controller/flightrecorder/flightrecorder_controller.go
ebaron: There shouldn't be an issue other than performance. The client we use reads from a cache, but writes directly to the cluster [1]. So this would mean two writes instead of one. I would remove the first status update, and write both events and templates at once. 

[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.6.1#hdr-Clients_and_Caches

deploy/crds/rhjmc.redhat.com_flightrecorders_crd.yaml
ebaron: In that case, we could make `type` an enum type like we do for `RecordingState` here:
https://github.com/rh-jmc-team/container-jfr-operator/blob/796a4f931f66b01bca513a0f33fc86b878a0d17e/pkg/apis/rhjmc/v1beta1/recording_types.go#L69-L86

Then we can also add OpenAPI validation on the field to enforce these values, like this:
https://github.com/rh-jmc-team/container-jfr-operator/blob/796a4f931f66b01bca513a0f33fc86b878a0d17e/pkg/apis/rhjmc/v1beta1/recording_types.go#L91-L94

ebaron (approved):
Looks good to me. Great job!

PR #132, opened 2020-11-23
Update container-jfr to 1.0.0-BETA1

andrewazores:


ebaron (approved):
I was just about to do the same. Thanks!

Issue #136, opened 2020-11-24
Add unit tests for Endpoints and ContainerJFR controllers


ebaron:
These two should be conceptually similar to the existing FlightRecorder and Recording controller tests, although without needing the Container JFR httptest server.

PR #137, opened 2020-11-27
Add unit tests for Containerjfr controller

Alexjsenn:
#136

### DIFF
--- a/pkg/controller/containerjfr/containerjfr_controller_test.go
+++ b/pkg/controller/containerjfr/containerjfr_controller_test.go
@@ -38,12 +38,15 @@ package containerjfr_test
 
 import (
 	"context"
-	"net/http"
+	"time"
 
 	certv1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
 	certMeta "github.com/jetstack/cert-manager/pkg/apis/meta/v1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	openshiftv1 "github.com/openshift/api/route/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -54,14 +57,13 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	rhjmcv1beta1 "github.com/rh-jmc-team/container-jfr-operator/pkg/apis/rhjmc/v1beta1"
 	"github.com/rh-jmc-team/container-jfr-operator/pkg/controller/containerjfr"
+	"github.com/rh-jmc-team/container-jfr-operator/pkg/controller/containerjfr/resource_definitions"
 	"github.com/rh-jmc-team/container-jfr-operator/test"
 )
 
 var _ = Describe("ContainerjfrController", func() {
 	var (
 		objs       []runtime.Object
-		handlers   []http.HandlerFunc
-		server     *test.ContainerJFRServer
 		client     client.Client
 		controller *containerjfr.ReconcileContainerJFR
 	)
@@ -71,7 +73,6 @@ var _ = Describe("ContainerjfrController", func() {
 		s := test.NewTestScheme()
 
 		client = fake.NewFakeClientWithScheme(s, objs...)
-		server = test.NewServer(client, handlers)
 		controller = &containerjfr.ReconcileContainerJFR{
 			Client:        client,
 			Scheme:        s,
@@ -79,15 +80,9 @@ var _ = Describe("ContainerjfrController", func() {
 		}
 	})
 
-	JustAfterEach(func() {
-		server.VerifyRequestsReceived(handlers)
-		server.Close()
-	})
-
 	BeforeEach(func() {
 		objs = []runtime.Object{
-			test.NewContainerJFR(), test.NewCACert(), test.NewTargetPod(),
-			test.NewContainerJFRService(), test.NewJMXAuthSecret(),
+			test.NewContainerJFR(),
 		}
 	})
 
@@ -99,26 +94,14 @@ var _ = Describe("ContainerjfrController", func() {
 		Context("after Containerjfr already reconciled successfully", func() {
 			It("should be idempotent", func() {
 				req := reconcile.Request{NamespacedName: types.NamespacedName{Name: "containerjfr", Namespace: "default"}}
-				result, err := controller.Reconcile(req)
-				Expect(err).ToNot(HaveOccurred())
-				//Expect(result).To(Equal(reconcile.Result{}))
+				ReconcileFully(client, *controller)
 
 				obj := &rhjmcv1beta1.ContainerJFR{}
-				err = client.Get(context.Background(), req.NamespacedName, obj)
+				err := client.Get(context.Background(), req.NamespacedName, obj)
 				Expect(err).ToNot(HaveOccurred())
 
-				caCert := &certv1.Certificate{}
-				err = client.Get(context.Background(), types.NamespacedName{Name: "containerjfr-ca", Namespace: "default"}, caCert)
-				Expect(err).ToNot(HaveOccurred())
-				caCert.Status.Conditions = append(caCert.Status.Conditions, certv1.CertificateCondition{
-					Type:   certv1.CertificateConditionReady,
-					Status: certMeta.ConditionTrue,
-				})
-				err = client.Status().Update(context.Background(), caCert)
-				Expect(err).ToNot(HaveOccurred())
-
-				// Reconcile same containerjfr again
-				result, err = controller.Reconcile(req)
+				// Reconcile again
+				result, err := controller.Reconcile(req)
 				Expect(err).ToNot(HaveOccurred())
 				Expect(result).To(Equal(reconcile.Result{}))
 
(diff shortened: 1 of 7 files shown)


pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+				caCert := &certv1.Certificate{}
+				err = client.Get(context.Background(), types.NamespacedName{Name: "containerjfr-ca", Namespace: "default"}, caCert)
+				Expect(err).ToNot(HaveOccurred())
+				caCert.Status.Conditions = append(caCert.Status.Conditions, certv1.CertificateCondition{
+					Type:   certv1.CertificateConditionReady,
+					Status: certMeta.ConditionTrue,
+				})
+				err = client.Status().Update(context.Background(), caCert)
+				Expect(err).ToNot(HaveOccurred())
```
ebaron, line 118: You'll have to do this as well for the other Certificates created by the operator: `containerjfr`, `containerjfr-grafana`.

Additionally, you'll have to create the associated Secrets as well, as cert-manager would. The secret names should match each Certificate's spec.SecretName property. For these tests, I think it will be sufficient to fill the Secrets with dummy values under the `corev1.TLSCertKey`, `corev1.TLSPrivateKeyKey` and `resource_definitions.CAKey` keys.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,134 @@
...
+	rhjmcv1beta1 "github.com/rh-jmc-team/container-jfr-operator/pkg/apis/rhjmc/v1beta1"
+	"github.com/rh-jmc-team/container-jfr-operator/pkg/controller/containerjfr"
+	"github.com/rh-jmc-team/container-jfr-operator/test"
+)
+
+var _ = Describe("ContainerjfrController", func() {
+	var (
+		objs       []runtime.Object
+		handlers   []http.HandlerFunc
+		server     *test.ContainerJFRServer
```
ebaron, line 64: I don't think you'll need these in this test.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+	JustAfterEach(func() {
+		server.VerifyRequestsReceived(handlers)
+		server.Close()
+	})
+
+	BeforeEach(func() {
+		objs = []runtime.Object{
+			test.NewContainerJFR(), test.NewCACert(), test.NewTargetPod(),
+			test.NewContainerJFRService(), test.NewJMXAuthSecret(),
```
ebaron, line 90: In the default case for this test, you'll want to ensure the controller can create all dependent objects. So we should remove the CACert, Service and Secret here. You can remove the TargetPod as well as it won't be needed.

ebaron (commented):
Hey Alex, you're off to a good start. Here are some comments to help get you unblocked.

Alexjsenn:
I stumbled upon a small discrepancy in the cjfr controller, where only some of the addresses in ServiceSpec are saved with the http/https prefix. I don't know if this breaks functionality or is just inconsistent, but its probably better to fix instead of changing the test to adapt to it.

ebaron:
This is not a bug, but I agree the naming is inconsistent. Core/CommandAddress is used as a hostname, while GrafanaAddress is used as a URL. I think we should change the name of these variables to Core/CommandHostname, and GrafanaURL.

Alexjsenn:
One of the things that it doesn't test for yet (apart from anything I missed) is that the controller successfully deletes resources of the other configuration, so for example if its minimal, it could make sure that existing non-minimal resources were correctly removed. I didn't know how encompassing these tests should be.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,563 @@
...
+				MakeCertificatesReady(client)
+				InitializeSecrets(client)
+
+				// Check for routes, ingress configuration needs to be added as each
+				// one is created so that they all reconcile successfully
+				result, err = controller.Reconcile(req)
+				IngressConfig(client, *controller, req, false)
+			})
+		})
+		Context("succesfully creates required resources", func() {
```
ebaron, line 123: I would combine this Context with the previous one, using this description. Same thing for the minimal case.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,563 @@
...
+				Expect(pvc.Spec.AccessModes).To(Equal(expectedPvc.Spec.AccessModes))
+				Expect(pvc.Spec.StorageClassName).To(Equal(expectedPvc.Spec.StorageClassName))
+
+				pvcStorage := pvc.Spec.Resources.Requests["storage"]
+				expectedPvcStorage := expectedPvc.Spec.Resources.Requests["storage"]
+				Expect(pvcStorage.Equal(expectedPvcStorage)).To(BeTrue())
+
+				// Check for owner
+				Expect(pvc.ObjectMeta.OwnerReferences[0].Kind).To(Equal("ContainerJFR"))
+				Expect(pvc.ObjectMeta.OwnerReferences[0].Name).To(Equal("containerjfr"))
```
ebaron, line 152: You could refactor this into a helper, along with other ObjectMeta checks. You should be able to use `metav1.Object` as the parameter type.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,563 @@
...
+				result, err := controller.Reconcile(req)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(result).To(Equal(reconcile.Result{}))
+			})
+		})
+
+	})
+})
+
+func NewFakeSecret(name string) *corev1.Secret {
```
ebaron, line 471: Please make the helper functions begin with a lower-case letter to make them private visibility.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,563 @@
...
+				Expect(pvc.Spec.AccessModes).To(Equal(expectedPvc.Spec.AccessModes))
+				Expect(pvc.Spec.StorageClassName).To(Equal(expectedPvc.Spec.StorageClassName))
+
+				pvcStorage := pvc.Spec.Resources.Requests["storage"]
+				expectedPvcStorage := expectedPvc.Spec.Resources.Requests["storage"]
+				Expect(pvcStorage.Equal(expectedPvcStorage)).To(BeTrue())
+
+				// Check for owner
+				Expect(pvc.ObjectMeta.OwnerReferences[0].Kind).To(Equal("ContainerJFR"))
+				Expect(pvc.ObjectMeta.OwnerReferences[0].Name).To(Equal("containerjfr"))
```
ebaron, line 315: If possible, could you refactor the specs (the body of the `It` calls) to helpers to reduce duplication between the minimal and non-minimal Contexts?

ebaron (commented):
Hey Alex, very nice job on the tests! I just have a few suggestions inline.

Alexjsenn:
Ok, I refactored all the tests that were duplicated.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -543,21 +287,193 @@ func IngressConfig(client client.Client, controller containerjfr.ReconcileContai
...
+	Expect(pvc.ObjectMeta.Name).To(Equal(expectedPvc.ObjectMeta.Name))
+	Expect(pvc.ObjectMeta.Namespace).To(Equal(expectedPvc.ObjectMeta.Namespace))
+	Expect(pvc.Spec.AccessModes).To(Equal(expectedPvc.Spec.AccessModes))
+	Expect(pvc.Spec.StorageClassName).To(Equal(expectedPvc.Spec.StorageClassName))
+
+	pvcStorage := pvc.Spec.Resources.Requests["storage"]
+	expectedPvcStorage := expectedPvc.Spec.Resources.Requests["storage"]
+	Expect(pvcStorage.Equal(expectedPvcStorage)).To(BeTrue())
+
+	checkOwner(&pvc.ObjectMeta)
```
ebaron, line 365: Does just passing `pvc` work here? Same for the other resource types.

Also I was thinking you could do like a `checkMetadata` (or similar), and check not only owner references, but Name, Namespace, Labels, Annotations, to further reduce duplication.

Alexjsenn: If I use metav1.Object then I have to pass in the metadata itself.

Alexjsenn: I refactored the objectMeta comparisons into checkMetadata.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,462 @@
...
+				err := client.Get(context.Background(), types.NamespacedName{Name: "containerjfr-grafana", Namespace: "default"}, service)
+				Expect(err).To(HaveOccurred())
+
+				reconcileFully(client, *controller, false)
+
+				err = client.Get(context.Background(), types.NamespacedName{Name: "containerjfr-grafana", Namespace: "default"}, service)
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedService := resource_definitions.NewGrafanaService(test.NewContainerJFR())
+				checkMetadata(&service.ObjectMeta, &expectedService.ObjectMeta)
```
ebaron, line 133:
```suggestion
				checkMetadata(service, expectedService)
```

ebaron: Just passing the objects works for me. Could you do this with the other calls to checkMetadata as well?

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -0,0 +1,462 @@
...
+	if minimal {
+		testContainer = test.NewMinimalContainerJFR()
+	} else {
+		testContainer = test.NewContainerJFR()
+	}
+	expectedDeployment := resource_definitions.NewDeploymentForCR(testContainer, &testSpecs, &testTLSConfig)
+	Expect(deployment.ObjectMeta.Name).To(Equal(expectedDeployment.ObjectMeta.Name))
+	Expect(deployment.ObjectMeta.Namespace).To(Equal(expectedDeployment.ObjectMeta.Namespace))
+	Expect(deployment.ObjectMeta.Labels).To(Equal(expectedDeployment.ObjectMeta.Labels))
+	Expect(deployment.ObjectMeta.Annotations).To(Equal(expectedDeployment.ObjectMeta.Annotations))
```
ebaron, line 433: This could be done with `checkMetadata` as well, I believe.

Alexjsenn: Yes, I missed that, I fix it!

pkg/controller/containerjfr/containerjfr_controller_test.go
Alexjsenn: Done!

ebaron:
Looking great! I checked the test coverage and see a couple more complex scenarios we should test. These are the transition between a minimal deployment and a non-minimal deployment, and vice-versa. We should check that the Grafana service and route are created or deleted, and that the deployment has the appropriate number of containers. You can trigger these scenarios by doing the following:
1. Reconcile until everything is successfully created.
2. Toggle the ContainerJFR object's `spec.minimal` property, and issue a `Client.Update`.
3. Reconcile again, and the controller should respond to the spec change.

You can check out the test coverage pretty easily like:
```
go test -coverprofile=coverage.out ./pkg/controller/containerjfr/
go tool cover -html coverage.out
```

Alexjsenn:
Ok, so I added those cases. Also thats a super useful tool to check what's being tested!!! Im wondering if its necessary to cover all the errors that can occurr, or is that all right?

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -201,7 +193,71 @@ var _ = Describe("ContainerjfrController", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+
+				req := reconcile.Request{NamespacedName: types.NamespacedName{Name: "containerjfr", Namespace: "default"}}
+				_, err = controller.Reconcile(req)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			It("should delete grafana resources", func() {
+				service := &corev1.Service{}
+				err := client.Get(context.Background(), types.NamespacedName{Name: "containerjfr-grafana", Namespace: "default"}, service)
+				Expect(err).To(HaveOccurred())
```
ebaron, line 251: To be sure that the error is due to the Service being deleted, you can check that it's a NotFound error with this function:
https://pkg.go.dev/k8s.io/apimachinery@v0.20.0/pkg/api/errors#IsNotFound

You can import this package as `kerrors` to avoid name collision with the standard errors package, as done elsewhere in the operator.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -201,7 +193,71 @@ var _ = Describe("ContainerjfrController", func() {
...
 
+				route := &openshiftv1.Route{}
+				err = client.Get(context.Background(), types.NamespacedName{Name: "containerjfr-grafana", Namespace: "default"}, route)
+				Expect(err).To(HaveOccurred())
```
ebaron, line 255: Same thing here.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -201,7 +193,71 @@ var _ = Describe("ContainerjfrController", func() {
...
 			})
 		})
+		Context("Switching from a minimal to a non-minimal deployment", func() {
+			BeforeEach(func() {
+				objs = []runtime.Object{
+					test.NewMinimalContainerJFR(),
+				}
+			})
+			JustBeforeEach(func() {
+				reconcileFully(client, *controller, true)
```
ebaron, line 203: Sorry I didn't notice this sooner, could you change your helper functions to take a pointer to `containerjfr.ReconcileContainerJFR`? This is more efficient, since otherwise a copy of the whole struct is created for each call. In general, you will want to pass structs by pointer, unless you're concerned about side-effects.

ebaron (commented):
Hey Alex, thanks for taking care of those new cases. Nearing the end now, just a couple of minor suggestions.

ebaron:
> Ok, so I added those cases. Also thats a super useful tool to check what's being tested!!! Im wondering if its necessary to cover all the errors that can occurr, or is that all right?

I think it's okay to skip these for this PR, some of them are difficult to test using the fake client. This will hopefully change in the near future: https://github.com/kubernetes-sigs/controller-runtime/issues/1048

pkg/controller/containerjfr/containerjfr_controller_test.go
Alexjsenn: Ah, this is much better, I added it to several places where I intended on checking that it wasn't found.

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -93,20 +94,20 @@ var _ = Describe("ContainerjfrController", func() {
...
 			})
 			It("should create persistent volume claim and set owner", func() {
-				expectPVC(client, *controller, false)
+				expectPVC(client, controller, false)
 			})
 			It("should create Grafana secret and set owner", func() {
 				secret := &corev1.Secret{}
 				err := client.Get(context.Background(), types.NamespacedName{Name: "containerjfr-grafana-basic", Namespace: "default"}, secret)
-				Expect(err).To(HaveOccurred())
+				Expect(kerrors.IsNotFound(err)).To(Equal(true))
```
ebaron, line 108: The `BeTrue` matcher is probably a better fit for these kind of expectations.
```suggestion
				Expect(kerrors.IsNotFound(err)).To(BeTrue())
```

ebaron (commented):
Just one last thing

pkg/controller/containerjfr/containerjfr_controller_test.go
Alexjsenn: Yes, definitely!

ebaron (approved):
Looks good to me! Great job Alex!

PR #138, opened 2020-12-10
add license header to suite_test file

Alexjsenn:
Forgot to add a license header to the suite file.


ebaron (approved):
I missed this as well. Thanks for noticing and fixing it.

PR #139, opened 2020-12-11
Add endpoint tests

### LINKED ISSUE #136: Add unit tests for Endpoints and ContainerJFR controllers

Alexjsenn:
The one main part its not covering is if the Service corresponding to the Endpoints is owned by ContainerJFR, I am a bit confused as to how the service gets created in the first place.  Fixes #136

### DIFF
--- a/pkg/controller/endpoints/endpoints_controller_test.go
+++ b/pkg/controller/endpoints/endpoints_controller_test.go
@@ -41,7 +41,10 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
@@ -78,7 +81,7 @@ var _ = Describe("EndpointsController", func() {
 	})
 
 	Describe("reconciling a request", func() {
-		Context("successfully reconcile", func() {
+		Context("successfully reconcile, without JMX auth", func() {
 			BeforeEach(func() {
 				objs = []runtime.Object{
 					test.NewContainerJFR(), test.NewContainerJFRService(),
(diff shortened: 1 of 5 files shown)


ebaron:
> The one main part its not covering is if the Service corresponding to the Endpoints is owned by ContainerJFR, I am a bit confused as to how the service gets created in the first place.

So normally this Service (and implicitly the Endpoints) would be created by the ContainerJFR controller. In this case, you would create it ahead of time as part of the test objects. There is already a `NewContainerJFRService` in test/resources.go that you can use, you'd just need to create a corresponding `NewContainerJFREndpoints` for it.

test/reconciler.go
```diff
@@ -59,6 +59,14 @@ type testClientFactory struct {
 	serverURL string
 }
 
+func NewServerlessTestReconciler(client client.Client) common.Reconciler {
```
ebaron, line 62: Let's use something like NoServer or WithoutServer instead of Serverless to avoid confusion with:
https://en.wikipedia.org/wiki/Serverless_computing

test/resources.go
```diff
@@ -116,9 +116,25 @@ func NewFlightRecorderBadJMXPassKey() *rhjmcv1beta1.FlightRecorder {
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "FlightRecorder",
+			APIVersion: "rhjmc.redhat.com/v1beta1",
+		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "test-pod",
-			Namespace: "default",
+			Name:            "test-pod",
+			Namespace:       "default",
+			ResourceVersion: "2",
```
ebaron, line 126: You should probably leave out ResourceVersion as this will vary by test, and compare Name, Namespace, Labels, OwnerReferences separately.

ebaron (commented):
Hey Alex, looking good so far. I have a couple of comments inline, and some suggestions for additional test cases:
- An Endpoints where there is no 'jfr-jmx' named port, but there is one with the 9091 port number. (Should create FlightRecorder)
- An Endpoints where there are neither 'jfr-jmx' or 9091 ports. (Should not create FlightRecorder)
- An Endpoints where there is no `TargetRef` defined. (Should not create FlightRecorder)

test/resources.go
Alexjsenn: Done!

test/reconciler.go
Alexjsenn: Makes sense!

Alexjsenn:
I added the test cases you mentioned, I tried reducing code duplication, but if you see better ways of refactoring the tests let me know!

pkg/controller/endpoints/endpoints_controller_test.go
```diff
@@ -0,0 +1,237 @@
...
+	Describe("reconciling a request", func() {
+		Context("successfully reconcile, without JMX auth", func() {
+			BeforeEach(func() {
+				objs = []runtime.Object{
+					test.NewContainerJFR(), test.NewContainerJFRService(),
+					test.NewTargetPod(), test.NewTestEndpoints(),
+				}
+			})
+			It("should create new flightrecorder", func() {
+				req := reconcile.Request{NamespacedName: types.NamespacedName{Name: "containerjfr", Namespace: "default"}}
```
ebaron, line 93: I think it makes sense to separate the Container JFR case from other Endpoints, because of the extra JMX auth steps involved and for general clarity.

For the typical case, we're reconciling Endpoints belonging to a user's application. Instead of using `test.NewContainerJFRService`, I would add a `test.NewTestService` that creates a Service called "test-svc". Likewise, rename the Endpoints created by `test.NewTestEndpoints` to "test-svc" (they have to be the same name).

Then for the Container JFR special case below (currently "with JMX auth"), use `test.NewContainerJFRService`, and create a `test.NewContainerJFREndpoints` and `test.NewContainerJFRPod` that create "containerjfr" and "containerjfr-pod" objects respectively.

I hope that all makes sense.

pkg/controller/endpoints/endpoints_controller_test.go
```diff
@@ -0,0 +1,237 @@
...
+				expected := test.NewFlightRecorderNoJMXAuth()
+				Expect(found.TypeMeta).To(Equal(expected.TypeMeta))
+				Expect(found.ObjectMeta.Name).To(Equal(expected.ObjectMeta.Name))
+				Expect(found.ObjectMeta.Namespace).To(Equal(expected.ObjectMeta.Namespace))
+				Expect(found.ObjectMeta.Labels).To(Equal(expected.ObjectMeta.Labels))
+				Expect(found.ObjectMeta.OwnerReferences).To(Equal(expected.ObjectMeta.OwnerReferences))
+				Expect(found.Spec).To(Equal(expected.Spec))
+			})
+		})
+		Context("successfully reconcile, with JMX auth", func() {
```
ebaron, line 111: Let's call this "successfully reconcile Container JFR", since it's a special case only for Container JFR itself.

ebaron (commented):
Thanks for implementing the changes, I have a few more recommendations below.

Alexjsenn:
Ok, I organized it more clearly into testing for the test-pod, and testing when it connects to itself!

ebaron (approved):
Looks good to go! Thanks, Alex!

PR #143, opened 2020-12-18
Use different subpaths within volume for recordings and templates

ebaron:
I noticed when archiving a recording, that the recording also ended up in the templates directory. This is because we use the same mounted volume for both recordings and templates, and by default it's the volume's root directory that is mounted. This fix specifies the `SubPath` property for each mount, which will mount different directories within the volume.

### DIFF
--- a/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
+++ b/pkg/controller/containerjfr/resource_definitions/resource_definitions.go
@@ -242,10 +242,12 @@ func NewCoreContainer(cr *rhjmcv1beta1.ContainerJFR, specs *ServiceSpecs, tls *T
 		{
 			Name:      cr.Name,
 			MountPath: "flightrecordings",
+			SubPath:   "flightrecordings",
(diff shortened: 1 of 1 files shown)


andrewazores:
https://github.com/rh-jmc-team/container-jfr/pull/356

There are also changes in this container-jfr PR for the default paths for things as well as the introduction of a new path for configuration files. Would it make sense to hold this PR until that one or some version of it goes in, so that the paths can be kept consistent upstream vs as deployed by the operator?

ebaron:
I think it would probably be good to get this in ASAP. I didn't mention in my summary, but this is breaking template functionality after you archive a recording. You'll see "Request failed (Status 500) java.text.ParseException: Content is not allowed in prolog.", when going to the Events tab or creating a new recording and selecting the Events dropdown.

Alexjsenn:
It fixes the prblem you mention!

Alexjsenn (approved):

PR #146, opened 2021-01-14
Update expected platform strategy package

andrewazores:
The class `OpenShiftPlatformStrategy` was moved in a recent upstream patch (rh-jmc-team/container-jfr#391), so this adjusts the environment variable that explicitly specifies this strategy should be selected. #393 is a backport of #391 for v1 and also moves the Strategy to the same location, so this applies for either version.


ebaron:
Makes sense. We should probably update the image tag to a new beta release that has https://github.com/rh-jmc-team/container-jfr/pull/393 in it, otherwise this won't deploy correctly.

andrewazores:
BETA4 PR here: https://github.com/rh-jmc-team/container-jfr/pull/394

And this PR has been updated to deploy that version.

ebaron (approved):
Thanks!

PR #147, opened 2021-01-21
Add list of certificate secrets to containerjfr CRD

### LINKED ISSUE #142: Add TrustedCertSecrets property to ContainerJFR CRD
We can add a property to the ContainerJFR CRD that allows the user to specify a list of certificate secrets that should be imported to the Container JFR truststore upon startup. When reconciling a ContainerJFR with an existing Deployment, we must check that the list of Secrets mounted to /truststore contains the following:
1. The self-signed ContainerJFR CA used by Container JFR itself
2. Exactly the same set of Secrets specified in the ContainerJFR's Spec.

If this check is not satisfied, we should modify the Deployment, which should redeploy Container JFR.

The format of the property can be a slice of something like the following type:
```go
// DefaultCertificateKey will be used when looking up the certificate within a secret,
// if a key is not manually specified
const DefaultCertificateKey = corev1.TLSCertKey

// CertificateSecret is a Secret containing a PEM-encoded X.509 certificate to be considered
// trusted by Container JFR
type CertificateSecret struct {
	// Name of secret in the local namespace
	SecretName string `json:"secretName"`
	// Key within secret containing the certificate, defaults to DefaultCertificateKey
	// +optional
	CertificateKey *string `json:"certificateKey,omitempty"`
}
```

Alexjsenn:
Fixes #142


Alexjsenn:
Its working with tls certificates. If the certificate is not supported by container jfr (or improperly formatted) then cjfr fails to import it from the truststore.

Alexjsenn:
Something else I am not sure about is wether the certificateKey field should be optional. If we do want it to be optional then I probably need to add a check to see if its null, in which case there'd be no subpath in the volumemount.

ebaron:
> Something else I am not sure about is wether the certificateKey field should be optional. If we do want it to be optional then I probably need to add a check to see if its null, in which case there'd be no subpath in the volumemount.

If not given, we should use the default value specified in #142.

pkg/apis/rhjmc/v1beta1/containerjfr_types.go
```diff
@@ -43,7 +43,8 @@ import (
 // ContainerJFRSpec defines the desired state of ContainerJFR
 // +k8s:openapi-gen=true
 type ContainerJFRSpec struct {
-	Minimal bool `json:"minimal"`
+	Minimal bool                `json:"minimal"`
+	Secrets []CertificateSecret `json:"secrets"`
```
ebaron, line 47: Let's call this `TrustedCertSecrets` to be more specific about the intended use. We should also add a comment here to document what the secrets are for.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -277,6 +277,36 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	// Check that secrets mounted in /truststore conincide with CRD
```
ebaron, line 280: s/conincide/coincide

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -277,6 +277,36 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
...
+	err = r.Client.Get(context.Background(), types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}, deployment)
+	if err == nil {
+		deploymentSecrets := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
+		secrets := instance.Spec.Secrets
+		for i := 0; i < len(secrets); i++ {
```
ebaron, line 285: Typically in Go, you'll see for-each loops of the form `for _, secret := range secrets`. I'd recommend that syntax instead. Same thing for other loops in this PR.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -277,6 +277,36 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
...
+					return reconcile.Result{}, err
+				}
+
+				err = r.Client.Update(context.Background(), deployment)
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+				break
+			}
+		}
```
ebaron, line 307: This doesn't seem to capture the case when a secret is removed from the ContainerJFR CR. How about calling `resources.NewDeploymentForCR` and just check that the mounts are identical to the existing deployment? If for some reason the user changes the ordering of the mounts without otherwise changing the content, we'll trigger a redeployment, but that seems like an uncommon scenario.

pkg/controller/containerjfr/resource_definitions/resource_definitions.go
```diff
@@ -287,10 +301,23 @@ func NewCoreContainer(cr *rhjmcv1beta1.ContainerJFR, specs *ServiceSpecs, tls *T
 			SubPath:   CAKey,
 			ReadOnly:  true,
 		}
-		// TODO add mechanism to add other certs to /truststore
 
 		mounts = append(mounts, keystoreMount, caCertMount)
 
+		// TODO add mechanism to add other certs to /truststore
```
ebaron, line 307: This TODO can be removed as it's fixed by this PR.

ebaron (commented):
Nice job Alex, some suggestions below.

ebaron (commented):
We should also have some tests for these changes.

pkg/controller/containerjfr/containerjfr_controller.go
Alexjsenn: Im not sure what s/coincide/coincide means?

ebaron: Spelling mistake, change conincide to coincide. It's `sed` syntax :smile:

deploy/crds/rhjmc.redhat.com_v1beta1_containerjfr_cr.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
   name: containerjfr
 spec:
   minimal: false
+  secrets: []
```
ebaron, line 7: This should be changed to match the `trustedCertSecrets` name.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -277,6 +277,25 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
...
 
+	// Check that secrets mounted in /truststore coincide with CRD
+	err = r.Client.Get(context.Background(), types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}, deployment)
+	if err == nil {
+		deploymentMounts := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
+		expectedMounts := resources.NewDeploymentForCR(instance, serviceSpecs, tlsConfig).Spec.Template.Spec.Containers[0].VolumeMounts
+		if !equals(deploymentMounts, expectedMounts) {
+			reqLogger.Info("cert secrets mounted do not conicide with those specified in CRD, modifying deployment")
+			// Modify deployment
+			deployment := resources.NewDeploymentForCR(instance, serviceSpecs, tlsConfig)
```
ebaron, line 288: I would just do `deployment.Spec.Template.Spec.Containers[0].VolumeMounts = expectedMounts` here instead of replacing the whole deployment.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -277,6 +277,25 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, err
 	}
 
+	// Check that secrets mounted in /truststore coincide with CRD
+	err = r.Client.Get(context.Background(), types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}, deployment)
+	if err == nil {
+		deploymentMounts := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
+		expectedMounts := resources.NewDeploymentForCR(instance, serviceSpecs, tlsConfig).Spec.Template.Spec.Containers[0].VolumeMounts
+		if !equals(deploymentMounts, expectedMounts) {
+			reqLogger.Info("cert secrets mounted do not conicide with those specified in CRD, modifying deployment")
```
ebaron, line 286: Spelling: coincide

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -376,3 +395,15 @@ func (r *ReconcileContainerJFR) createObjectIfNotExists(ctx context.Context, ns
...
+	if len(a) != len(b) {
+		return false
+	}
+	for i, val := range a {
+		if val != b[i] {
+			return false
+		}
+	}
+	return true
+}
```
ebaron, line 409: Let's use `cmp.Equal` instead of this function. It has some extra handling for pointer fields in structs, of which there's one in VolumeMount. It's also a transitive dependency of ours already, so we're not adding anything new.

ebaron (commented):
Thanks for taking care of those issues. Here's a few more suggestions. Looking good!

pkg/controller/containerjfr/containerjfr_controller.go
Alexjsenn: Ah my bad, didn't know we had that!

pkg/controller/containerjfr/containerjfr_controller.go
Alexjsenn: Wow seems I can't spell it lol!

pkg/controller/containerjfr/containerjfr_controller.go
ebaron: It's new to me too :smile:

Alexjsenn:
So I just fixes those issues, and I added two tests, I can't seem to figure out why the second one is not working (it's the one that actually tests if the reconciler works). Im thinking it might be an issue with changing the runtime objects mid test? If you think of anything let me know!

pkg/controller/containerjfr/containerjfr_controller_test.go
```diff
@@ -259,6 +259,61 @@ var _ = Describe("ContainerjfrController", func() {
...
+						test.NewContainerJFR(),
+					}
+				})
+				JustBeforeEach(func() {
+					reconcileFully(client, controller, false)
+				})
+				It("Should update the corresponding deployment", func() {
+					objs = []runtime.Object{
+						test.NewContainerJFRWithSecrets(), newFakeSecret("testCert1"), newFakeSecret("testCert2"),
+					}
```
ebaron, line 297: Modifying `objs` within the It block has no effect because the fake client has already been initialized. You could do this instead:
```suggestion
					err := client.Update(context.Background(), test.NewContainerJFRWithSecrets())
					Expect(err).ToNot(HaveOccurred())
```

This should work because `test.NewContainerJFR` and `test.NewContainerJFRWithSecrets` have the same name.

You might want to move the `newFakeSecret` calls to where you initialize `objs` in the `BeforeEach` block.

ebaron (commented):
Looking at the tests again, I think it would be more valuable if we specify exactly what we expect the VolumeMounts to be. As it is now, we have no way of knowing that `NewDeploymentForCR` is creating them properly and handling the default key as expected. You could add a `NewVolumeMountsWithSecrets` (maybe in test/resources.go) that have the exact VolumeMounts you expect to see produced for `test.NewContainerJFRWithSecrets`. Then use `NewVolumeMountsWithSecrets` as the expected value to compare against the VolumeMounts created by the reconciler.

Alexjsenn:
> Looking at the tests again, I think it would be more valuable if we specify exactly what we expect the VolumeMounts to be. As it is now, we have no way of knowing that `NewDeploymentForCR` is creating them properly and handling the default key as expected. You could add a `NewVolumeMountsWithSecrets` (maybe in test/resources.go) that have the exact VolumeMounts you expect to see produced for `test.NewContainerJFRWithSecrets`. Then use `NewVolumeMountsWithSecrets` as the expected value to compare against the VolumeMounts created by the reconciler.

Yeah that makes sense!

deploy/crds/rhjmc.redhat.com_v1beta1_containerjfr_cr.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
   name: containerjfr
 spec:
   minimal: false
+  trustedCertSecrets: []
+
+
```
ebaron, line 9: This trailing whitespace seems to have snuck in.

test/resources.go
```diff
@@ -65,13 +65,36 @@ func NewTestScheme() *runtime.Scheme {
...
+		},
+		Spec: rhjmcv1beta1.ContainerJFRSpec{
+			Minimal:            false,
+			TrustedCertSecrets: []rhjmcv1beta1.CertificateSecret{},
+		},
+	}
+}
+
+func NewContainerJFRWithSecrets() *rhjmcv1beta1.ContainerJFR {
+	key := "tls.crt"
```
ebaron, line 81: Let's use a different key name here. Just to differentiate from the default key. It'll make for a more robust test.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -277,6 +278,25 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
...
 
+	// Check that secrets mounted in /truststore coincide with CRD
+	err = r.Client.Get(context.Background(), types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}, deployment)
+	if err == nil {
+		deploymentMounts := deployment.Spec.Template.Spec.Containers[0].VolumeMounts
+		expectedMounts := resources.NewDeploymentForCR(instance, serviceSpecs, tlsConfig).Spec.Template.Spec.Containers[0].VolumeMounts
+		if !cmp.Equal(deploymentMounts, expectedMounts) {
+			reqLogger.Info("cert secrets mounted do not coincide with those specified in CRD, modifying deployment")
+			// Modify deployment
+			deployment.Spec.Template.Spec.Containers[0].VolumeMounts = expectedMounts
```
ebaron, line 289: Sorry, I overlooked that we need to update the Volumes as well as VolumeMounts here. Otherwise any new mounts we define will be missing the Volumes.

pkg/controller/containerjfr/containerjfr_controller.go
```diff
@@ -277,6 +278,25 @@ func (r *ReconcileContainerJFR) Reconcile(request reconcile.Request) (reconcile.
...
+			if err := controllerutil.SetControllerReference(instance, deployment, r.Scheme); err != nil {
+				return reconcile.Result{}, err
+			}
```
ebaron, line 292: We can get rid of this now because we're no longer recreating the whole deployment. So it will already be owned by our CR.

pkg/controller/containerjfr/resource_definitions/resource_definitions.go
```diff
@@ -287,10 +301,28 @@ func NewCoreContainer(cr *rhjmcv1beta1.ContainerJFR, specs *ServiceSpecs, tls *T
...
+		for _, secret := range cr.Spec.TrustedCertSecrets {
+			var key string
+			if secret.CertificateKey != nil {
+				key = *secret.CertificateKey
+			} else {
+				key = rhjmcv1beta1.DefaultCertificateKey
+			}
+			mount := corev1.VolumeMount{
+				Name:      secret.SecretName,
+				MountPath: fmt.Sprintf("/truststore/%s-%s", secret.SecretName, key),
```
ebaron, line 317: I realized there's still potential for naming collision here. Say we have two secrets:
1. foo-bar, with key baz
2. foo, with key bar-baz
These are both created as foo-bar-baz.

Underscores aren't valid in Kubernetes object names. So I think we should be okay if we use an underscore instead of a hyphen as a separator between the secret name and key when mounting the file.

Alexjsenn: Oh okok, yeah that would prevent that kind of collision!

ebaron (commented):
When building, I see go-cmp is now added to our go.mod. We should commit this change with this PR. This seems to be the version already used by cert-manager.

```diff
diff --git a/go.mod b/go.mod
index 5ebbf3b..98dcc3a 100644
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,7 @@ module github.com/rh-jmc-team/container-jfr-operator
 
 require (
        github.com/go-logr/logr v0.2.1-0.20200730175230-ee2de8da5be6
+       github.com/google/go-cmp v0.4.1
        github.com/jetstack/cert-manager v1.0.2
        github.com/onsi/ginkgo v1.12.1
        github.com/onsi/gomega v1.10.1
```

Alexjsenn:
Yes, my bad, for some reason I lost it in the midst of the other temporary files I didn't want to commit!

ebaron (approved):
Good to go now. Great job!

Issue #122, opened 2020-10-19
Grafana not reconfigured when pod is killed


ebaron:
The Grafana controller watches services, and when the Container JFR pod is killed and recreated, the service does not receive an update event. This means the reconciler is never invoked to configure the new Grafana container.

andrewazores:
#110 is generally related to this as well

PR #151, opened 2021-02-04
Configure datasource in Grafana image

ebaron:
This PR modifies our customized Grafana image to configure the jfr-datasource datasource as part of the image. This adds a slight wrapper around the default entrypoint that substitutes the `JFR_DATASOURCE_URL` environment variable into the datasources.yaml file. It uses a default value of `http://0.0.0.0:8080`, which is the default for jfr-datasource. For the operator we override this to `http://127.0.0.1:8080` to restrict traffic to within the pod.

This allows us to get rid of the Grafana controller and fix the bug where the datasource would not be reconfigured if the pod was killed.

I haven't pushed the revised image listed in this PR yet, but you can use `quay.io/ebaron/container-jfr-grafana-dashboard:1.0.0-BETA1` to test.

Fixes: #122


andrewazores:
Makes sense, and of course I really like to see things being done as much at image creation time or by the entrypoint than by something like an external controller.

Since the Grafana base image simply places its entrypoint at `/run.sh`, might it make sense to mimic this and place the new entrypoint at `/cjfr-run.sh` (rather than in `/usr/bin`)? I'm fine with it either way however.

andrewazores (dismissed):

ebaron:
Sure, that makes sense. Just need a quick re-review now.

andrewazores (approved):

ebaron:
Image is pushed to `quay.io/rh-jmc-team/container-jfr-grafana-dashboard:1.0.0-BETA1`

Issue #140, opened 2020-12-13
ConsoleLink CR


andrewazores:
https://github.com/openshift/api/blob/master/console/v1/types_console_link.go

PR #154, opened 2021-02-11
Add ConsoleLink when running on OpenShift

ebaron:
This is a bit more complicated than you would expect because we're (currently) a namespace-scoped operator and the ConsoleLink CRD is cluster-scoped. This means we have to use ClusterRole/Bindings, and cannot rely on Kubernetes GC to remove the link when the operator is deleted.

The Makefile is modified to query the current context's namespace, and substitutes that into the ClusterRoleBinding in order to locate the ServiceAccount we create. The version of the OpenShift API is bumped as well since the version currently pulled in doesn't have support for ConsoleLinks in the namespace dashboard area. In the Administrator view, you should see the link appear when navigating to the ~Home -> Overview~ Home -> Project -> (project name) pane on a namespace that has the operator deployed. In the Developer view, you can find it in the Project tab.

Since ConsoleLinks are cluster-scoped, I used the GenerateName option to avoid potential naming conflicts with other namespaces. The mapping between Container JFR instances and ConsoleLinks is done using a pair of labels. In order to delete the ConsoleLink when the ContainerJFR CR is deleted, I added a finalizer. This works similarly to how we use finalizers for Recordings. I also did a bit of refactoring to avoid duplication of finalizer-related functionality.

Fixes: #140


Alexjsenn:
When I make image it modifies deploy/crds/rhjmc.redhat.com_containerjfrs_crd.yaml (it removes line 13 & 14), not sure if that's because of some difference on my machine or if it should be changed.

ebaron:
> When I make image it modifies deploy/crds/rhjmc.redhat.com_containerjfrs_crd.yaml (it removes line 13 & 14), not sure if that's because of some difference on my machine or if it should be changed.

I think this is an unrelated side-effect of the Operator SDK upgrade.

Alexjsenn:
Ok, no prooblem, it looks good, works as expected!

Alexjsenn (approved):

PR #155, opened 2021-02-11
Fix error where datasource input can't be deleted

ebaron:
There's a problem with the Grafana dashboard image when deployed on OpenShift. The container is run with a random UID (not `grafana`) and GID 0. This was causing it to crash due to being unable to delete the `datasource.yaml.in` file. This PR changes ownership of our datasource/dashboard files to have GID 0. This allows them to be edited in OpenShift, as they would be with Podman. I also changed the input file in /tmp to the actual destination, and just use in-place sed. This was because /tmp's sticky bit still prevented deletion, even with the GID fix.


andrewazores (approved):

ebaron:
Image pushed to `quay.io/rh-jmc-team/container-jfr-grafana-dashboard:1.0.0-BETA2`

PR #157, opened 2021-02-22
Update to 1.0.0-BETA5

ebaron:
This updates the Container JFR image and operator version to 1.0.0-BETA5. I'll tag this commit as well. This should provide a stable option for users while we make changes leading to the 1.0 release. The catalog source will still pull in the old 0.5.0 version, but it should work. We don't want to release a new bundle until we're certain the v1beta1 API is settled.


andrewazores (approved):

ebaron:
Pushed image: `quay.io/rh-jmc-team/container-jfr-operator:1.0.0-BETA5` and `quay.io/rh-jmc-team/container-jfr-operator:latest`.

PR #160, opened 2021-02-23
Work with Kubernetes 

### LINKED ISSUE #120: Work with Kubernetes
We should detect whether the operator is running on OpenShift or Kubernetes and it should run properly in both scenarios. This will mainly mean using Ingress for Kubernetes and Route for OpenShift.

Alexjsenn:
Fixes #120


Alexjsenn:
Some things to consider:

- Should the Ingress configuration be included (or customizable)?
  - Should the Ingress configuration include NGINX specific annotations? This might be a problem in deployments that aren't in minkube.
  - The Ingress host most likely needs at least a configurable host if the user were to want to use DNS/ use a different host?
- I haven't tried with the changes in #166  but given that Minikube does not offer dynamic provisioning (and that PR only adds customization to persistent volume claims) I still think a persistent volume needs to be created beforehand (like pv-volume.yaml). could/should this be automated?

Alexjsenn:
I think it is pretty much ready for review. A few things that I am not sure how to deal with:
- What to do with pv-volume.yaml
- I should remove the ingress config from the yaml file (I have it there so you can corroborate its working) but im not sure if it can be includeded somewhere else of if I should just delete it
- Should more tests be added for a kube environment?

ebaron (commented):
Hey Alex, looking good so far. I have some answers to your latest questions and some minor comments below, but I think you should rebase this PR before proceeding too far with the review. A lot has changed with the boilerplate code and file location with the SDK upgrade (#153).

> I think it is pretty much ready for review. A few things that I am not sure how to deal with:
> 
>     * What to do with pv-volume.yaml

This should no longer be required with #166. I did a quick test with Minikube and the PVC is bound without any extra steps. FWIW, my Minikube has the default-storageclass and storage-provisioner addons enabled, which might be required. I don't think I enabled these manually though.

> 
>     * I should remove the ingress config from the yaml file (I have it there so you can corroborate its working) but im not sure if it can be includeded somewhere else of if I should just delete it

I think they should probably be removed. So far, the YAML file represents the default install scenario. By default, we want the user to define their ingress specs because we can't yet make any default ingresses that work.

> 
>     * Should more tests be added for a kube environment?

Some more tests would be good. Some ideas:
- What happens in Kubernetes when no Ingress config is supplied
- Ensure Routes are not created in Kubernetes, Ingress is not created in OpenShift
- Check what happens when Ingress configs are nil, and when Ingress config is non-nil, but spec is nil

ebaron (commented):
I think your hanging problem on `make destroy_containerjfr_cr` is happening because this section needs to be special-cased.

https://github.com/rh-jmc-team/container-jfr-operator/blob/b2c2dacb0baa0a4dba129fcf1a669334c93b6f0f/controllers/containerjfr_controller.go#L112-L120

The `deleteConsoleLinks` call should be conditional on `IsOpenShift`. The finalizer removal needs to happen either way though, so perhaps move the `RemoveFinalizer` call from `deleteConsoleLinks` to just after this new conditional branch.

Alexjsenn:
It now works as intended, it can be tested by adding the following to the sample containerjfr.yaml

ingressOptions:
    exporterConfig:
      annotations: 
        nginx.ingress.kubernetes.io/backend-protocol : HTTPS
      spec:
        rules:
        - host: testing.containerjfr
          http:
            paths:
            - path: /
              pathType: Prefix
              backend:
                service:
                  name: containerjfr-sample
                  port:
                    number: 8181
    commandConfig:
      annotations: 
        nginx.ingress.kubernetes.io/backend-protocol : HTTPS
      spec:
        rules:
        - host: testing.containerjfr-command
          http:
            paths:
            - path: /
              pathType: Prefix
              backend:
                service:
                  name: containerjfr-sample-command
                  port:
                    number: 9090

And of course to test that it works, the output of minikube ip needs to be added to /etc/hosts as follows:
[minikube ip] testing.containerjfr testing.containerjfr-command

api/v1beta1/containerjfr_types.go
```diff
@@ -50,6 +51,9 @@ type ContainerJFRSpec struct {
 	// Options to customize the storage for Flight Recordings and Templates
 	// +optional
 	StorageOptions *StorageConfiguration `json:"storageOptions,omitempty"`
+	// Ingress specifications when running outside of OpenShift environment
+	// +optional
+	IngressOptions *IngressConfigurationList `json:"ingressOptions,omitempty"`
```
ebaron, line 56: What if we called this `NetworkOptions` or something similar, and rename the types accordingly. I think a logical next step might be to expand what is currently the `IngressConfiguration` type to allow the user to specify a RouteSpec to customize the Routes we create when running on OpenShift. So calling it `NetworkConfiguration` would be more generic for this purpose.

controllers/containerjfr_controller.go
```diff
@@ -113,7 +116,13 @@ func (r *ContainerJFRReconciler) Reconcile(ctx context.Context, request ctrl.Req
 	// Check if this Recording is being deleted
 	if instance.GetDeletionTimestamp() != nil {
 		if controllerutil.ContainsFinalizer(instance, cjfrFinalizer) {
-			return r.deleteConsoleLinks(context.Background(), instance)
+			if r.IsOpenShift {
+				r.deleteConsoleLinks(context.Background(), instance)
```
ebaron, line 120: We should check for an error returned from deleteConsoleLinks here since we're no longer returning its return values.

controllers/containerjfr_controller.go
```diff
@@ -389,6 +436,52 @@ func (r *ContainerJFRReconciler) createRouteForService(controller *rhjmcv1beta1.
...
 
+func (r *ContainerJFRReconciler) createIngressForService(controller *rhjmcv1beta1.ContainerJFR, svc *corev1.Service, exposePort corev1.ServicePort,
+	tlsConfig *openshiftv1.TLSConfig, ingressConfig *rhjmcv1beta1.IngressConfiguration) (string, error) {
+	logger := log.WithValues("Request.Namespace", svc.Namespace, "Name", svc.Name, "Kind", fmt.Sprintf("%T", &netv1.Ingress{}))
+	/*
+		tls := netv1.IngressTLS{}
+		if tlsConfig.DestinationCACertificate != "" {
+			tls.SecretName = tlsConfig.CACertificate
+		}
+	*/
```
ebaron, line 447: Let's remove this comment.

controllers/containerjfr_controller.go
```diff
@@ -424,27 +517,33 @@ func (r *ContainerJFRReconciler) getConsoleLinks(cr *rhjmcv1beta1.ContainerJFR)
...
 		reqLogger.Info("deleted ConsoleLink", "linkName", link.Name)
 	}
+	return nil
+}
 
-	// Remove finalizer upon success
-	err = common.RemoveFinalizer(ctx, r.Client, cr, cjfrFinalizer)
-	if err != nil {
-		return reconcile.Result{}, err
+func getIngressCofig(controller *rhjmcv1beta1.ContainerJFR, svc *corev1.Service) (*rhjmcv1beta1.IngressConfiguration, error) {
```
ebaron, line 539: typo: getIngressCofig

controllers/containerjfr_controller.go
```diff
@@ -330,7 +360,24 @@ func (r *ContainerJFRReconciler) createService(ctx context.Context, controller *
...
+			ingressConfig, err := getIngressCofig(controller, svc)
+			if err != nil {
+				return "", err
+			}
+			if ingressConfig == nil {
+				return "", nil
+			}
+			if ingressConfig.IngressSpec == nil {
+				return "", nil
+			}
```
ebaron, line 378: You could join these checks with a `||`.

ebaron (commented):
Hey Alex, very nice work so far. Here are a few things I've come across so far.

bundle/manifests/container-jfr-operator.clusterserviceversion.yaml
```diff
@@ -126,7 +126,7 @@ spec:
                 - --upstream=http://127.0.0.1:8080/
                 - --logtostderr=true
                 - --v=10
-                image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
+                image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
```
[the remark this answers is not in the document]
Alexjsenn: I can definitely remove it, but I haven't figured out how to make it run in Minikube without this fix. I just tried removing it and I get the same error again (Error: container has runAsNonRoot and image will run as root)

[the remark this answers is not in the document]
andrewazores: That's interesting. I wonder if this is a CRC/Minikube difference or if this is more fundamental to OpenShift/Kubernetes.

controllers/containerjfr_controller.go
Alexjsenn: you're right, I made it return an error on purpose but never checked it!

api/v1beta1/containerjfr_types.go
Alexjsenn: Makes sense!

api/v1beta1/containerjfr_types.go
```diff
@@ -53,7 +53,7 @@ type ContainerJFRSpec struct {
 	StorageOptions *StorageConfiguration `json:"storageOptions,omitempty"`
 	// Ingress specifications when running outside of OpenShift environment
 	// +optional
-	IngressOptions *IngressConfigurationList `json:"ingressOptions,omitempty"`
+	NetworkOptions *NetworkConfigurationList `json:"networkOptions,omitempty"`
```
ebaron, line 56: The comment here could use an update now. Something generic like "Options to control how the operator exposes the application over a network".

bundle/manifests/container-jfr-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
ebaron: I was hoping we'd have a workaround for this by now. We may have to look into disabling the metrics endpoint temporarily, which would allow us to remove kube-rbac-proxy for now.

api/v1beta1/containerjfr_types.go
Alexjsenn: Right, I forgot to make it less specific for the future!

controllers/containerjfr_controller.go
```diff
@@ -382,6 +429,50 @@ func (r *ContainerJFRReconciler) createRouteForService(controller *rhjmcv1beta1.
...
+	}
+
+	logger.Info("Ingress created", "Service.Status", fmt.Sprintf("%#v", found.Status))
+	host := ""
+	if networkConfig.IngressSpec.Rules != nil && networkConfig.IngressSpec.Rules[0].Host != "" {
+		host = networkConfig.IngressSpec.Rules[0].Host
+	}
+
+	return &url.URL{
+		Scheme: getProtocol(tlsConfig),
```
Alexjsenn, line 471: This is one thing that I am not sure about. Is it relevant to get the protocol from the tlsConfig at all, or should it just be hardocoded? I was looking if it can be specified in the ingressSpecs, bit in the ingressRule it says only HTTP is supported so would that just be hardcoded into the Scheme field?

ebaron: I think the protocol is determined by the ingress's spec.tls property. It's probably good enough to use `https` if spec.tls is non-empty.

controllers/containerjfr_controller.go
```diff
@@ -328,7 +361,21 @@ func (r *ContainerJFRReconciler) createService(ctx context.Context, controller *
 			InsecureEdgeTerminationPolicy: openshiftv1.InsecureEdgeTerminationPolicyRedirect,
 		}
 	}
-	return r.createRouteForService(controller, svc, *exposePort, tlsConfig)
+	if r.IsOpenShift {
+		return r.createRouteForService(controller, svc, *exposePort, tlsConfig)
+	} else {
+		if controller.Spec.NetworkOptions == nil {
+			return &url.URL{Host: ""}, nil
```
ebaron, line 368: I would return `nil, nil` here, then add a nil-check when this URL is to be used. e.g. Only add the respective environment variables (`CONTAINER_JFR_WEB_HOST`, `GRAFANA_DASHBOARD_URL`, etc.) if the URLs are non-nil.

controllers/containerjfr_controller.go
```diff
@@ -328,7 +361,21 @@ func (r *ContainerJFRReconciler) createService(ctx context.Context, controller *
...
+		}
+		networkConfig, err := getNetworkConfig(controller, svc)
+		if err != nil {
+			return nil, err
+		}
+		if networkConfig == nil || networkConfig.IngressSpec == nil {
+			return &url.URL{Host: ""}, nil
```
ebaron, line 375: Same thing here.

ebaron (commented):
Hey Alex, thanks for taking care of another rebase. Please see the comments below.

bundle/manifests/container-jfr-operator.clusterserviceversion.yaml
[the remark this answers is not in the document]
ebaron: I'll submit a PR to disable the metrics endpoint and kube-rbac-proxy until #180 is fixed.

ebaron (approved):
Works very well. Excellent work Alex!

PR #166, opened 2021-02-25
Add StorageOptions to ContainerJFR CRD 

### LINKED ISSUE #159: Support Dynamic Provisioning for PVC
From: https://kubernetes.io/blog/2017/03/dynamic-provisioning-and-storage-classes-kubernetes/
> If storageClassName is not specified in the PVC, the default storage class will be used for provisioning.
>    Existing, “Available”, PVs that do not have the default storage class label will not be considered for binding to the PVC
>
> If storageClassName is set to an empty string (‘’) in the PVC, no storage class will be used (i.e.; dynamic provisioning is disabled for this PVC)
>    Existing, “Available”, PVs (that do not have a specified storageClassName) will be considered for binding to the PVC

We currently set the `storageClassName` for our PVC to the empty string, which works well for testing in CRC since it doesn't use Dynamic Provisioning out of the box. For real world clusters though, it might make sense to not specify `storageClassName`, deferring to the default StorageClass.

If this is not sufficient, we may want to add a `storageClassName` to the ContainerJFR CRD spec, to allow the user to specify which class they want used for our PVC.

For our CRC testing setups there are some options for Dynamic Provisioning here:
https://github.com/code-ready/crc/wiki/Dynamic-volume-provisioning

ebaron:
This PR adds a `StorageOptions` field to the ContainerJFR CRD. At the moment this contains just a PVC spec, but I used a separate type to make future expansion less likely to break things. With this PVC spec, users can customize the PVC that the operator creates. The key features I have in mind is changing the StorageClass and altering the resource usage.

This also removes the `storageClassName: ""` property from the PVC spec by default. This allows the PVC to work with Dynamic Provisioning [1]. I added an environment variable `PVC_STATIC` to patch this empty storageClassName into the CR, but upon testing in CRC it appears to be unneeded. I since removed the environment variable. If it works in CRC with storageClassName undefined, then I can't think of a need for the environment variable.

Fixes #159 

[1] https://v1-19.docs.kubernetes.io/docs/concepts/storage/persistent-volumes/#provisioning


deploy/crds/rhjmc.redhat.com_containerjfrs_crd.yaml
```diff
@@ -68,6 +66,142 @@ spec:
...
+              storageOptions:
+                description: Options to customize the storage for Flight Recordings
+                  and Templates
+                properties:
+                  pvcSpec:
+                    description: Spec for a Persistent Volume Claim, whose options
+                      will override the defaults used by the operator. Unless overriden,
+                      the PVC will be created with the default Storage Class and 500MiB
+                      of storage. Once the operator has created the PVC, changes to
+                      this field have no effect.
```
andrewazores, line 78: This is almost certainly something for a future enhancement, but I think it would be best if changing this field resulted in the Operator migrating the data from the old PVC to a new one matching the new spec, and then removing the old PVC. Otherwise, end users will end up having to do something like download every recording and template locally, undeploy the operator, redeploy it with a patched pvcSpec, and then re-upload everything.

andrewazores:
Also, should `deploy/crds/rhjmc.redhat.com_v1beta1_containerjfr_cr.yaml` be updated to include an empty `storageOptions` object field? The `alm-example` in the CSV could also use an addition of an example `storageOptions`.

Anyway, I did this:

```
--- a/deploy/crds/rhjmc.redhat.com_v1beta1_containerjfr_cr.yaml
+++ b/deploy/crds/rhjmc.redhat.com_v1beta1_containerjfr_cr.yaml
@@ -5,3 +5,11 @@ metadata:
 spec:
   minimal: false
   trustedCertSecrets: []
+  storageOptions:
+    pvcSpec:
+      accessModes: ['ReadWriteMany']
+      resources:
+        limits:
+          storage: '2Gi'
+        requests:
+          storage: '1Gi'
```

and then created the CR and found that the created PVC reflected the limits/requests that I had set, and still successfully bound to one of the existing PVs available and created automatically in CRC. Everything made sense from a readthrough as well.

deploy/crds/rhjmc.redhat.com_containerjfrs_crd.yaml
ebaron: That sounds like a useful feature. I filed #167.

ebaron:
I did a slight revision of the API to add configurable labels and annotations for the PVC as well. I think it's easier to do this now, rather than break API later.

ebaron:
> Also, should `deploy/crds/rhjmc.redhat.com_v1beta1_containerjfr_cr.yaml` be updated to include an empty `storageOptions` object field? The `alm-example` in the CSV could also use an addition of an example `storageOptions`.

I added some blank fields to the CR: `storageOptions.pvc.{labels,annotations,spec}`

andrewazores:
The latest changes look good. I still think updating the CSV's `alm-examples` might be beneficial, although I'm not sure if this might also have some impact on the OpenShift Console wizard when installing the Operator (ex. the toggle button that is normally displayed for the `minimal` spec field). Otherwise, I think this is good to go.

ebaron:
> The latest changes look good. I still think updating the CSV's `alm-examples` might be beneficial, although I'm not sure if this might also have some impact on the OpenShift Console wizard when installing the Operator (ex. the toggle button that is normally displayed for the `minimal` spec field). Otherwise, I think this is good to go.

This will have to come in the next release's CSV won't it? The `alm-examples` from all previous CSVs are still accurate for the versions of CRDs in their associated bundles.

andrewazores:
Yea, I guess that makes sense!

andrewazores (approved):

PR #181, opened 2021-03-22
Add check in deploy_bundle to make sure cert manager is installed

Alexjsenn:
This check existed before in the deploy target but got lost in the operator sdk upgrade. I added it to the deploy_bundle given it's the command I have been using, but it probably makes sense to add it to deploy as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,7 @@ remove_cert_manager:
 
 .PHONY: check_cert_manager
 check_cert_manager:
-       @api_versions=$$(kubectl api-versions) &&\
+	@api_versions=$$(kubectl api-versions) &&\
        if [ $$(echo "$${api_versions}" | grep -c '^cert-manager.io/v1$$') -eq 0 ]; then if [ "$${DISABLE_SERVICE_TLS}" != "true" ]; then\
                        echo 'cert-manager is not installed, install using "make cert_manager" or disable TLS for services by setting DISABLE_SERVICE_TLS to true' >&2\
                        && exit 1;\
--- a/Makefile
+++ b/Makefile
@@ -181,6 +181,11 @@ bundle-build:
 
 .PHONY: deploy_bundle
 deploy_bundle: undeploy_bundle
+ifeq ($(shell kubectl api-versions | grep -c '^cert-manager.io/v1$$'), 0)
+ifneq ($(DISABLE_SERVICE_TLS), true)
+	$(error cert-manager is not installed, install using "make cert-manager" or disable TLS for services by setting DISABLE_SERVICE_TLS to true)
+endif
+endif
 	operator-sdk run bundle $(IMAGE_NAMESPACE)/$(OPERATOR_NAME)-bundle:$(IMAGE_VERSION)
 ifeq ($(DISABLE_SERVICE_TLS), true)
 	@echo "Disabling TLS for in-cluster communication between Services"
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ print_deploy_config: predeploy
 
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 .PHONY: deploy
-deploy: manifests kustomize predeploy
+deploy: check_cert_manager manifests kustomize predeploy
 	$(KUSTOMIZE) build config/default | $(CLUSTER_CLIENT) apply -f -
 ifeq ($(DISABLE_SERVICE_TLS), true)
 	@echo "Disabling TLS for in-cluster communication between Services"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -169,6 +169,11 @@ bundle-build:
 
 .PHONY: deploy_bundle
 deploy_bundle: undeploy_bundle
+ifeq ($(shell kubectl api-versions | grep -c '^cert-manager.io/v1$$'), 0)
+ifneq ($(DISABLE_SERVICE_TLS), true)
+	$(error cert-manager is not installed, install using "make cert-manager" or disable TLS for services by setting DISABLE_SERVICE_TLS to true)
```
ebaron, line 174: This should be "make cert_manager"

ebaron (commented):
We'll also need this check for the `deploy` target. Maybe you could extract this check into a new target and use it as a prerequisite for both `deploy` and `deploy_bundle`?

Makefile
Alexjsenn: Fixed!

Makefile
```diff
@@ -128,7 +128,13 @@ cert_manager: remove_cert_manager
...
 	- $(CLUSTER_CLIENT) delete -f $(CERT_MANAGER_MANIFEST)
 
-
+.PHONY: check_cert_manager
+check_cert_manager:
+ifeq ($(shell kubectl api-versions | grep -c '^cert-manager.io/v1$$'), 0)
+ifneq ($(DISABLE_SERVICE_TLS), true)
+	$(error cert-manager is not installed, install using "make cert_manager" or disable TLS for services by setting DISABLE_SERVICE_TLS to true)
+endif
+endif
```
ebaron, line 137: I tweaked this check to use the shell instead of being evaluated by Make when parsing the Makefile. What do you think?
```suggestion
       @api_versions=$$(kubectl api-versions) &&\
       if [ $$(echo "$${api_versions}" | grep -c '^cert-manager.io/v1$$') -eq 0 ]; then if [ "$${DISABLE_SERVICE_TLS}" != "true" ]; then\
                       echo 'cert-manager is not installed, install using "make cert_manager" or disable TLS for services by setting DISABLE_SERVICE_TLS to true' >&2\
                       && exit 1;\
               fi;\
       fi
```

ebaron (commented):
Thanks for fixing that Alex!

One other thing I've noticed about this check is that Make runs it no matter what target is specified. This is by design, but not ideal for us. If we did it using the shell, we'd get the behaviour we're looking for.

For example, if CRC isn't running. I get the following when building the operator:
```
$ make manager
error: couldn't get available api versions from server: Get "https://api.crc.testing:6443/api?timeout=32s": dial tcp 192.168.130.11:6443: connect: no route to host
/home/ebaron/git/container-jfr-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
go build -o bin/manager main.go
```

It doesn't break anything, but causes a bit of slowdown and confusion since this is a local build.

Implementing the check as a shell command, we get the following behaviour:

CRC not running:
```
$ make manager
/home/ebaron/git/container-jfr-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
go build -o bin/manager main.go

$ make deploy
error: couldn't get available api versions from server: Get "https://api.crc.testing:6443/api?timeout=32s": dial tcp 192.168.130.11:6443: connect: no route to host
make: *** [Makefile:133: check_cert_manager] Error 1
```

CRC running:
```
$ make deploy
cert-manager is not installed, install using "make cert_manager" or disable TLS for services by setting DISABLE_SERVICE_TLS to true
make: *** [Makefile:133: check_cert_manager] Error 1

$ make deploy DISABLE_SERVICE_TLS=true
/home/ebaron/git/container-jfr-operator/bin/controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" rbac:roleName=role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
pushd config/manager && /home/ebaron/git/container-jfr-operator/bin/kustomize edit set image controller=quay.io/ebaron/container-jfr-operator:1.0.0-beta5 && popd
[...]
```

Makefile
```diff
@@ -140,7 +140,14 @@ cert_manager: remove_cert_manager
...
 
-
+.PHONY: check_cert_manager
+check_cert_manager:
+       @api_versions=$$(kubectl api-versions) &&\
+       if [ $$(echo "$${api_versions}" | grep -c '^cert-manager.io/v1$$') -eq 0 ]; then if [ "$${DISABLE_SERVICE_TLS}" != "true" ]; then\
+                       echo 'cert-manager is not installed, install using "make cert_manager" or disable TLS for services by setting DISABLE_SERVICE_TLS to true' >&2\
+                       && exit 1;\
+               fi;\
+       fi
```
ebaron, line 150: These lines seem to be indented with spaces instead of tabs, because of this Make doesn't recognize them as a recipe.

ebaron (changes requested):

Makefile
Alexjsenn: My bad, I was sure it worked when I tested it out but it must have missed it.

ebaron (approved):
Looks good now, thanks!

PR #182, opened 2021-03-25
operator-sdk 1.5

andrewazores:


Makefile
```diff
@@ -81,18 +81,25 @@ install: manifests kustomize
...
-	pushd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG) && popd
-	$(KUSTOMIZE) build config/default | sed 's/container-jfr-operator-system/$(DEPLOY_NAMESPACE)/' | $(CLUSTER_CLIENT) apply -f -
+deploy: manifests kustomize predeploy
+	$(KUSTOMIZE) build config/default | $(CLUSTER_CLIENT) apply -f -
 
 # UnDeploy controller from the configured Kubernetes cluster in ~/.kube/config
 .PHONY: undeploy
-undeploy:
+undeploy: destroy_containerjfr_cr
 	- $(CLUSTER_CLIENT) delete recording --all
```
ebaron, line 101: The order here unfortunately matters. If the Container JFR instance is deleted, the Recording reconciler will fail to clean up the Recording CRs and require manual intervention. So we need to delete Recordings before the ContainerJFR CR.

bundle.Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 
+# Core bundle labels.
 LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=container-jfr-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.4.2
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.4.0+git
```
ebaron, line 10: Should this be `-v1.5.0`?

bundle/metadata/annotations.yaml
```diff
@@ -1,11 +1,14 @@
...
-  operators.operatorframework.io.bundle.manifests.v1: manifests/
+  # Core bundle annotations.
   operators.operatorframework.io.bundle.mediatype.v1: registry+v1
+  operators.operatorframework.io.bundle.manifests.v1: manifests/
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: container-jfr-operator
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.4.2
+  operators.operatorframework.io.bundle.channels.v1: alpha
   operators.operatorframework.io.metrics.mediatype.v1: metrics+v1
+  operators.operatorframework.io.metrics.builder: operator-sdk-v1.4.0+git
```
ebaron, line 9: Same thing here.

ebaron (commented):
Hey Andrew, thanks for doing this. Just a few minor comments below.

Makefile
andrewazores: Ah right. Fixed.

bundle.Dockerfile
andrewazores: This is autogenerated in `make bundle` - I'm not sure why the version isn't the expected one here, but that's what gets generated.

bundle/metadata/annotations.yaml
andrewazores: Also autogenerated

ebaron (approved):
Looks good to me!

Issue #165, opened 2021-02-25
DISABLE_SERVICE_TLS environment variable not set in operator


ebaron:
This environment variable must be set in the operator's pod template in order to work.

ebaron:
I think documenting `kubectl set env deployment/container-jfr-operator-controller-manager DISABLE_SERVICE_TLS=true` is sufficient.

PR #183, opened 2021-03-26
Fix DISABLE_SERVICE_TLS

ebaron:
This PR fixes the `DISABLE_SERVICE_TLS` environment variable functionality to properly set up Container JFR without requiring cert-manager. I modified the ServiceSpecs struct to use URLs instead of strings. This allows us to pass the scheme, host and port to `NewDeploymentForCR`, which can adjust the environment variables as necessary. Right now, Routes always use some form of TLS even without cert-manager, but with incoming Ingress support, this will not be true. This PR makes the code more flexible to support that use case.

When `DISABLE_SERVICE_TLS` is set to true in the Makefile, the `deploy` and `deploy_bundle` targets use `kubectl set env` to add the environment variable to the operator deployment.

I added some tests to exercise the cases where code paths diverge when cert-manager is disabled. I also modified the effected tests to treat the `resource_definitions` package as a black box. Instead of using this package for the expectation for the tests, the expected Kubernetes objects are placed in the `test` package. We might be able to improve this further by moving the objects to golden JSON/YAML files.

There are a number of related minor fixes as well:
- Print a log message and requeue when the Route configuration is not ready, instead of returning an error. This tidies up the log a bit.
- The `redhat.com/containerJfrUrl` contains the full URL, including the scheme. The README has been updated accordingly.
- The test arguments are changed in the Makefile in order to ensure the `ginkgo` CLI outputs the coverage file.
- Removed the leftover package-scoped log in recording_controller.go, and adjusted log calls to use their controller's log.

Fixes: #165

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,9 @@ print_deploy_config: predeploy
 .PHONY: deploy
 deploy: manifests kustomize predeploy
 	$(KUSTOMIZE) build config/default | $(CLUSTER_CLIENT) apply -f -
+ifeq ($(DISABLE_SERVICE_TLS), true)
+	$(CLUSTER_CLIENT) -n $(DEPLOY_NAMESPACE) set env deployment/container-jfr-operator-controller-manager DISABLE_SERVICE_TLS=true
+endif
 
 # UnDeploy controller from the configured Kubernetes cluster in ~/.kube/config
 .PHONY: undeploy
@@ -178,6 +181,9 @@ bundle-build:
 .PHONY: deploy_bundle
 deploy_bundle: undeploy_bundle
 	operator-sdk run bundle $(IMAGE_NAMESPACE)/$(OPERATOR_NAME)-bundle:$(IMAGE_VERSION)
+ifeq ($(DISABLE_SERVICE_TLS), true)
+	$(CLUSTER_CLIENT) set env deployment/container-jfr-operator-controller-manager DISABLE_SERVICE_TLS=true
+endif
 
 .PHONY: undeploy_bundle
 undeploy_bundle:
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ endif
 GINKGO ?= $(shell go env GOPATH)/bin/ginkgo
 GO_TEST ?= go test
 ifneq ("$(wildcard $(GINKGO))","")
-GO_TEST="$(GINKGO)"
+GO_TEST="$(GINKGO)" -cover -outputdir=.
 endif
 
 all: manager
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test-envtest: generate fmt vet manifests
 	mkdir -p $(ENVTEST_ASSETS_DIR)
 	test -f $(ENVTEST_ASSETS_DIR)/setup-envtest.sh || curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.7.2/hack/setup-envtest.sh
-	source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); $(GO_TEST) -v ./... -coverprofile cover.out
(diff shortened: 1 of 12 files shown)


andrewazores:
The changes make sense and look good, and testing the various scenarios of `DISABLE_SERVICE_TLS=true|''` and cert manager installed or not looks good with `make deploy`. The standard TLS-enabled `make deploy_bundle` looks good too. However, when I tried `DISABLE_SERVICE_TLS=true make bundle_deploy create_containerjfr_cr`, I see in the make output that the deployment was (supposedly) updated, however the logs for the manager container show that it's failing to create the ContainerJFR instance with:

```
2021-03-26T22:36:22.673Z	ERROR	controller-runtime.manager.controller.containerjfr	Reconciler error	{"reconciler group": "rhjmc.redhat.com", "reconciler kind": "ContainerJFR", "name": "containerjfr-sample", "namespace": "myproject", "error": "no matches for kind \"Issuer\" in version \"cert-manager.io/v1\""}
```

Inspecting the pod and the deployment show that the `DISABLE_SERVICE_TLS` env var is not present.

ebaron:
> The changes make sense and look good, and testing the various scenarios of `DISABLE_SERVICE_TLS=true|''` and cert manager installed or not looks good with `make deploy`. The standard TLS-enabled `make deploy_bundle` looks good too. However, when I tried `DISABLE_SERVICE_TLS=true make bundle_deploy create_containerjfr_cr`, I see in the make output that the deployment was (supposedly) updated, however the logs for the manager container show that it's failing to create the ContainerJFR instance with:
> 
> ```
> 2021-03-26T22:36:22.673Z	ERROR	controller-runtime.manager.controller.containerjfr	Reconciler error	{"reconciler group": "rhjmc.redhat.com", "reconciler kind": "ContainerJFR", "name": "containerjfr-sample", "namespace": "myproject", "error": "no matches for kind \"Issuer\" in version \"cert-manager.io/v1\""}
> ```
> 
> Inspecting the pod and the deployment show that the `DISABLE_SERVICE_TLS` env var is not present.

It looks like the deployment spec change gets reverted by OLM. I'll have to think of some other way to do this.

ebaron:
I finally found a way to do this by modifying the Subscription created by `operator-sdk run bundle`. There's a `config` section where you can override what's specified in the CSV. The code to find and patch the subscription within the Makefile isn't the most straightforward, but it seems to work.

Makefile
```diff
@@ -178,6 +182,20 @@ bundle-build:
 .PHONY: deploy_bundle
 deploy_bundle: undeploy_bundle
 	operator-sdk run bundle $(IMAGE_NAMESPACE)/$(OPERATOR_NAME)-bundle:$(IMAGE_VERSION)
+ifeq ($(DISABLE_SERVICE_TLS), true)
+	@echo "Disabling TLS for in-cluster communication between Services"
+	@current_ns=`$(CLUSTER_CLIENT) config view --minify -o 'jsonpath={.contexts[0].context.namespace}'` && \
```
andrewazores, line 187: This seems like a good and more generic way to check for the current namespace rather than using `oc project -q`, which I had documented in the `README.md` before. Maybe this can be document there instead instead/as well?

Makefile
```diff
@@ -178,6 +182,20 @@ bundle-build:
...
+	if [ -z "$${current_ns}" ]; then \
+		echo "Failed to determine Namespace in current context" >&2; \
+		exit 1; \
+	fi; \
+	set -f -- `$(CLUSTER_CLIENT) get sub -l "operators.coreos.com/$(OPERATOR_NAME).$${current_ns}" -o name` && \
+	if [ "$${#}" -ne 1 ]; then \
+		echo -e "Expected 1 Subscription, found $${#}:\n$${@}" >&2; \
+		exit 1; \
+	fi; \
+	$(CLUSTER_CLIENT) -n $(DEPLOY_NAMESPACE) patch --type=merge -p '{"spec":{"config":{"env":[{"name":"DISABLE_SERVICE_TLS","value":"true"}]}}}' "$${1}"
```
andrewazores, line 197: The `contexts[0].context.namespace` check earlier will determine the current context namespace, but the `DEPLOY_NAMESPACE` may be set to a different value. ex. I can simply not set the `DEPLOY_NAMESPACE` and allow it to default to `container-jfr-operator-system`, and then `DISABLE_SERVICE_TLS=true make deploy_bundle` will succeed in performing a bundle deployment in my current namespace, but will fail to modify the `Subscription` because this command looks in the wrong namespace.

Makefile
ebaron: Sure, I changed it in the README.

Makefile
ebaron: Thanks for noticing. I had meant to remove the `-n $(DEPLOY_NAMESPACE)`. I think it should work as expected now.

andrewazores (approved):

Issue #180, opened 2021-03-18
Deployment broken in `default` namespace


andrewazores:
See previous discussion: https://github.com/rh-jmc-team/container-jfr-operator/pull/176

andrewazores:
I'm seeing the same problem as here:

https://bugzilla.redhat.com/show_bug.cgi?id=1934177

My CRC environment does use 4.7, so I think it is the exact same failure scenario. I have tried many permutations of `runAsNonRoot` and `runAsUser`/`runAsGroup` and haven't had any success in deploying `kube-rbac-proxy:v0.8.0`.

ebaron:
It sounds like this is exclusive to the distroless image. We could try `registry.access.redhat.com/ubi8/ubi-minimal:latest`?

andrewazores:
This may be better suited for a separate bug, but `DEPLOY_NAMESPACE=foo make undeploy` will also attempt to destroy the `foo` namespace. That's probably *not good* if a user is trying to deploy this in a namespace they're actually using for anything beyond sandbox testing of our operator.

ebaron:
I've noticed this as well. I'm surprised it's the default behaviour. We could remove the namespace object from `config/manager/manager.yaml` and create it manually (if necessary) in our Makefile, I suppose.

andrewazores:
I addressed the Namespace creation/deletion in my operator-sdk 1.5.0 PR (linked above). If a user tries to manually deploy to a namespace that doesn't exist then the deployment will simply fail.

I've tried updating the `kube-rbac-proxy` to v0.8.0 and different permutations of ubi/distroless, runAsNonRoot/runAsUser and other securityContext settings, but still haven't found a working solution that gets the deployment to work in `default` and non-default namespaces. I'll go through the most sensible permutations that seemed like likely candidates and document what they are and how they fail.

andrewazores:
- Updating `kube-rbac-proxy` to v0.8.0 and setting `runAsNonRoot: true` for its container:

```
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -10,7 +10,7 @@ spec:
     spec:
       containers:
       - name: kube-rbac-proxy
-        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
+        image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
         args:
[...]

PR #190, opened 2021-04-09
Temporarily disable metrics endpoint

ebaron:
This PR temporarily disables the metrics endpoint until #180 is resolved. This is done by setting the metrics address to `0`. From manager.Options:
```go
	// MetricsBindAddress is the TCP address that the controller should bind to
	// for serving prometheus metrics.
	// It can be set to "0" to disable the metrics serving.
	MetricsBindAddress string
```

With metrics disabled, the `kube-rbac-proxy` is no longer required, so this PR disables it. This avoids the symptoms of #180.


andrewazores:
There are some other `auth_proxy_*.yaml`s in `config/rbac` which look like they can be commented out in the `kustomization.yaml` too. I still have a `service/container-jfr-operator-controller-manager-metrics-service` for instance, which I don't think is needed if the `kube-rbac-proxy` container isn't running, right?

ebaron:
Good eye. Those are commented out too now.

andrewazores:
I think the `bundle` needs to be regenerated as well to remove the YAMLs which have already been copied. I tried `make bundle` on my own and it didn't remove them, but `rm -rf bundle && make bundle` looks like it regenerates the expected result in the end. I built/bundled/deployed this myself and everything looks good from a manual check over, so I think that's the last thing needed here.

andrewazores (approved):

ebaron:
Thanks for noticing that too. Starting with a fresh bundle directory removed the leader election role/binding files as well. I noticed that this gets folded into the role defined in the CSV though, so those files should be unnecessary.

PR #194, opened 2021-04-20
Rename project to cryostat-operator

andrewazores:

PR #195, opened 2021-04-21
Generate Kustomize patch for image tag vars 

ebaron:
Builds with OSBS attempt to pin image tags to specific digests. In order to find the images used by our operator, they have to be defined in a predetermined location in the CSV [1]. In our case, this means our environment variables need to be renamed and can no longer refer to a config map for their values.

Unfortunately, Kustomize doesn't appear to have a nice way to handle this scenario. I ended up placing a template patch in our `hack` directory and using `envsubst` to fill in the image tags with the values from our Makefile. The resulting patch goes into the `config/default` Kustomize directory, which then gets picked up by the `deploy` and `bundle` targets. I'm not sure whether we should add `envsubst` to the list of dependencies in the README, or if it's considered a common enough shell utility.

[1] https://osbs.readthedocs.io/en/latest/users.html#pullspec-locations

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ undeploy:
 .PHONY: manifests
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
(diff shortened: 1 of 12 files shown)


ebaron:
Keeping this as a draft until #194 is merged and this is rebased.

andrewazores (approved):

Issue #214, opened 2021-07-20
Port CI improvements from Cryostat


ebaron:
The Cryostat repo has several additions to CI that we should also include in the operator.

- [x] #223
- [x] Enforcing Conventional Commits standard (#216)
- [x] Issue dependencies (#218)
- [x] Release Drafter for release notes (#217)
- [x] Mergify config for backport automation (#215)

PR #219, opened 2021-07-22
ci(quay): push CI builds to Quay

ebaron:
This PR adds a Github workflow to push CI builds to Quay when PRs are merged. This is fairly similar to the implementation in the Cryostat repo, but the method for figuring out the image tag is different. This one is using Make's `--eval` argument to invoke a rule to print the `IMAGE_VERSION` Makefile variable.

The build targets for both PRs and pushes have been changed to `oci-build`, which is the same as before but also builds a container image. This slows down the job a bit, but we can maybe look into caching intermediate container layers. Another possibility is we split up the build step into different PR and push steps, where only the later does the container build.

Some example jobs using my fork and Quay repository:
PR: https://github.com/ebaron/cryostat-operator/actions/runs/1056438894
Push: https://github.com/ebaron/cryostat-operator/actions/runs/1056500603

Relates to: #214

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,4 +50,3 @@ jobs:
     - name: Print image URL
       run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
       if: ${{ github.event_name == 'push' && github.repository_owner == 'cryostatio' }}
-
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,8 +29,8 @@ jobs:
     - name: Get image tag
       id: image-tag
       run: |
-        IMG="$(make --eval='print-img: ; @echo $(IMG)' print-img)"
-        echo "::set-output name=img-tag::$IMG"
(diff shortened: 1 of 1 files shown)


andrewazores:
> This slows down the job a bit, but we can maybe look into caching intermediate container layers.

I've definitely noticed that not having the intermediate layers cached locally slows down the OCI build by a very significant factor, so I think this would be something to look into for sure. It would probably apply to Cryostat too, although the caching mechanism might be different since it's using Jib.

> Another possibility is we split up the build step into different PR and push steps, where only the later does the container build.

I'm not sure I follow... in general I think it might make sense to split it up, but would that specifically gain something in build performance?

ebaron:
> > Another possibility is we split up the build step into different PR and push steps, where only the later does the container build.
> 
> I'm not sure I follow... in general I think it might make sense to split it up, but would that specifically gain something in build performance?

For PRs, we could leave the build as `make generate manifests manager test-envtest`, and for pushes we would instead use `make oci-build`. This way PR checks are not any longer than they are currently.

andrewazores:
Oh, I see. I think that makes a lot of sense then.

andrewazores (approved):

Issue #223, opened 2021-07-23
Push builds to Quay


ebaron:
Upon new pushes, CI container builds should be pushed to Quay for `main` and maintenance branches. Versions in development should be tagged with the next version with a `-dev` suffix. Only `main` should be tagged with `latest`.

Example:
`main` branch -> `2.0.0-dev`, `latest`
`v1` branch -> `1.1.0-dev`

Related to: https://github.com/cryostatio/cryostat/issues/595

PR #224, opened 2021-07-23
chore(version): bump version to 1.1.0-dev

ebaron:
Bumps the v1 operator version to the next development version.

Relates to: #223


andrewazores:
Does the `containerImage` on line 53 and `version` on line `261` of `config/manifests/bases/cryostat-operator.clusterserviceversion.yaml` need to be updated as well?

ebaron:
> Does the `containerImage` on line 53 and `version` on line `261` of `config/manifests/bases/cryostat-operator.clusterserviceversion.yaml` need to be updated as well?

The first one for sure. the second one seems to get updated in the generated CSV by `make bundle`, but still good to change. I'll have to fix these in main as well. Thanks for catching it!

andrewazores (approved):

Issue #236, opened 2021-08-26
Project only listed TrustedCertSecret keys


ebaron:
When creating a Secret-backed volume, it's possible to specify which keys of that Secret are projected into the volume [1]. We currently only mount the keys mentioned in the `TrustedCertSecrets` list into the Cryostat container, but the volume available to the Pod contains the whole Secret. We could offer additional protection to these Secrets by only including the necessary keys in the Pod's volume.

[1] https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/volume/#projections (see: secret.items)

Issue #239, opened 2021-08-26
Improve cert-manager install experience


ebaron:
It's not great from a usability perspective to require the user to know ahead of time to install cert-manager before installing the operator. Defaulting to securing services with cert-manager is good from a security perspective, but maybe we can make it more obvious to users that they need to either install cert-manager or disable this functionality. If neither installed or disabled, the operator will fail to deploy Cryostat, but the reason is only found within the operator's logs.

Currently, disabling cert-manager integration is done by setting the `DISABLE_SERVICE_TLS` environment variable in the operator to `true`. This is not very visible and doing this when installed through OLM requires editing the Subscription object, which may not be familiar to many users.

We could instead add a boolean property to the Cryostat CRD. This makes the cert-manager integration more visible and easier to disable. This could work as follows:
- `enableCertManager == true` (default in sample CR), deploy Cryostat with cert-manager integration
- `enableCertManager == false`, deploy Cryostat without cert-manager integration
- `enableCertManager == nil`, defer to `DISABLE_SERVICE_TLS` environment variable for backward compatibility

Something else we could look at is checking for the presence of cert-manager inside `certmanager.go:setupTLS`. We could check for it using the Discovery API, as we do for OpenShift Routes. If it's not available, we could emit a Warning Event that cert-manager is not installed and direct the user to install or disable it.

PR #241, opened 2021-08-30
feat(tls): add cert-manager toggle to Cryostat CRD

ebaron:
This PR adds an API field to the Cryostat CRD to allow the user to more easily adjust whether cert-manager should be enabled. This also has the benefit of making the default requirement on cert-manager more visible to the user.

The IsCertManagerEnabled function now takes the Cryostat CR and checks for the `spec.certManagerEnabled` property. If absent, the check falls back to the `DISABLE_SERVICE_TLS` environment variable for backward compatibility. Because this allows cert-manager enablement to be adjusted at runtime, some adjustments had to be made to allow the deployment to swap between enabled/disabled cert-manager configurations.

The deployment now uses CreateOrUpdate to update the whole pod template to add/remove various TLS-related volumes, mounts, environment variables. Likewise for Routes to change their TLS termination and destination CA cert. When the CA cert mounted under `/truststore` was removed when transitioning from enabled to disabled cert-manager, an empty file for the mount point was left behind. Cryostat attempted to add this empty file to its trust store and crashed. To avoid this, I combined the CA cert and any user-supplied certs into a single projected volume [1]. This volume then gets mounted under `/truststore/operator/`. This works because Cryostat looks for certs within subdirectories as well.

Screenshot from Cryostat CR creation page:
Screenshot 2021-08-31 at 12-43-51 Create Cryostat · Red Hat OpenShift Container Platform

Relates to: #239 
Fixes: #236 

[1] https://kubernetes.io/docs/concepts/storage/volumes/#projected

### DIFF
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -449,7 +449,7 @@ func NewCoreContainer(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTa
 		}
 
 		// Mount the CA cert and user certificates in the expected /truststore location
-		caCertMount := corev1.VolumeMount{ // FIXME this is persisting after redeployment with 0 byte file. Try projected volume instead of subpath?
+		caCertMount := corev1.VolumeMount{
 			Name:      "cert-secrets",
 			MountPath: "/truststore/operator",
 			ReadOnly:  true,
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -182,17 +182,76 @@ func NewPodForCR(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTags *I
 			},
 		},
 	}
+	readOnlyMode := int32(0440)
 	if tls != nil {
 		// Create certificate secret volumes in deployment
-		secretVolume := corev1.Volume{
-			Name: "tls-secret",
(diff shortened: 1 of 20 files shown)


ebaron:
To test this, I would recommend toggling `enableCertManager` off and then back on again, and ensuring nothing breaks.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -391,39 +436,20 @@ func NewCoreContainer(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTa
...
 			ReadOnly:  true,
 		}
 
-		// Mount the CA cert in the expected /truststore location
-		caCertMount := corev1.VolumeMount{
-			Name:      "tls-secret",
-			MountPath: fmt.Sprintf("/truststore/%s-ca.crt", cr.Name),
-			SubPath:   CAKey,
+		// Mount the CA cert and user certificates in the expected /truststore location
+		caCertMount := corev1.VolumeMount{ // FIXME this is persisting after redeployment with 0 byte file. Try projected volume instead of subpath?
```
andrewazores, line 445: This comment is outdated now, right?

andrewazores (commented):
Code looks good, and manually testing this seems to work as advertised (very smooth switchover, it's neat!)

internal/controllers/common/resource_definitions/resource_definitions.go
ebaron: It is, thanks for catching it!

PR #252, opened 2021-09-02
fix(pvc): derive fsGroup from SCC annotation

### LINKED ISSUE #247: fsGroup is not always permitted
It seems that a fix beyond #244 is needed to solve #243 in all cluster configurations. 

In OpenShift, SecurityContextConstraints can forbid the `fsGroup` value we added:
```
Error creating: pods "cryostat-sample-775b9b7d9d-" is forbidden: unable to validate against any security context constraint: [provider restricted: .spec.securityContext.fsGroup: Invalid value: []int64{18500}: 18500 is not an allowed group]
```
https://docs.openshift.com/container-platform/4.8/authentication/managing-security-context-constraints.html#security-context-constraints-pre-allocated-values_configuring-internal-oauth

There are also PodSecurityPolicies, where a range of GIDs can be specified for allowable `fsGroup` values:
https://kubernetes.io/docs/concepts/policy/pod-security-policy/

ebaron:
This PR adds some logic for OpenShift to ensure Cryostat runs under the `restricted` SecurityContextConstraint. The range of permitted `fsGroup` values is taken from the `openshift.io/sa.scc.supplemental-groups` annotation on the pod's namespace:
https://docs.openshift.com/container-platform/4.8/authentication/managing-security-context-constraints.html#security-context-constraints-pre-allocated-values_configuring-internal-oauth

I'm still unsure why Cryostat deployed to the `default` namespace isn't subject to SCCs, but this fix appears to make Cryostat work for both `default` and other namespaces.

Fixes #247

### DIFF
--- a/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
@@ -253,6 +253,14 @@ spec:
       clusterPermissions:
       - rules:
         - apiGroups:
+          - ""
+          resources:
+          - namespaces
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
           - authentication.k8s.io
           resources:
           - tokenreviews
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,14 @@ metadata:
   name: role
 rules:
 - apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - authentication.k8s.io
   resources:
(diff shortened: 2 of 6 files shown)


andrewazores:
Seems to make sense, I'll test it out soon. I don't suppose there is any similar annotation for UIDs for #180?

andrewazores (approved):
Seems to work as expected in `default` and a different ("`myproject`") namespace in CRC.

ebaron:
> Seems to make sense, I'll test it out soon. I don't suppose there is any similar annotation for UIDs for #180?

There is an annotation for UID ranges, but the operator wouldn't be able to handle it in the same way because the problem with #180 is the operator's own deployment.

Issue #254, opened 2021-09-02
ImagePullPolicy should be Always for development images


ebaron:
For unreleased images, where the tag is continually updated to the newest CI build, the `ImagePullPolicy` should be set to `Always` instead of the default `IfNotPresent`. The latter causes problems where we think we're testing the latest builds, but the cluster is actually using older cached images from its registry.

Since there are a variety of suffixes we use, `-SNAPSHOT`, `.BETA*`, `-dev`, perhaps it's best to set the policy to `Always` if the tag doesn't match something like `[0-9]*.[0-9]*.[0-9]*`. For operand images, this check can be done in the operator code. For the operator image itself, we might need a Kustomize patch or similar.

PR #286, opened 2021-10-25
feat(deploy): use Always pull policy for development images 

ebaron:
Uses a regular expression to check image tags for the following suffixes:
* latest
* SNAPSHOT
* dev
* BETA[0-9]+

If any of the above are found the pull policy for that image will be set to `Always`.

For the operator controller deployment, this is done using a Kustomize patch that is updated by the Makefile depending on the regex test. For the operand images, this same check is done in resource_definitions.go when creating the Deployment spec.

Fixes: #254

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,10 @@ GOBIN=$(shell go env GOBIN)
 endif
 
 # Check whether this is a development or release version
-ifeq (,$(shell echo $(IMAGE_VERSION) | grep -iE '(:latest|SNAPSHOT|dev|BETA[[:digit:]]+)$$'))
-PULL_POLICY ?= IfNotPresent
-else
+ifneq (,$(shell echo $(IMAGE_VERSION) | grep -iE '(:latest|SNAPSHOT|dev|BETA[[:digit:]]+)$$'))
 PULL_POLICY ?= Always
+else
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -54,6 +54,14 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Check whether this is a development or release version
+ifeq (,$(shell echo $(IMAGE_VERSION) | grep -iE '(:latest|SNAPSHOT|dev|BETA[[:digit:]]+)$$'))
```
andrewazores, line 58: Patch overall makes sense to me. I found this conditional counter-intuitive, since it "feels" like it should be checking if the version ends with `:latest`/`xSNAPSHOT`/`xdev` etc., but in fact it's checking the opposite because it's comparing the result to the empty string. IMO it'd be slightly easier to read if this was `ifneq` and the `PULL_POLICY`s swapped, but YMMV. I'm fine with it as is.

andrewazores (approved):

Makefile
ebaron: Agreed. I've switched this to `ifneq` swapped the branches.

andrewazores (approved):

PR #292, opened 2021-11-05
feat(auth): Add OAuth client to service account

### LINKED ISSUE #289: Create OAuthClient resource
The `OAuthClient` resource in OpenShift should be created by the Operator, which can generate and store the client secret and then use that to create the resource. If it's needed it can also supply it to Cryostat, either by mounting the secret as a volume to the Cryostat container or by populating an environment variable with it, etc.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/717#issuecomment-954791706_

jan-law:
Fixes #289 

Lets the cryostat-sample service account access the OpenShift OAuth API by acting as an OAuth client. The redirect uri is created dynamically.

To request a [scoped access token](https://docs.openshift.com/container-platform/4.9/authentication/tokens-scoping.html#scoping-tokens-role-scope_configuring-internal-oauth) with the necessary permissions for the web-UI to create and manage recordings, I added permissions to the existing ClusterRole. The names of the ClusterRole and ServiceAccount are both referenced when making a backend request to the OAuth server in this PR: https://github.com/cryostatio/cryostat/pull/748

```
oc describe sa cryostat-sample
Name:                cryostat-sample
Namespace:           cryostat-operator-system
Labels:              app=cryostat
Annotations:        serviceaccounts.openshift.io/oauth-redirectreference.route:
                       {"metadata":{"creationTimestamp":null},"reference":{"group":"","kind":"Route","name":"cryostat-sample"}}
                     serviceaccounts.openshift.io/oauth-redirecturi.route: https://
Image pull secrets:  cryostat-sample-dockercfg-8x9n4
Mountable secrets:   cryostat-sample-dockercfg-8x9n4
                     cryostat-sample-token-vh7gw
Tokens:              cryostat-sample-token-glb24
                     cryostat-sample-token-vh7gw
Events:              <none>
```

More info:
https://docs.openshift.com/container-platform/4.9/authentication/using-service-accounts-in-applications.html

### DIFF
--- a/config/rbac/cryostat_role.yaml
+++ b/config/rbac/cryostat_role.yaml
@@ -13,45 +13,42 @@ rules:
   verbs:
   - create
 - apiGroups:
-  - authorization.k8s.io
-  resources:
-  - selfsubjectaccessreviews
-  verbs:
-  - create
-- apiGroups:
   - ""
   resources:
   - pods
   verbs:
-  - get
   - create
+  - get
 - apiGroups:
   - operator.cryostat.io
   resources:
   - cryostats
   verbs:
-  - get 
   - create
+  - delete
+  - get
 - apiGroups:
   - operator.cryostat.io
   resources:
   - flightrecorders
   verbs:
-  - get  
+  - create
+  - delete
+  - get
   - patch
 - apiGroups:
   - operator.cryostat.io
   resources:
   - recordings
   verbs:
-  - create 
-  - delete 
-  - get 
-  - patch 
+  - create
+  - delete
+  - get
+  - patch
 - apiGroups:
   - apps
   resources:
   - deployments
   verbs:
-  - get
   - create
+  - get
--- a/config/rbac/cryostat_role.yaml
+++ b/config/rbac/cryostat_role.yaml
@@ -20,6 +20,13 @@ rules:
   - create
   - get
 - apiGroups:
+  - ""
+  resources:
+  - replicationcontrollers
+  - endpoints
+  verbs:
+  - get
+- apiGroups:
   - operator.cryostat.io
   resources:
   - cryostats
@@ -31,14 +38,6 @@ rules:
   - operator.cryostat.io
   resources:
   - flightrecorders
-  verbs:
-  - create
-  - delete
-  - get
-  - patch
-- apiGroups:
-  - operator.cryostat.io
-  resources:
   - recordings
   verbs:
   - create
@@ -51,4 +50,12 @@ rules:
   - deployments
   verbs:
   - create
+  - get
+- apiGroups:
+  - apps
+  resources:
+  - daemonsets
+  - replicasets
+  - statefulsets
+  verbs:
   - get
--- a/config/rbac/cryostat_role.yaml
+++ b/config/rbac/cryostat_role.yaml
@@ -1,4 +1,5 @@
-# Permissions for Cryostat to validate tokens and check permissions
+# Permissions for Cryostat to validate tokens, check permissions, and perform
+# actions with the web-ui
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -16,4 +17,41 @@ rules:
   resources:
   - selfsubjectaccessreviews
   verbs:
+  - create
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - create
+- apiGroups:
+  - operator.cryostat.io
+  resources:
+  - cryostats
+  verbs:
+  - get 
+  - create
+- apiGroups:
+  - operator.cryostat.io
+  resources:
+  - flightrecorders
+  verbs:
+  - get  
+  - patch
+- apiGroups:
+  - operator.cryostat.io
+  resources:
+  - recordings
+  verbs:
+  - create 
+  - delete 
+  - get 
+  - patch 
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - get
   - create
--- a/config/rbac/cryostat_role.yaml
+++ b/config/rbac/cryostat_role.yaml
@@ -1,5 +1,4 @@
-# Permissions for Cryostat to validate tokens, check permissions, and perform
-# actions with the web-ui
+# Permissions for Cryostat to validate tokens and check permissions
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -13,49 +12,8 @@ rules:
   verbs:
   - create
 - apiGroups:
-  - ""
+  - authorization.k8s.io
   resources:
-  - pods
+  - selfsubjectaccessreviews
   verbs:
   - create
-  - get
-- apiGroups:
-  - ""
-  resources:
-  - replicationcontrollers
-  - endpoints
-  verbs:
-  - get
-- apiGroups:
-  - operator.cryostat.io
-  resources:
-  - cryostats
-  verbs:
-  - create
-  - delete
-  - get
-- apiGroups:
-  - operator.cryostat.io
-  resources:
-  - flightrecorders
-  - recordings
-  verbs:
-  - create
-  - delete
-  - get
-  - patch
-- apiGroups:
-  - apps
-  resources:
-  - deployments
-  verbs:
-  - create
-  - get
-- apiGroups:
-  - apps
-  resources:
-  - daemonsets
-  - replicasets
-  - statefulsets
-  verbs:
-  - get
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -376,6 +376,14 @@ func NewCoreContainer(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTa
 				Name:  "CRYOSTAT_AUTH_MANAGER",
 				Value: "io.cryostat.net.OpenShiftAuthManager",
 			},
+			{
+				Name:  "CRYOSTAT_OAUTH_CLIENT_ID",
+				Value: cr.Name,
+			},
+			{
+				Name:  "CRYOSTAT_OAUTH_ROLE",
+				Value: "cryostat-operator-cryostat",
+			},
 		}
 		envs = append(envs, openshiftEnvs...)
 	}
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -837,33 +837,43 @@ func NewKeystoreSecretForCR(cr *operatorv1beta1.Cryostat) *corev1.Secret {
 	}
 }
 
-func NewServiceAccountForCR(cr *operatorv1beta1.Cryostat) (*corev1.ServiceAccount, error) {
-	OAuthRedirectReference := &oauthv1.OAuthRedirectReference{
-		Reference: oauthv1.RedirectReference{
-			Kind: "Route",
-			Name: cr.Name,
-		},
+func NewServiceAccountForCR(cr *operatorv1beta1.Cryostat, isOpenShift bool) (*corev1.ServiceAccount, error) {
+	if isOpenShift {
+		OAuthRedirectReference := &oauthv1.OAuthRedirectReference{
+			Reference: oauthv1.RedirectReference{
+				Kind: "Route",
+				Name: cr.Name,
+			},
+		}
+
+		ref, err := json.Marshal(OAuthRedirectReference)
+		if err != nil {
+			return nil, err
+		}
+
+		sa := &corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cr.Name,
+				Namespace: cr.Namespace,
+				Labels: map[string]string{
+					"app": "cryostat",
+				},
+				Annotations: map[string]string{
+					"serviceaccounts.openshift.io/oauth-redirecturi.route":       "https://",
+					"serviceaccounts.openshift.io/oauth-redirectreference.route": string(ref),
+				},
+			},
+		}
+		return sa, nil
 	}
 
-	ref, err := json.Marshal(OAuthRedirectReference)
-	if err != nil {
-		return nil, err
-	}
-
-	sa := &corev1.ServiceAccount{
+	return &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      cr.Name,
 			Namespace: cr.Namespace,
-			Labels: map[string]string{
-				"app": "cryostat",
-			},
-			Annotations: map[string]string{
-				"serviceaccounts.openshift.io/oauth-redirecturi.route":       "https://",
-				"serviceaccounts.openshift.io/oauth-redirectreference.route": string(ref),
-			},
 		},
-	}
-	return sa, nil
+	}, nil
+
 }
 
 func NewRoleForCR(cr *operatorv1beta1.Cryostat) *rbacv1.Role {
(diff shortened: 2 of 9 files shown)


internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -768,13 +770,33 @@ func NewKeystoreSecretForCR(cr *operatorv1beta1.Cryostat) *corev1.Secret {
...
+	sa := &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      cr.Name,
 			Namespace: cr.Namespace,
+			Labels: map[string]string{
+				"app": "cryostat",
+			},
+			Annotations: map[string]string{
+				"serviceaccounts.openshift.io/oauth-redirecturi.route":       "https://",
+				"serviceaccounts.openshift.io/oauth-redirectreference.route": string(ref),
```
jan-law, line 795: Question: the `serviceaccounts.openshift.io/oauth-redirecturi.route: ` from `oc describe` doesn't seem to have a URL value. How can I check that the URI has actually been generated?

ebaron: If I understand the documentation, this is the correct behaviour. In this case, the scheme provided in the `oauth-redirecturi.route` will overwrite the scheme (is there one?) determined by the route. So while the scheme comes from `oauth-redirecturi.route`, the hostname, port, and path will come from the route supplied in `oauth-redirectreference.route`.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -876,6 +876,26 @@ func NewClusterRoleBindingForCR(cr *operatorv1beta1.Cryostat) *rbacv1.ClusterRol
 	}
 }
 
+func NewOAuthClusterRoleBindingForCR(cr *operatorv1beta1.Cryostat) *rbacv1.ClusterRoleBinding {
```
jan-law, line 879: How could I get more information to debug this error?
Logs from `oc logs -f cryostat-operator-controller-manager-blah`:
```
2021-11-15T19:24:40.410Z	INFO	controllers.Cryostat	Not found	{"Request.Namespace": "", "Name": "cryostat-oauth", "Kind": "*v1.ClusterRoleBinding"}
2021-11-15T19:24:40.415Z	ERROR	controllers.Cryostat	Could not be created	{"Request.Namespace": "", "Name": "cryostat-oauth", "Kind": "*v1.ClusterRoleBinding", "error": "clusterroles.rbac.authorization.k8s.io \"cryostat-oauth\" not found"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.2.0/zapr.go:132
github.com/cryostatio/cryostat-operator/internal/controllers.(*CryostatReconciler).createObjectIfNotExists
	/workspace/internal/controllers/cryostat_controller.go:513
github.com/cryostatio/cryostat-operator/internal/controllers.(*CryostatReconciler).createRBAC
	/workspace/internal/controllers/cryostat_controller.go:570
github.com/cryostatio/cryostat-operator/internal/controllers.(*CryostatReconciler).Reconcile
	/workspace/internal/controllers/cryostat_controller.go:227
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.2/pkg/internal/controller/controller.go:263
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.2/pkg/internal/controller/controller.go:235
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1.1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.2/pkg/internal/controller/controller.go:198
k8s.io/apimachinery/pkg/util/wait.JitterUntilWithContext.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/wait/wait.go:185
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/wait/wait.go:156
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/wait/wait.go:133
k8s.io/apimachinery/pkg/util/wait.JitterUntilWithContext
	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/wait/wait.go:185
k8s.io/apimachinery/pkg/util/wait.UntilWithContext
	/go/pkg/mod/k8s.io/apimachinery@v0.19.2/pkg/util/wait/wait.go:99
```

ebaron: In this case you're missing the ClusterRole, and Kubernetes is refusing to create the ClusterRoleBinding if the referenced ClusterRole is missing. The existing ClusterRole used by the Cryostat deployment is specified here: https://github.com/cryostatio/cryostat-operator/blob/d7678d83aaf4781d230b693b61fce255b031c808/config/rbac/cryostat_role.yaml.

Since the same ClusterRole can be shared by any number of Cryostat deployments, the operator simply creates it at startup. The same isn't true for the ClusterRoleBinding, which binds that ClusterRole to the unique ServiceAccount belonging to a single Cryostat deployment. That's why the ClusterRole is defined in YAML, and the ClusterRoleBInding is created programmatically.

Feel free to use the existing ClusterRole (and its Binding) for these new permissions you need.

jan-law: If I try to add permissions to the ClusterRole in `cryostat_role.yaml`, it looks like the `cryostat-operator-service-account` that creates this ClusterRole needs the same permissions before it can grant them to the ClusterRole. Where are the service account permissions set?
```
ERROR	controller-runtime.manager.controller.cryostat	
Reconciler error	{"reconciler group": "operator.cryostat.io", "reconciler kind": "Cryostat",
 "name": "cryostat-sample", "namespace": "default", "error":
 "clusterrolebindings.rbac.authorization.k8s.io \"cryostat-d6f4ddc1e7a1dbb5e1a1b9cac4fa216d0d90e8f734fe70ebb132a5c400f1c345\" is forbidden: 
user \"system:serviceaccount:default:cryostat-operator-service-account\" (groups=
[\"system:serviceaccounts\" \"system:serviceaccounts:default\" \"system:authenticated\"]) is
 attempting to grant RBAC permissions not currently held:\n{APIGroups: 
<list of new permissions>}
```

ebaron: The Role and ClusterRole for the operator deployment are generated using marker comments, which get processed at build time by the `make manifests` target [1]. Here's where you can add the new permission(s):
https://github.com/cryostatio/cryostat-operator/blob/d7678d83aaf4781d230b693b61fce255b031c808/internal/controllers/cryostat_controller.go#L104-L120

To add permissions to the ClusterRole, don't specify the namespace attribute, otherwise they'll get added to the namespaced Role instead.

[1] https://sdk.operatorframework.io/docs/building-operators/golang/tutorial/#specify-permissions-and-generate-rbac-manifests

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -374,6 +376,14 @@ func NewCoreContainer(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTa
 				Name:  "CRYOSTAT_AUTH_MANAGER",
 				Value: "io.cryostat.net.OpenShiftAuthManager",
 			},
+			{
+				Name:  "CRYOSTAT_OAUTH_CLIENT_ID",
+				Value: cr.Name,
+			},
+			{
+				Name:  "CRYOSTAT_OAUTH_ROLE",
+				Value: "cryostat-operator-cryostat",
```
jan-law, line 385: @ebaron I tried adding environment variables inside `openshiftEnvs := []corev1.EnvVar{...}` for cryostat-sample to access the names of the `serviceaccount` and `clusterRole`, but I don't see these env vars listed on OCP > Pods > cryostat-sample > Environment > Container: cryostat-sample. `env.getEnv("CRYOSTAT_OAUTH_ROLE")` also returns `null`. Any ideas?

ebaron: It looks okay to me, I'll try building and deploying it myself and let you know how it goes.

ebaron:
It worked for me doing the following:
```
$ make generate manifests manager 
$ BUILDAH_FORMAT=docker podman build -t quay.io/ebaron/cryostat-operator:latest .
$ podman push quay.io/ebaron/cryostat-operator:latest
$ make deploy IMG=quay.io/ebaron/cryostat-operator:latest
$ make create_cryostat_cr 
$ oc describe deploy cryostat-sample
[...]
    Environment:
[...]
      CRYOSTAT_OAUTH_CLIENT_ID:      cryostat-sample
      CRYOSTAT_OAUTH_ROLE:           cryostat-operator-cryostat
```

Normally, I'd use `make oci-build` instead of the first two steps, but this bypasses the test suite.

internal/controllers/common/resource_definitions/resource_definitions.go
jan-law: Thanks, your build steps worked!

jan-law:
This PR is ready for review. I'd also appreciate it if you could confirm the ClusterRole permissions give the exact permissions that the web-UI needs and nothing more.

config/rbac/cryostat_role.yaml
```diff
@@ -12,8 +13,49 @@ rules:
...
+  - create
+  - get
+- apiGroups:
+  - apps
+  resources:
+  - daemonsets
+  - replicasets
+  - statefulsets
+  verbs:
+  - get
```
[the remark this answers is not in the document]
jan-law: Setting this PR to draft mode while I add the new ClusterRole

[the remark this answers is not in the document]
jan-law: I've removed `create TokenReviews` and `create SelfSubjectAccessReviews` from the new ClusterRole. The backend can request a token with enough permissions using `user:info user:check-access role:cryostat-operator-oauth-client:namespace`.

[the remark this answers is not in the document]
ebaron: Hey Janelle, I haven't tried it out yet, but do you know if the `user:info` scope allows the backend to create a TokenReview, or does that code need to be changed to use the OpenShift users API? I assumed it was the latter.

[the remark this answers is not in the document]
jan-law: I'm assuming the `user:info` scope lets the backend create a TokenReview, because when I tested my latest image on OpenShift with the following ClusterRole permissions and `scope=user%3Ainfo+user%3Acheck-access+role%3Acryostat-operator-oauth-client` as the role scope, I didn't see any errors on the frontend or in the `cryostat-sample` logs.

```
Name:         cryostat-operator-oauth-client
Labels:       <none>
Annotations:  <none>
PolicyRule:
  Resources                             Non-Resource URLs  Resource Names  Verbs
  ---------                             -----------------  --------------  -----
  flightrecorders.operator.cryostat.io  []                 []              [create delete get patch]
  recordings.operator.cryostat.io       []                 []              [create delete get patch]
  cryostats.operator.cryostat.io        []                 []              [create delete get]
  pods                                  []                 []              [create get]
  deployments.apps                      []                 []              [create get]
  endpoints                             []                 []              [get]
  replicationcontrollers                []                 []              [get]
  daemonsets.apps                       []                 []              [get]
  replicasets.apps                      []                 []              [get]
  statefulsets.apps                     []                 []              [get]
```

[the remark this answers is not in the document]
ebaron: I think what we're seeing is the TokenReview is being done using Cryostat's ServiceAccount: https://github.com/cryostatio/cryostat/blob/7c010b29b83677ba0a815315ddd50ada6afb5659/src/main/java/io/cryostat/net/OpenShiftAuthManager.java#L270-L288

It still has the `create tokenreviews` permission from the existing ClusterRole. So the `user:info` scope is not being used right now. Later on, if we change the TokenReview to an OpenShift user API query using the scoped token, the `user:info` scope will be necessary. I think that's probably best saved for another PR though.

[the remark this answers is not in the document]
jan-law: Thanks for the clarification. I'll remove `user:info` from the backend request then.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -827,13 +837,33 @@ func NewKeystoreSecretForCR(cr *operatorv1beta1.Cryostat) *corev1.Secret {
...
 			Namespace: cr.Namespace,
+			Labels: map[string]string{
+				"app": "cryostat",
+			},
+			Annotations: map[string]string{
+				"serviceaccounts.openshift.io/oauth-redirecturi.route":       "https://",
+				"serviceaccounts.openshift.io/oauth-redirectreference.route": string(ref),
+			},
 		},
 	}
```
ebaron, line 865: Although it's probably harmless, I think we should only set these annotations when running on OpenShift. A Kubernetes user might be confused why there are OpenShift annotations pointing to a non-existent Route. You can pass `r.IsOpenShift` when calling this method like we do for `NewDeploymentForCR`.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -827,13 +837,43 @@ func NewKeystoreSecretForCR(cr *operatorv1beta1.Cryostat) *corev1.Secret {
...
+					"app": "cryostat",
+				},
+				Annotations: map[string]string{
+					"serviceaccounts.openshift.io/oauth-redirecturi.route":       "https://",
+					"serviceaccounts.openshift.io/oauth-redirectreference.route": string(ref),
+				},
+			},
+		}
+		return sa, nil
+	}
```
ebaron, line 868: Just one more suggestion to reduce duplication, you could:
1. Create an empty annotations map
2. Put the OAuth annotations in the map inside the OpenShift branch
3. Return the ServiceAccount with annotations set to the map. In Kubernetes this will be empty, in OpenShift it'll contain the OAuth annotations

ebaron (commented):
I noticed one more thing, this new ClusterRole will need to be added to our operator bundle. Running `make bundle` should copy it to the necessary place.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -827,13 +837,38 @@ func NewKeystoreSecretForCR(cr *operatorv1beta1.Cryostat) *corev1.Secret {
...
+			},
+		}
+
+		ref, err := json.Marshal(OAuthRedirectReference)
+		if err != nil {
+			return nil, err
+		}
+
+		annotations = map[string]string{
+			"serviceaccounts.openshift.io/oauth-redirecturi.route":       "https://",
```
ebaron, line 857: I think perhaps we should remove this annotation. Right now the operator always creates Routes with some form of TLS enabled, but I think in the near future we may want to allow the user to customize the Routes we create like we do with Ingresses and now Services. In the interest of future-proofing this, I suggest we don't assume HTTPS here.

I tracked down the code that parses the route provided in the `oauth-redirectreference.route` annotation:
https://github.com/openshift/library-go/blob/7a65fdb398e28782ee1650959a5e0419121e97ae/pkg/oauth/oauthserviceaccountclient/oauthclientregistry.go#L418-L441

It will use HTTPS for the scheme, unless the Route has no TLS configuration, which sounds like what we'd want. We should be fine to not override this behaviour by forcing HTTPS.

ebaron (approved):
Looks good, nice work!

PR #305, opened 2021-11-22
docs(ingress): Add link to ingress docs

### LINKED ISSUE #303: Document Ingress config in CSV
It would be helpful to include or link to this piece of documentation, since it details an additional step users must take to run the operator on Kubernetes:
https://github.com/cryostatio/cryostat-operator/blob/v2.0.0/docs/config.md#network-options

jan-law:
Fixes #303

Also fixed a typo in the Cryostat description from `capbility` to `capability`.

Question: what is the difference between the `docs` and the `documentation` label?


andrewazores:
`docs` should be used since it matches the commit message and would correspond to the desired section in the autogenerated release notes. `documentation` is just one of the GitHub default ones and we should probably remove it - maybe we would need to go back and retroactively migrate old issues/PRs using it to use `docs` though.

ebaron (approved):
Thanks Janelle, nicely done!

ebaron:
> `docs` should be used since it matches the commit message and would correspond to the desired section in the autogenerated release notes. `documentation` is just one of the GitHub default ones and we should probably remove it - maybe we would need to go back and retroactively migrate old issues/PRs using it to use `docs` though.

No sense having both, I added the `docs` label to any issues/PRs with the `documentation` label and deleted the latter.

PR #312, opened 2021-11-30
fix(ingress): Recommend Ingress timeout as 3600s

### LINKED ISSUE #311: Extend Ingress timeout on Kubernetes
If the 3600 timeout is useful for packaging/productization/downstream type things like OperatorHub then I think it's probably a good idea to include that configuration for now. Once we have an upstream release that includes application-level keepalive then the downstream configuration for increasing the timeout can be removed.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat/issues/757#issuecomment-983070243_

@jan-law Try adding these annotations to each Ingress:
```
nginx.ingress.kubernetes.io/proxy-read-timeout: 3600
nginx.ingress.kubernetes.io/proxy-send-timeout: 3600
```
This seems to be the recommended solution from the NGINX Ingress documentation: https://kubernetes.github.io/ingress-nginx/user-guide/miscellaneous/#websockets

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/issues/279#issuecomment-983039470_

jan-law:
Fixes #311 

Since users are responsible for configuring the Ingress configurations, I added the recommended annotations in the readme.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -93,6 +93,8 @@ spec:
     coreConfig:
       annotations:
         nginx.ingress.kubernetes.io/backend-protocol : HTTPS
+        nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
+        nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
       ingressSpec:
         tls:
         - {}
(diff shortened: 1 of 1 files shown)


ebaron (approved):
Looks good, thanks Janelle!

PR #313, opened 2021-12-06
feat(api): add service customization to CRD

### LINKED ISSUE #304: Support Service annotations
Currently there is no way how to change the Service type to anything else but LoadBalancer. On top of that annotations can not be set and so when the Cryostat is deployed into a public cloud where the decision whether to create external or internal Load Balancer is based on annotations, the deployment always creates public Load Balancer.

There should be a way to set the annotations so that one can decide whether to deploy an external or internal Load Balancer among other configurations of those Load Balancers which are controlled by the annotations.

ebaron:
This PR adds a `spec.serviceOptions` property to the Cryostat CRD. This allows the user to override the service type, port numbers, labels and annotations of the Cryostat and Grafana services that the operator creates. I wasn't able to reuse the existing `networkOptions` property since the user may want different labels/annotations applied to the Service and Ingress.

You can test this by creating a Cryostat CR like this instead of using `make create_cryostat_cr`:
```
$ cat << 'EOF' | kubectl create -f -
apiVersion: operator.cryostat.io/v1beta1
kind: Cryostat
metadata:
  name: cryostat-sample
spec:
  minimal: false
  serviceOptions:
    coreConfig:
      labels:
        my-custom-label: some-value
      annotations: 
        my-custom-annotation: some-value
      serviceType: NodePort
      httpPort: 8080
      jmxPort: 9095
    grafanaConfig:
      labels:
        my-custom-label: some-value
      annotations: 
        my-custom-annotation: some-value
      serviceType: NodePort
      httpPort: 8080
EOF
```

Observe that the `cryostat-sample` and `cryostat-sample-grafana` services are created with the values specified. The deployed Cryostat application should work as normal.

Fixes #304


api/v1beta1/cryostat_types.go
```diff
@@ -86,6 +92,57 @@ type StorageConfiguration struct {
...
+
+// CoreServiceConfig provides customization for the service handling
+// traffic for the Cryostat application
+type CoreServiceConfig struct {
+	// HTTP port number for the Cryostat application service.
+	// Defaults to 8181.
+	// +optional
+	HTTPPort *int32 `json:"httpPort,omitempty"`
+	// Remote JMX port number for the Cryostat application service.
+	// Defaults to 9191.
```
andrewazores, line 119: 9091, isn't it?

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -653,30 +653,61 @@ func NewJfrDatasourceContainer(cr *operatorv1beta1.Cryostat, imageTag string) co
...
+	svcType := corev1.ServiceTypeClusterIP
+	httpPort := int32(8181)
+	jmxPort := int32(9091)
+	labels := map[string]string{}
+	annotations := map[string]string{}
+
+	// Override with any user-specified properties
+	if cr.Spec.ServiceOptions != nil && cr.Spec.ServiceOptions.CoreConfig != nil {
+		config := cr.Spec.ServiceOptions.CoreConfig
+		if config != nil {
```
andrewazores, line 667: Can it be `nil` here? Seems like the RHS of the first `if` two lines up should prevent that

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -653,30 +653,61 @@ func NewJfrDatasourceContainer(cr *operatorv1beta1.Cryostat, imageTag string) co
...
+				labels = config.Labels
+			}
+			if config.Annotations != nil {
+				annotations = config.Annotations
+			}
+		}
+	}
+
+	// Add required labels, overriding any user-specified labels with the same keys
+	labels["app"] = cr.Name
```
andrewazores, line 687: Would it make sense to emit a warning or something if the user has specified labels that we are going to override?

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -711,24 +742,51 @@ func NewCommandChannelService(cr *operatorv1beta1.Cryostat) *corev1.Service {
...
```
andrewazores, line 771: Same here re. emitting a warning

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -711,24 +742,51 @@ func NewCommandChannelService(cr *operatorv1beta1.Cryostat) *corev1.Service {
...
 
 func NewGrafanaService(cr *operatorv1beta1.Cryostat) *corev1.Service {
+	// Initialize with defaults
+	svcType := corev1.ServiceTypeClusterIP
+	httpPort := int32(3000)
+	labels := map[string]string{}
+	annotations := map[string]string{}
+
+	// Override with any user-specified properties
+	if cr.Spec.ServiceOptions != nil && cr.Spec.ServiceOptions.GrafanaConfig != nil {
```
andrewazores, line 752: Can this if-statement pattern and the various variables it assigns be extracted out to a function that returns either multiple values or some structure that has fields for each of the values? Then it can be reused here and in `NewCoreService` - DRY

api/v1beta1/cryostat_types.go
ebaron: Oops, good catch.

internal/controllers/common/resource_definitions/resource_definitions.go
ebaron: Right, I've removed this redundant check.

internal/controllers/common/resource_definitions/resource_definitions.go
ebaron: I'm not sure how best to go about this. This method gets called each iteration of the reconcile loop, regardless of whether it ends up creating or changing the service. With either an Event or log message, it could get noisy. It's at least documented in the property description.

internal/controllers/common/resource_definitions/resource_definitions.go
ebaron: I tried refactoring this into something nicer, but didn't succeed. I'll make another attempt.

internal/controllers/common/resource_definitions/resource_definitions.go
andrewazores: Ah, yea that's true.

internal/controllers/common/resource_definitions/resource_definitions.go
ebaron: I think I've come up with a good refactoring of this. What do you think?

ebaron:
One thing I should note: this changes the default service type from LoadBalancer to ClusterIP. This means the services are only accessible outside the cluster using the Route or Ingress for them.

internal/controllers/common/resource_definitions/resource_definitions.go
andrewazores: Looks better, thanks

andrewazores:
> You can test this by creating a Cryostat CR like this instead of using make create_cryostat_cr:

Would it make sense to add some example `serviceConfig` to `config/samples/operator_v1beta1_cryostat.yaml`? Maybe just using the default port numbers and adding some annotations/labels, much like the one you gave in the original PR text.

ebaron:
> > You can test this by creating a Cryostat CR like this instead of using make create_cryostat_cr:
> 
> Would it make sense to add some example `serviceConfig` to `config/samples/operator_v1beta1_cryostat.yaml`? Maybe just using the default port numbers and adding some annotations/labels, much like the one you gave in the original PR text.

The problem with doing this is the sample CR ends up in the CSV's `alm-examples` annotation. The OpenShift console UI then treats this as default values, preloading these values when the user goes to create a CR. So we need to be conservative about the samples and treat them like the defaults they actually are.

andrewazores (commented):
LGTM other than that last q about the sample CR

andrewazores:
> > > You can test this by creating a Cryostat CR like this instead of using make create_cryostat_cr:
> > 
> > 
> > Would it make sense to add some example `serviceConfig` to `config/samples/operator_v1beta1_cryostat.yaml`? Maybe just using the default port numbers and adding some annotations/labels, much like the one you gave in the original PR text.
> 
> The problem with doing this is the sample CR ends up in the CSV's `alm-examples` annotation. The OpenShift console UI then treats this as default values, preloading these values when the user goes to create a CR. So we need to be conservative about the samples and treat them like the defaults they actually are.

Ah, I see. If the `serviceConfig` section is just completely missing then the UI will load everything initialized as `false`/`nil` I guess?

ebaron:
> > > > You can test this by creating a Cryostat CR like this instead of using make create_cryostat_cr:
> > > 
> > > 
> > > Would it make sense to add some example `serviceConfig` to `config/samples/operator_v1beta1_cryostat.yaml`? Maybe just using the default port numbers and adding some annotations/labels, much like the one you gave in the original PR text.
> > 
> > 
> > The problem with doing this is the sample CR ends up in the CSV's `alm-examples` annotation. The OpenShift console UI then treats this as default values, preloading these values when the user goes to create a CR. So we need to be conservative about the samples and treat them like the defaults they actually are.
> 
> Ah, I see. If the `serviceOptions` section is just completely missing then the UI will load everything initialized as `false`/`nil` I guess?

Right, it'll be empty.

> 
> The sample CR does have ex. `spec.storageOptions.pvc.labels: {}` in it - can that be removed then? Or should a "blank outline" like this also be added for `serviceOptions`?

I think all could be removed except for `minimal` and `enableCertManager`, which would match our Initialization Resource annotation: https://github.com/cryostatio/cryostat-operator/blob/main/bundle/manifests/cryostat-operator.clusterserviceversion.yaml#L65-L76

andrewazores (approved):

Issue #279, opened 2021-10-12
Release on OperatorHub.io


ebaron:
We should make Cryostat Operator releases available on https://operatorhub.io/

ebaron:
A guide on contributing an operator to OperatorHub: https://operatorhub.io/contribute. We will be using the Bundle format, not the PackageManifest format.

jiekang:
The docs seem to be outdated in one area or another as https://k8s-operatorhub.github.io/community-operators/packaging-operator/ suggests `semver-mode` being the default, as well as another mode, `semver-skippatch`.

After reading the explanation I linked, I recommend `semver-skippatch` or `semver-mode`

ebaron:
> After reading through the contribution docs, Hareet and I had a few questions:
> 
>     1. Which version of the cryostat operator bundle should we use? This one from the v2 branch? https://github.com/cryostatio/cryostat-operator/tree/v2

You'll want the `v2.0.0` tag. https://github.com/cryostatio/cryostat-operator/tree/v2.0.0
> 
>     2. The [Operator Publishing / Review Settings](https://k8s-operatorhub.github.io/community-operators/operator-ci-yaml/) > Operator Versioning lists two version modes: `updateGraph: replaces-mode` and `updateGraph: semver-mode`. Which mode should we choose and why?

I think `semver-mode` is the way to go. If I understand the docs correctly, `replaces-mode` means we'll have to manually manage the update graph. `semver-skippatch` may be fine but complicates testing by adding additional upgrade paths (e.g. need to ensure both 2.0.0 -> 2.0.1 -> 2.1.0 and 2.0.0 -> 2.1.0 work).

> 
>     3. We were wondering what the `replaces-mode` description meant by saying, "It means, that there is a possibility to omit some versions from the update graph. Best practice is to put them to a separate channel then." We also found some [channel naming conventions](https://olm.operatorframework.io/docs/best-practices/channel-naming/) and were wondering how the channel name `alpha` (from [bundle.Dockerfile](https://github.com/cryostatio/cryostat-operator/blob/v2/bundle.Dockerfile)) affects the operator versioning?

We've been using the `alpha` for nightly builds essentially. In the `v2.0.0` tag, you'll see we use the `stable` channel instead. Users subscribing to the stable channel should get whatever the newest release is, and be able to upgrade as newer releases come out.

> 
[...]

PR #314, opened 2021-12-07
docs(about): Match operatorHub naming convention

jan-law:
Related #279 

The OperatorHub PR checklist suggested adding a "description about the managed Application and where to find more information". @hareetd and I decided to add a link to the [cryostat.io](https://cryostat.io/) website as well as update the `displayName` to `Cryostat Operator` to match the naming convention of the operators on OperatorHub.io. For consistency, I've suggested the same changes in this PR.

image
image


ebaron (approved):
Good idea to add the website. The name change seems reasonable, were there any guidelines about it? I do see that most operators include it in their name on OperatorHub.io.

jan-law:
The docs didn't explicitly specify a name change, but the Bundle Format section under [Package your Operator](https://operatorhub.io/contribute#Package-your-Operator) mentioned "the bundle format has a top-level directory named after your Operator name in the CSV.", and our bundle files all start with "cryostat-operator". The OperatorHub PR checklist also referred to an "Application" and an "Operator" as separate items, so we concluded the name "Cryostat" referred to the Application and "Cryostat Operator" referred to the Operator. 
I'm not sure what the OpenShift naming guidelines are though, so let me know if you think Cryostat is more appropriate.

ebaron:
I think it's fine to go with "Cryostat Operator". You make a good point about the distinction between the application and operator names. cert-manager is itself an operator, but it also used to have an operator that deployed cert-manager. It was unclear whether what I was installing gave me cert-manager itself, or if I had to create a CR first to have the operator deploy cert-manager.

PR #319, opened 2021-12-16
feat(auth): Add logout permissions to service account

jan-law:
Related https://github.com/cryostatio/cryostat/issues/717

Adds permissions needed for the backend to `list` and `delete` an `oauthaccesstoken`. 

I believe the `oc logout` command in the OpenShift docs corresponds to the cryostat frontend `POST https://oauth-server-namespace-route/logout` request that removes the session cookie stored in the OAuth server. This `/logout` call invalidates the token but does not delete any existing invalid tokens, which is why the backend needs to make a second request to delete the access token.

[OpenShift OAuth logout docs](https://docs.openshift.com/container-platform/4.9/authentication/managing-oauth-access-tokens.html#oauth-delete-tokens_managing-oauth-access-tokens)
[OpenShift OAuth endpoint for deleting access tokens](https://docs.openshift.com/container-platform/4.9/rest_api/oauth_apis/oauthaccesstoken-oauth-openshift-io-v1.html#oauthaccesstoken-oauth-openshift-io-v1)

### DIFF
--- a/bundle/manifests/cryostat-operator-cryostat_rbac.authorization.k8s.io_v1_clusterrole.yaml
+++ b/bundle/manifests/cryostat-operator-cryostat_rbac.authorization.k8s.io_v1_clusterrole.yaml
@@ -16,3 +16,10 @@ rules:
   - selfsubjectaccessreviews
   verbs:
   - create
+- apiGroups:
+  - oauth.openshift.io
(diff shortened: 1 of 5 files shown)


ebaron (approved):
Looks good!

Issue #323, opened 2022-01-04
Operator has vulnerable dgrijalva/jwt-go dependency


ebaron:
A transitive dependency of Kubernetes client-go, dgrijalva/jwt-go, is vulnerable to CVE-2020-26160.

Issue #287, opened 2021-11-01
Upgrade Operator SDK to at least 1.8.2


ebaron:
The operator's service account should not be included in the bundle. It should instead be created and managed by OLM. Having this service account present in the bundle results in the following scorecard failure:
`Error: Value cryostat-operator-service-account: invalid service account found in bundle. sa name cannot match service account defined for deployment spec in CSV`

Operator SDK excludes the operator's service account when generating bundles with this fix: https://github.com/operator-framework/operator-sdk/pull/5120. This fix is present in 1.8.2, 1.9.2, 1.10.1, 1.11.0+. We should upgrade to one of these versions.

ebaron:
We are currently using Operator SDK 1.5.0. The migration guide outlines steps to be taken when upgrading: https://sdk.operatorframework.io/docs/upgrading-sdk-version/

tthvo:
Hi @ebaron I am wondering which release version we are aiming to upgrade to? 

From the test suite of the operator-community repo, they are using `v1.16.0`. Would `v1.16.0` be our target?

https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/992cb71004646184da86688a48466e7de84dba4b/roles/install_operator_prereqs/defaults/main.yml#L3

ebaron:
I would say target the newest release, unless some major incompatibility is mentioned in the migration docs between 1.16 and the newest.

PR #324, opened 2022-01-04
build(deps): update dgrijalva/jwt-go to patched version

ebaron:
This PR uses module replacement to patch the vulnerable transitive dependency, now maintained in a fork. Once we upgrade the Operator SDK (#287), we can upgrade client-go and remove this replacement.

Fixes: #323


andrewazores (approved):
Patch is pretty straightforward, builds and appears to run as expected.

PR #325, opened 2022-01-06
feat(cryostat): implement sidecar report config

### LINKED ISSUE #327: Services should not only use `"app": cr.Name` label selector
ex. https://github.com/cryostatio/cryostat-operator/blob/91179772f4277810d4d82f4f21f0c5c57acf7c43/internal/controllers/common/resource_definitions/resource_definitions.go#L707

The Cryostat "core" service, command-channel service, and Grafana service all have labels `app: cr.Name` and a `component: [cryostat|command-channel|grafana]`. The `Selector` on each service is defined as `app: cr.Name`. This [means](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#service-and-replicationcontroller) that the services target all `Pod`s with the `app: cr.Name` label as well - currently this makes sense because the only such pod is the Cryostat instance pod itself. This breaks in #325 because a new `Pod` appears for each -reports sidecar replica - these `Pod`s do have the label `app: cr.Name` since they are part of the same Cryostat instance, but they should not be considered "targeted" by the main Cryostat service or the Grafana service. IMO the fix is pretty simple - the label selector on each service should instead be the `component: [cryostat|grafana]` label, not the `app`.

```yaml
# oc get -o yaml endpoints cryostat-sample
apiVersion: v1
kind: Endpoints
metadata:
  annotations:
    endpoints.kubernetes.io/last-change-trigger-time: "2022-01-11T19:13:41Z"
  creationTimestamp: "2022-01-11T19:10:02Z"
  labels:
    app: cryostat-sample
    component: cryostat
  name: cryostat-sample
  namespace: default
  resourceVersion: "224138"
  uid: a76488aa-cd7c-4023-b5a6-d14b26d3faf9
subsets:
- addresses:
  - ip: 10.217.0.110
    nodeName: crc-ktfxm-master-0
    targetRef:
      kind: Pod
      name: cryostat-sample-576c465fcb-4bzzp
      namespace: default
      resourceVersion: "224130"
      uid: 68f1433d-f585-4c43-b116-85bece4dc1a8
  - ip: 10.217.0.111
    nodeName: crc-ktfxm-master-0
    targetRef:
      kind: Pod
      name: cryostat-sample-reports-67c8b658c7-6nkfb
      namespace: default
      resourceVersion: "224000"
      uid: 73320601-4add-4f6a-9924-bbe8b4a3cba7
  - ip: 10.217.0.112
    nodeName: crc-ktfxm-master-0
    targetRef:
      kind: Pod
      name: cryostat-sample-reports-67c8b658c7-7pnnk
      namespace: default
      resourceVersion: "223986"
      uid: f4226520-6ecf-4f76-87e4-ce061889b6d9
  ports:
  - name: http
    port: 8181
    protocol: TCP
  - name: jfr-jmx
    port: 9091
    protocol: TCP
```

```yaml
# oc get -o yaml endpoints cryostat-sample-grafana
apiVersion: v1
kind: Endpoints
metadata:
  annotations:
    endpoints.kubernetes.io/last-change-trigger-time: "2022-01-11T19:13:41Z"
  creationTimestamp: "2022-01-11T19:10:01Z"
  labels:
    app: cryostat-sample
    component: grafana
  name: cryostat-sample-grafana
  namespace: default
  resourceVersion: "224136"
  uid: 657d5046-2665-447d-a5d2-b39241d7c70c
subsets:
- addresses:
  - ip: 10.217.0.110
    nodeName: crc-ktfxm-master-0
    targetRef:
      kind: Pod
      name: cryostat-sample-576c465fcb-4bzzp
      namespace: default
      resourceVersion: "224130"
      uid: 68f1433d-f585-4c43-b116-85bece4dc1a8
  - ip: 10.217.0.111
    nodeName: crc-ktfxm-master-0
    targetRef:
      kind: Pod
      name: cryostat-sample-reports-67c8b658c7-6nkfb
      namespace: default
      resourceVersion: "224000"
      uid: 73320601-4add-4f6a-9924-bbe8b4a3cba7
  - ip: 10.217.0.112
    nodeName: crc-ktfxm-master-0
    targetRef:
      kind: Pod
      name: cryostat-sample-reports-67c8b658c7-7pnnk
      namespace: default
      resourceVersion: "223986"
      uid: f4226520-6ecf-4f76-87e4-ce061889b6d9
  ports:
  - name: http
    port: 3000
    protocol: TCP
```

andrewazores:
Fixes https://github.com/cryostatio/cryostat-operator/issues/328
Fixes #327
Depends on https://github.com/cryostatio/cryostat/pull/779

`make DEPLOY_NAMESPACE=$(oc project -q) CORE_IMG=quay.io/andrewazores/cryostat:sidecar-reports-1 IMG=quay.io/andrewazores/cryostat-operator:latest deploy`

This adds configuration to the Cryostat CRD for a number of reports replica sidecars and the resource request/limits to apply to each replica, as described in #328 . The `cryostat_controller` reconciles this by creating, updating, or deleting the relevant Deployment and Service for the `reports` component. If these exist then the main Cryostat container is updated with the `CRYOSTAT_REPORT_GENERATOR` environment variable set to point to the URL of the `reports` `Service`. If they don't exist then the main Cryostat container is updated so that the environment variable is not set, allowing it to fall back on subprocess report generation.


andrewazores:
@ebaron I think I'm mostly done with this PR, although I still need to add tests.

I have one strange bug that appears when I create a Cryostat CR, ex. `make create_cryostat_cr`, if that CR has >0 `replicas` configured. When I do this, the `-reports` `Service`/`Deployment`/`Pod` are all created successfully and the `Pod` quickly spins up and enters the `Running` state. The main Cryostat instance however gets stuck on `Starting` with `2/3 Ready`. `oc describe` on the Cryostat pod shows:

```
...
Containers:
  cryostat-sample:
    Container ID:   cri-o://ea4119888b4c14455c86f57a5f85c1441c6e5ac478f19bc4767419a7b062d0c4
    Image:          quay.io/andrewazores/cryostat:sidecar-reports-1
    Image ID:       quay.io/andrewazores/cryostat@sha256:795a3bd31ae201bd3d2d56cc9da5eff2117b7424e108be8b19b0dd03710c36f0
    Ports:          8181/TCP, 9090/TCP, 9091/TCP
    Host Ports:     0/TCP, 0/TCP, 0/TCP
    State:          Running
      Started:      Tue, 11 Jan 2022 14:54:45 -0500
    Last State:     Terminated
      Reason:       Error
      Exit Code:    143
      Started:      Tue, 11 Jan 2022 14:51:45 -0500
      Finished:     Tue, 11 Jan 2022 14:54:44 -0500
    Ready:          False
    Restart Count:  2
    Liveness:       http-get https://:8181/health delay=0s timeout=1s period=10s #success=1 #failure=3
    Startup:        http-get https://:8181/health delay=0s timeout=1s period=10s #success=1 #failure=18
...
Events:
  Type     Reason          Age                     From               Message
  ----     ------          ----                    ----               -------
  Normal   Scheduled       7m41s                   default-scheduler  Successfully assigned default/cryostat-sample-576c465fcb-frgvf to crc-ktfxm-master-0
  Normal   AddedInterface  7m39s                   multus             Add eth0 [10.217.0.131/23] from openshift-sdn
  Normal   Pulled          7m39s                   kubelet            Container image "quay.io/andrewazores/cryostat:sidecar-reports-1" already present on machine
  Normal   Created         7m38s                   kubelet            Created container cryostat-sample
  Normal   Started         7m38s                   kubelet            Started container cryostat-sample
  Normal   Pulled          7m38s                   kubelet            Container image "quay.io/cryostat/cryostat-grafana-dashboard:2.0.0" already present on machine
  Normal   Created         7m37s                   kubelet            Created container cryostat-sample-grafana
  Normal   Started         7m37s                   kubelet            Started container cryostat-sample-grafana
  Normal   Pulled          7m37s                   kubelet            Container image "quay.io/cryostat/jfr-datasource:2.0.0" already present on machine
  Normal   Created         7m36s                   kubelet            Created container cryostat-sample-jfr-datasource
  Normal   Started         7m36s                   kubelet            Started container cryostat-sample-jfr-datasource
  Warning  Unhealthy       2m30s (x30 over 7m30s)  kubelet            Startup probe failed: Get "https://10.217.0.131:8181/health": net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

Oddly enough however, `oc logs` on the pod shows that it is receiving what appear to be startup probe attempts and is responding as expected:

```
Jan 11, 2022 7:57:34 PM io.cryostat.core.log.Logger info
INFO: (10.217.0.1:51680): GET /health 200 975ms
```

If I do `oc edit cryostat` and reduce `replicas` to 0 then this leads to the instance being redeployed and successfully entering the `Ready` state. If I then do `oc edit cryostat` once more and increase `replicas` back to 1+, it again redeploys, but this time everything works as expected and the main Cryostat instance container does also enter the `Ready` state. Any idea what might be going wrong here?

andrewazores:
Ah, Deployment selectors... yea, maybe that's it. I'll try adjusting those tomorrow and see what happens.

andrewazores:
> Might this be related to #327? Not for the services but for the Deployment selectors? If the Deployments are getting confused about which pods belong to it.

I think that did do the trick - it now seems to work as expected for me in all cases of creating/editing/deleting the CR, regardless of the number of replicas specified.

api/v1beta1/cryostat_types.go
```diff
@@ -92,6 +94,17 @@ type StorageConfiguration struct {
...
 
+// ReportConfiguration is used to determine how many replicas of cryostat-reports
+// the operator should create and what the resource limits of those containers
+// should be. If no replicas are created then Cryostat is configured to use basic
+// subprocess report generation. If at least one replica is created then Cryostat
+// is configured to use remote report generation, pointed at a load balancer service
+// in front of the cryostat-reports replicas.
+type ReportConfiguration struct {
+	Replicas                    int32 `json:"replicas,omitempty"`
+	corev1.ResourceRequirements `json:",inline"`
```
ebaron, line 105: Could you add a comment for Replicas? It will be shown to the user when creating a CR using the form view. Also I think some OLM Descriptors make sense here:
- [urn:alm:descriptor:com.tectonic.ui:podCount](https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#1-podcount)
- [urn:alm:descriptor:com.tectonic.ui:resourceRequirements](https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#2-resourcerequirements)

config/default/image_tag_patch.yaml
```diff
@@ -15,3 +15,5 @@ spec:
           value: "quay.io/cryostat/jfr-datasource:2.0.0"
         - name: RELATED_IMAGE_GRAFANA
           value: "quay.io/cryostat/cryostat-grafana-dashboard:2.0.0"
+        - name: RELATED_IMAGE_REPORTS
+          value: "quay.io/cryostat/cryostat-reports:1.0.0-SNAPSHOT
```
ebaron, line 19: It looks like there's a missing end-quote here. Did `make manifests` generate this? If so, there must be a bug with it.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -774,6 +900,35 @@ func NewGrafanaService(cr *operatorv1beta1.Cryostat) *corev1.Service {
...
+			Ports: []corev1.ServicePort{
+				{
+					Port:       reportsPort,
+					TargetPort: intstr.IntOrString{IntVal: reportsPort},
+				},
+			},
+		},
+	}
+}
+
```
ebaron, line 931: This should be customizable like the other services are now: https://github.com/cryostatio/cryostat-operator/blob/0ee99b4242c5b53d2186ce56ec5af6531179b36f/api/v1beta1/cryostat_types.go#L111-L144

We can do this in a followup PR if you prefer.

internal/controllers/cryostat_controller.go
```diff
@@ -397,6 +406,59 @@ func (r *CryostatReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		op, err := controllerutil.CreateOrUpdate(ctx, r.Client, deployment, func() error {
+			deployment.Spec.Template.Spec = podTemplate.Spec
+			deployment.Spec.Replicas = &desired
+			return nil
+		})
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		serviceSpecs.ReportsURL = &url.URL{
+			Scheme: "http",
```
ebaron, line 454: We should create issue(s) to set up secure traffic between Cryostat and the report generator using cert-manager.

ebaron (commented):
Hey Andrew, code changes look good overall, just a few comments below.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -303,6 +353,70 @@ func NewPodForCR(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTags *I
...
+	if requests := resources.Requests; requests != nil {
+		if cpu := requests.Cpu(); cpu != nil {
+			cpus = cpu.Value()
+		}
+	}
+	if limits := resources.Limits; limits != nil {
+		if cpu := limits.Cpu(); cpu != nil {
+			cpus = cpu.Value()
+		}
+	}
```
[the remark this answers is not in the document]
andrewazores: For this case yes. If the limits are present they should be greater than or equal to the requests, so I think this is effectively equivalent to taking the maximum of the two values. Whichever is the maximum is what we tell the JVM is how much CPU it can expect to be able to use. Maybe that should ignore the limits and only use the requests?

internal/controllers/common/resource_definitions/resource_definitions.go
andrewazores: If this and the cert-manager integration are the only two major changes to be done then I don't mind doing this change as part of this PR, and cert-manager in a follow-up.

andrewazores:
Review comments mostly all addressed in the latest commits. `cert-manager` isn't handled yet, and the question about how to handle requests vs limits is still open.

docs/config.md
```diff
@@ -168,4 +189,22 @@ spec:
...
+          http:
+            paths:
+            - path: /
+              pathType: Prefix
+              backend:
+                service:
+                  name: cryostat-sample-reports
+                  port:
+                    number: 10000
+```
```
ebaron, line 209: I think there is no Ingress config for the reports service, so this change can be removed. We don't need to expose the report generator service outside the cluster do we?

andrewazores: True, no, I don't think there is really any reason to expose the service outside the cluster. Clients can make requests through the normal Cryostat HTTP API and it'll get internally routed to the sidecar containers.

config/samples/operator_v1beta1_cryostat.yaml
```diff
@@ -12,3 +12,8 @@ spec:
       labels: {}
       annotations: {}
       spec: {}
+  reportOptions:
+    replicas: 1
+    requests:
+      cpu: 1000m
+      memory: 512Mi
```
ebaron, line 19: Could you update the CSV with `make bundle`? Also if these values should be made defaults in the form view, could you add them to the CR in this annotation:
https://github.com/cryostatio/cryostat-operator/blob/0ee99b4242c5b53d2186ce56ec5af6531179b36f/config/manifests/bases/cryostat-operator.clusterserviceversion.yaml#L56-L67

andrewazores: Yea, I'm not sure what the defaults should be - that really depends on the size of the recordings a user is generally handling. Maybe in the interest of keeping the initial deployed footprint small the default should be 0 replicas and no resources config.

ebaron: Makes sense

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -761,7 +886,8 @@ func NewGrafanaService(cr *operatorv1beta1.Cryostat) *corev1.Service {
 		Spec: corev1.ServiceSpec{
 			Type: *config.ServiceType,
 			Selector: map[string]string{
-				"app": cr.Name,
+				"app":       cr.Name,
+				"component": "grafana",
```
ebaron, line 890: I noticed the selector with component doesn't work. All 3 services select the same pod, but the deployment's pod template has `component: cryostat`. We end up with the Grafana container running normally, but the service doesn't route traffic to it.

andrewazores: Ah, right...

andrewazores: What do you think is the best way through this then? Should I set the `component` for the Grafana service to just be `cryostat`?

ebaron: We could do that, or just skip the `component` label and use `app: cr.Name + "-reports"` for the reports deployment and service.

ebaron (approved):
Looks good!

PR #329, opened 2022-01-13
docs(ingress): Remove unneeded ingress timeouts

jan-law:
Related #757

This reverts commit e524857960ffdfe44eaa360e9fd215d14b968953.

Now that the server keepalive pings in https://github.com/cryostatio/cryostat/pull/792 is merged, we can remove the extra ingress annotations to extend the idle timeout.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -120,8 +120,6 @@ spec:
     coreConfig:
       annotations:
         nginx.ingress.kubernetes.io/backend-protocol : HTTPS
-        nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
-        nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
       ingressSpec:
         tls:
         - {}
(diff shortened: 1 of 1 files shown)


jan-law:
Since v2 cryostat users might still benefit from these timeout annotations, how would you recommend updating `config.md`?

andrewazores:
Hmm. It's unfortunate that the `docs/config.md` in the `v2` branch doesn't contain these annotations. Maybe it's worth adding a backport documentation commit to that branch just for that reason. I would rather not have the documents themselves reference various versioned changes - we should just be able to rely on version control to keep the doc in sync with the source it accompanies. @ebaron WDYT?

ebaron:
> Hmm. It's unfortunate that the `docs/config.md` in the `v2` branch doesn't contain these annotations. Maybe it's worth adding a backport documentation commit to that branch just for that reason. I would rather not have the documents themselves reference various versioned changes - we should just be able to rely on version control to keep the doc in sync with the source it accompanies. @ebaron WDYT?

Agreed. I just asked Mergify to backport @jan-law's original PR. I'm glad we're now at the point these are no longer needed and the WebSocket connection seems to be much more resilient.

ebaron (approved):

PR #333, opened 2022-01-17
feat(printer-column): Printer Column for Application URL

jaadbarg:
Fixes #309

Issue #309, opened 2021-11-22
Printer Column for Application URL


ebaron:
It would be a simple and useful improvement to add a printer column for the application URL of a Cryostat CR, so `kubectl get cryostats` would print the associated application URL. See: https://book.kubebuilder.io/reference/generating-crd.html#additional-printer-columns

PR #334, opened 2022-01-18
fix(printer-column): add printer column to CSV

ebaron:
Regenerated bundle after adding `additionalPrinterColumns` metadata in #333. 

Related to: #309

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=cryostat-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.6.1+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.4.0+git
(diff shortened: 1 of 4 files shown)


bundle.Dockerfile
```diff
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=cryostat-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.6.1+git
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.4.0+git
```
andrewazores, line 10: I did `make generate manifests bundle` on current `main` and got the same `additionalPrinterColumns` patch chunk as below, but I didn't get any of these `builder=operator-sdk` changes. Which invocation causes these and are these labels/annotations meaningful?

ebaron: It's just from `make bundle`. You must be using a newer `operator-sdk` binary than I am. I'm using 1.5.0, but for some reason it reports as `1.4.0+git`. #325 changed these labels to their current values, this is just changing them back.

ebaron: I'm not sure if it serves a purpose beyond documenting the version the operator was built with.

andrewazores (approved):

PR #335, opened 2022-01-20
feat(env): Make selected env vars configurable

### LINKED ISSUE #278: Cryostat environment variables should be configurable
We should allow the user to supply environment variables to the Cryostat deployment, either by adding them individually to the Cryostat CRD where it makes sense, or by accepting an arbitrary list of environment variables in the CRD.

jan-law:
Fixes #278 

Environment variables added to the Cryostat CRD:
- `CRYOSTAT_REPORT_GENERATION_MAX_HEAP`. This is only set when the sidecar `reportsUrl` does not exist
- `CRYOSTAT_MAX_WS_CONNECTIONS`, default 2
- `CRYOSTAT_TARGET_CACHE_SIZE`, default -1
- `CRYOSTAT_TARGET_CACHE_SIZE`, default 10

> I wonder if CRYOSTAT_ALLOW_UNTRUSTED_SSL=true would ever be needed. jfr-datasource doesn't use HTTPS. The reports sidecar doesn't yet, but Cryostat talks to it using the internal service hostname whose certificates would be managed by cert-manager.

I also removed `CRYOSTAT_ALLOW_UNTRUSTED_SSL`.


ebaron (commented):
Hi Janelle, really sorry it took me so long to get to this. I was playing around with using [`resource.Quantity`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity) for the report subprocess heap size, but it didn't pan out.

I have a few suggestions to add some validation to the fields below. I also think each of these new fields could use the [`number`](https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#7-number) OLM spec descriptor. I think this would prevent the user from entering in non-numeric text.

Aside from that, would you mind adding a test where these options are set to non-default values?

jan-law:
No worries Elliott, thanks for the review

ebaron (approved):
Looks good! Thanks Janelle!

Issue #332, opened 2022-01-14
Add Resource Requirements for Cryostat deployment


ebaron:
A user should be able to specify resource requests/limits for the main Cryostat deployment using the Cryostat CRD.

andrewazores:
Related to #277

jaadbarg:
Hi @ebaron, I'd like to take this one

ebaron:
ResourceRequirements type: https://pkg.go.dev/k8s.io/api/core/v1#ResourceRequirements

ebaron:
OLM Spec Descriptor: https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#2-resourcerequirements

ebaron:
Resource requirements are set at the container level, so we should expose properties for each of the 3 containers. Something like this perhaps:

```go
type CryostatSpec struct {
...
	// Define resource requirements for containers within the main Cryostat deployment
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	Resources ResourceConfigList `json:"resources,omitempty"`
}

// ResourceConfigList includes resource requirements settings for each container in the main Cryostat deployment
type ResourceConfigList struct {
	// Resource requirements for the Cryostat application
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
	CoreResources corev1.ResourceRequirements `json:"coreResources,omitempty"`
	// Resource requirements for the JFR Data Source container
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
	DataSourceResources corev1.ResourceRequirements `json:"dataSourceResources,omitempty"`
	// Resource requirements for the Grafana container
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
	GrafanaResources corev1.ResourceRequirements `json:"grafanaResources,omitempty"`
}
```

PR #344, opened 2022-02-25
feat(resources): Add Resource Config for Main Deployment

jaadbarg:
fixes: #332


ebaron (commented):
Hey John, I had a new idea about unit testing this feature. We could add a new test Context for testing the case when a user specifies resource requirements, such as:
```go
Context("with resource requirements", func() {
	BeforeEach(func() {
		t.objs = append(t.objs, test.NewCryostatWithResources())
	})
	It("should create expected deployment", func() {
		t.expectDeployment()
	})
})
```

Then you'd need to add a `NewCryostatWithResources` function inside test/resources.go. It can call `NewCryostat`, but add in resource requirements for each type of container before returning.

To add expectations about the resource requirements present in the deployment, you can modify `checkMainDeployment`:
https://github.com/cryostatio/cryostat-operator/blob/3948e7dedd0c25f1de14a47f06dcde3196a014fd/internal/controllers/cryostat_controller_test.go#L1657-L1714

Before the call to `checkCoreContainer`, the Cryostat instance is available in the variable `cr`. You can add a new parameter to `checkCoreContainer` of type `corev1.ResourceRequirements`. Then pass the core resource requirements from `cr` to this function. Then you can add an expectation comparing the argument to the resource requirements found in the container definition. Repeat this for `checkDatasourceContainer` and `checkGrafanaContainer`.

api/v1beta1/cryostat_types.go
```diff
@@ -86,6 +86,25 @@ type CryostatSpec struct {
...
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="JMX Connections Cache Options"
 	JmxCacheOptions *JmxCacheOptions `json:"jmxCacheOptions,omitempty"`
+	// Resource requirements for the Cryostat deployment
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
+	Resources ResourceConfigList `json:"resources,omitempty"`
+}
+
+type ResourceConfigList struct {
+	// Resource requirements for the Cryostat application
```
ebaron, line 96: Hey John, would you mind amending this comment as follows:
```suggestion
	// Resource requirements for the Cryostat application. If specifying a memory limit, at least 768MiB is recommended.
```

We recently encountered some out-of-memory issues during testing, so this would help users. Once added, please re-run `make bundle`.

api/v1beta1/cryostat_types.go
```diff
@@ -87,6 +87,25 @@ type CryostatSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="JMX Connections Cache Options"
 	JmxCacheOptions *JmxCacheOptions `json:"jmxCacheOptions,omitempty"`
+	// Resource requirements for the Cryostat deployment
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
```
ebaron, line 92: This should be:
```suggestion
	// +operator-sdk:csv:customresourcedefinitions:type=spec
```
Having the descriptor here prevents the Core/DataSource/Grafana resources from showing in the UI.

ebaron: This will need another `make bundle` after changing.

jaadbarg:
@ebaron for some reason when I run `make bundle` in `cryostat-operator.clusterserviceversion.yaml`,

`operators.operatorframework.io/builder: operator-sdk-v1.4.0+git
    operators.operatorframework.io/internal-objects: '["flightrecorders.operator.cryostat.io","recordings.operator.cryostat.io"]'`

gets changed to 

`    operators.operatorframework.io/builder: operator-sdk-v1.5.0
`

and that's causing the merge conflict

ebaron (approved):
Looks good, thanks John!

PR #348, opened 2022-03-01
fix(WebSocket): Update ws limit

### LINKED ISSUE #342: Update docs for max WS connections env var
It'd be a good idea to update the operator docs before the 2.1 release too:
https://github.com/cryostatio/cryostat-operator/blob/main/docs/config.md#cryostat-client-options
https://github.com/cryostatio/cryostat-operator/blob/main/api/v1beta1/cryostat_types.go#L83

_Originally posted by @jan-law in https://github.com/cryostatio/cryostat/pull/832#pullrequestreview-883726129_

jan-law:
Fixes #342

### DIFF
--- a/api/v1beta1/cryostat_types.go
+++ b/api/v1beta1/cryostat_types.go
@@ -78,11 +78,10 @@ type CryostatSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	ReportOptions *ReportConfiguration `json:"reportOptions,omitempty"`
-	// The maximum number of WebSocket client connections allowed (minimum 1, maximum 64, default 2)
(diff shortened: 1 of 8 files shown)


ebaron (commented):
Hi Janelle! I was thinking a bit about how to approach this change. I think this might get confusing since `Integer.MAX_VALUE` refers to a Java constant. What about referring to the default as "unlimited", and just omitting the maximum in our description.

We could change `MaxWsConnections` to a `*int32` and only set `CRYOSTAT_MAX_WS_CONNECTIONS` if non-nil. This would cause Cryostat to use its "unlimited" max connections, unless the user explicitly sets `spec.maxWsConnections`.

What do you think?

jan-law:
Good point, makes sense. I'll add that in.

andrewazores (commented):
LGTM from the environment variable values POV

ebaron (approved):

Issue #349, opened 2022-03-02
Deprecate FlightRecorder/Recording APIs


ebaron:
These APIs are falling behind in capabilities compared to the Cryostat HTTP API and web client. While it is interesting to have recording management functionality available through CRDs, it's currently too great of a maintenance burden. The operator will focus on deploying and customizing Cryostat.

ebaron:
Deprecating should be fine, but before these can be removed, these CRDs must be replaced in the Cryostat permissions API:
https://github.com/cryostatio/cryostat/blob/42d92207a318b11083332adb671c03b98f958aaa/src/main/java/io/cryostat/net/OpenShiftAuthManager.java#L579-L580

PR #356, opened 2022-03-25
fix(api): deprecate FlightRecorder and Recording CRDs

ebaron:
The CRD spec has properties for whether a version is deprecated and a warning string to show the user when they try to use that version. Newer controller-tools versions support these properties using marker comments, but this requires Operator SDK 1.10 and we're still quite a bit behind that (see #287). In the meantime, I've added Kustomize patches to add these properties.

Example:
```
$ kubectl get recording
Warning: operator.cryostat.io/v1beta1 Recording is deprecated; please use the Cryostat web application or the Cryostat HTTP API to manage recordings
No resources found in cryostat-operator-system namespace.
```

I didn't see this have an effect in the UI, so I've added the `operators.operatorframework.io/internal-objects` annotation to our CSV. This results in the FlightRecorder and Recording APIs being hidden from the OpenShift Console's operator UI. The functionality remains available via command line or through the CustomResourceDefinitions page in the OpenShift Console.

Bundle for testing:
```
make deploy_bundle BUNDLE_IMG=quay.io/ebaron/cryostat-operator-bundle:deprecate-api-01
```

Fixes: #349

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,6 @@ kubectl get secret ${CRYOSTAT_NAME}-jmx-auth -o jsonpath='{$.data.CRYOSTAT_RJMX_
 kubectl get secret ${CRYOSTAT_NAME}-jmx-auth -o jsonpath='{$.data.CRYOSTAT_RJMX_PASS}' | base64 -d
 ```
 
-## Kubernetes API
-The operator provides a Kubernetes API to directly create and manage Flight Recordings.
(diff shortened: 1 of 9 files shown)


andrewazores (approved):

hareetd (approved):
Works well.

jan-law (approved):

Issue #220, opened 2021-07-23
Allow emptyDir volumes instead of PVCs


ebaron:
We should add an option in `Spec.StorageOptions` to deploy Cryostat using `emptyDir` volumes instead of creating a Persistent Volume Claim. If neither are specified, we can retain the same default of creating a PVC.

ebaron:
@jan-law would you be interested in this one?

jan-law:
Sure!

PR #358, opened 2022-03-29
docs(storage): fix EmptyDir option example

jan-law:
Related #220 

I realized I forgot to update `config.md` with the `Enabled` flag after the PR review changes


ebaron (approved):
Good catch!

Issue #353, opened 2022-03-18
Update Dockerfile to Go 1.17 or newer


ebaron:
Our Dockerfile is using 1.15, which is no longer supported.

Issue #361, opened 2022-04-05
GitHub CI workflow not installing controller-gen


ebaron:
The `setup-go` GitHub action now appears to be using Golang 1.18. This changes how `go get` works and no longer installs `controller-gen` [1]. Golang 1.18 is also not supported yet with the Operator SDK: https://github.com/operator-framework/operator-sdk/issues/5481

[1] https://go.dev/doc/go1.18#go-command

PR #362, opened 2022-04-05
build(golang): update to Golang 1.17

ebaron:
This updates the following:
* Minimum Go version to 1.16. This was just recently [EOL'd](https://endoflife.date/go), but many platforms have not yet updated to 1.17 or newer.
* Dockerfile version to 1.17
* CI workflow to 1.17

Fixes: #353, #361


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/cryostatio/cryostat-operator
 
-go 1.15
+go 1.16
```
hareetd, line 3: Does this mean the project will use `go` v1.16 in its dependencies while the `Dockerfile` will use v1.17 to build the image?

ebaron: This specifies that you must be using at least Go 1.16 to build the operator. https://go.dev/doc/modules/gomod-ref#go
If you have Go 1.16 installed locally, then you'll still be able to build the operator locally with `make manager`. We can look at upgrading this to 1.17+ once available on more platforms. For example, Fedora 35 is still using 1.16.
When building the container image (e.g. `make oci-build`), it will build using 1.17.

hareetd: Makes sense, thanks!

hareetd (approved):

jan-law (approved):

Issue #408, opened 2022-06-13
Existing deployments should be updated


ebaron:
Right now the deployment's pod templates and replica count for reports are updated when the current state doesn't match the operator's spec. This should also include the selector, labels and annotations. Since we don't have an API yet to configure deployment labels and annotations, we should merge the operator-managed labels/annotations with any user-specified ones.

PR #416, opened 2022-06-13
feat(deploy): update deployment metadata and selector

ebaron:
Updates pod selector and labels/annotations, in addition to the already handled pod template. Removed the deployment deletion for minimal/non-minimal deployments since this should automatically be handled by updating the deployment spec. Added tests to verify that fields get updated as expected.

Fixes: #408


andrewazores (approved):
Looks good. I tested it out by just toggling on/off the `minimal` flag and it works as expected still.

PR #418, opened 2022-06-14
build(version): set versions for 2.1.1 release

ebaron:
Fixes: https://github.com/cryostatio/cryostat/issues/997
Depends on: https://github.com/cryostatio/cryostat/pull/998

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 SHELL := /bin/bash
 
 # Current Operator version
-IMAGE_VERSION ?= 2.1.0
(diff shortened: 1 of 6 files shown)


andrewazores:
Hmm, seems like the "Dependent Issues" check should be blocking the merge on this PR, no?

andrewazores:
Ah, we have all those status checks configured for the `main` branch only. I think that's probably true across all repos.

andrewazores:
I think I've had "Depends on #PR" block merge before, but maybe I'm misremembering...

PR #421, opened 2022-06-29
fix(crd): fix missing spec descriptor for flightRecorder

### LINKED ISSUE #420: Scorecard test failure for flightRecorder
It appears the test case for flightRecorder fails since it does not have a spec descriptor.

```
Image:      quay.io/operator-framework/scorecard-test:v1.5.0
Entrypoint: [scorecard-test olm-spec-descriptors]
Labels:
	"suite":"olm"
	"test":"olm-spec-descriptors-test"
Results:
	Name: olm-spec-descriptors
	State: fail

	Suggestions:
		Add a spec descriptor for flightRecorder
	Errors:
		flightRecorder does not have a spec descriptor
	Log:
		Loaded ClusterServiceVersion: cryostat-operator.v2.2.0-dev
		Loaded 3 Custom Resources from alm-examples
```

tthvo:
Fixes #420 

Fix flightRecorder missing spec descriptor.


tthvo:
@ebaron any suggestions?

config/manifests/xdescriptors_patch.yaml
```diff
@@ -1,6 +1,9 @@
-- op: replace
-  path: /spec/customresourcedefinitions/owned/2/specDescriptors/3/path
-  value: flightRecorder.name
+- op: add
+  path: /spec/customresourcedefinitions/owned/2/specDescriptors/3
+  value:
+    description: Flight Recorder Name
+    displayName: Name
+    path: flightRecorder.name
```
ebaron, line 6: Can we combine these two patches so that we also add the `x-descriptors` array here?

config/manifests/xdescriptors_patch.yaml
```diff
@@ -1,6 +1,9 @@
-- op: replace
-  path: /spec/customresourcedefinitions/owned/2/specDescriptors/3/path
-  value: flightRecorder.name
+- op: add
+  path: /spec/customresourcedefinitions/owned/2/specDescriptors/3
+  value:
+    description: Flight Recorder Name
```
ebaron, line 4: For the description, maybe we could split the existing description for the `flightRecorder` field into two:
https://github.com/cryostatio/cryostat-operator/blob/7bb319658635c3f220da3d902b09880122f91fba/api/v1beta1/recording_types.go#L67-L68
Remove the second sentence from that file, and use it here. "Select the FlightRecorder with the name of the target Pod for this Recording."

ebaron (commented):
Hey Thuan, looks good. Just a couple of suggestions below. Make sure to re-run `make bundle` after you've made these changes. Thanks!

config/manifests/xdescriptors_patch.yaml
tthvo: @ebaron So the patch would become as below?

```yaml
value:
    description: Select the FlightRecorder with the name of the target Pod for this Recording.
    displayName: Name
    path: flightRecorder.name
    x-descriptors: [ "urn:alm:descriptor:io.kubernetes:operator.cryostat.io:v1beta1:FlightRecorder" ]
```

ebaron: Right

tthvo:
Any other suggestions I should be working on @ebaron?

config/manifests/xdescriptors_patch.yaml
```diff
@@ -1,6 +1,7 @@
...
-  value: flightRecorder.name
 - op: add
-  path: /spec/customresourcedefinitions/owned/2/specDescriptors/3/x-descriptors
-  value: [ "urn:alm:descriptor:io.kubernetes:operator.cryostat.io:v1beta1:FlightRecorder" ]
+  path: /spec/customresourcedefinitions/owned/2/specDescriptors/3
+  value:
+    description: Select the FlightRecorder with the name of the target Pod for this Recording.
+    displayName: Name
+    path: flightRecorder.name
+    x-descriptors: [ "urn:alm:descriptor:io.kubernetes:operator.cryostat.io:v1beta1:FlightRecorder" ]
```
ebaron, line 7: Could you re-save this file with the newline at the end of it? Some editors do not do this by default. If using VSCode, you can enable this behaviour by searching settings for 'Insert Final Newline', and checking it.

tthvo: yes I was looking to how to do this! Thanks a lot! Just configured and added newline!

ebaron (approved):
Fixes the test for me. Good work @tthvo!

PR #422, opened 2022-06-30
feat(license): add license header checking script

### LINKED ISSUE #81: Script to check/update license headers
It would be nice to have a simple script that checks source files for license headers, and adds them if they're missing.

tthvo:
Resolves #81 

Added a script to check and add license header to source files if needed. Run `make add-license`.

We are using `addlicense v1.0.0` from https://github.com/google/addlicense.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,6 +172,7 @@ LICENSE_FILE=$(shell pwd)/LICENSE
 add-license: addlicense
 	@echo "Checking/Adding license..."
 	@$(ADDLICENSE) -v -f $(LICENSE_FILE) api internal 2>&1
+	@go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | sort | uniq | xargs $(ADDLICENSE) -v -f LICENSE
 	@echo "All good"
 
 # Build the OCI image
--- a/Makefile
+++ b/Makefile
@@ -152,7 +152,7 @@ manifests: controller-gen
 
 # Run go fmt against code
 .PHONY: fmt
-fmt:
+fmt: add-license
 	go fmt ./...
 
 # Run go vet against code
@@ -170,9 +170,9 @@ generate: controller-gen
 LICENSE_FILE=$(shell pwd)/LICENSE
 .PHONY: add-license 
 add-license: addlicense
-	@echo "[RUNNING] Checking/Adding license..."
+	@echo "Checking/Adding license..."
 	@$(ADDLICENSE) -v -f $(LICENSE_FILE) api internal 2>&1
-	@echo "[DONE] All good"
+	@echo "All good"
 
 # Build the OCI image
 .PHONY: oci-build
--- a/Makefile
+++ b/Makefile
@@ -171,8 +171,7 @@ LICENSE_FILE=$(shell pwd)/LICENSE
 .PHONY: add-license 
 add-license: addlicense
 	@echo "Checking/Adding license..."
-	@$(ADDLICENSE) -v -f $(LICENSE_FILE) api internal 2>&1
-	@go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | sort | uniq | xargs $(ADDLICENSE) -v -f $(LICENSE_FILE)
+	@go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | sort | uniq | xargs $(ADDLICENSE) -v -f $(LICENSE_FILE) 2>&1
 	@echo "All good"
 
 # Build the OCI image
--- a/Makefile
+++ b/Makefile
@@ -171,8 +171,7 @@ LICENSE_FILE=$(shell pwd)/LICENSE
 .PHONY: add-license 
 add-license: addlicense
 	@echo "Checking/Adding license..."
-	@go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | uniq | xargs $(ADDLICENSE) -v -f $(LICENSE_FILE) 2>&1
-	@echo "All good"
+	@go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | uniq | xargs $(ADDLICENSE) -v -f $(LICENSE_FILE)
 
 # Build the OCI image
 .PHONY: oci-build
--- a/Makefile
+++ b/Makefile
@@ -170,8 +170,9 @@ generate: controller-gen
 LICENSE_FILE=$(shell pwd)/LICENSE
 .PHONY: add-license 
 add-license: addlicense
-	@echo "Checking/Adding license..."
-	@go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | uniq | xargs $(ADDLICENSE) -v -f $(LICENSE_FILE)
+	echo "Checking/Adding license..."
+	$(eval GO_PACKAGES  := $(shell go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | uniq))
+	$(ADDLICENSE) -v -f $(LICENSE_FILE) ${GO_PACKAGES}
 
 # Build the OCI image
 .PHONY: oci-build
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,8 @@ CERT_MANAGER_MANIFEST ?= \
 
 KUSTOMIZE_VERSION ?= 3.8.7
 
+ADDLICENSE_VERSION ?= 1.0.0
+
 DEPLOY_NAMESPACE ?= cryostat-operator-system
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
@@ -164,6 +166,18 @@ generate: controller-gen
 	go generate ./...
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+# Check and add (if missing) license header
+LICENSE_FILE=$(shell pwd)/LICENSE
+.PHONY: add-license 
+add-license: addlicense
+	@echo "Checking/Adding license..."
+	@for f in $$(find . -path "./vendor" -prune -o \( -iname '*.go' ! -iname '*.deepcopy.go' \) -type f -print); \
+	do \
+		$(ADDLICENSE) -check -f $(LICENSE_FILE) $${f} &>/dev/null && echo "[OK] $${f}" && continue; \
+		$(ADDLICENSE) -f $(LICENSE_FILE) $${f} &>/dev/null && echo "[Modified] $${f}"; \
+	done
+	@echo "All good"
+
 # Build the OCI image
 .PHONY: oci-build
 oci-build: generate manifests manager test-envtest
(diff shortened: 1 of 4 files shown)


tthvo:
Though executable is said to recursively search, its `**` only match a single level. So I have to run a loop with `find`. Not sure if its my syntax error or something for their feature release.

Originall use, `addlicense -f LICENSE "**/*.go"`

tthvo:
I wonder if we should also have a check in github workflow to check license too?

andrewazores:
> I wonder if we should also have a check in github workflow to check license too?

Yea, definitely. @ebaron maybe we just hook it in as a dependency for `fmt`? This way it gets run regularly for both GH CI and normal development builds.

api/v1beta1/groupversion_info.go
```diff
@@ -1,18 +1,38 @@
-/*
-Copyright The Cryostat Authors
-
-Licensed under the Apache License, Version 2.0 (the "License");
```
andrewazores, line 4: @ebaron should these have been Apache licensed to begin with?

andrewazores:
> Though the executable is said to recursively search source files, its `**` expression only match a single level. So I have to run a loop with `find`. Not sure if its my syntax error or something for their feature release.
> 
> Originally use, `addlicense -f LICENSE "**/*.go"` but only matches go files at project root (there are none in this case).

Weird. I'm trying out `addlicense` manually and seeing the same thing. Normally the `**` expression means it expands out to any number of subdirectories, but maybe it really is broken here. It'd be nice to not have to do the `find` loop, but that isn't so bad if we do get stuck with it.

api/v1beta1/groupversion_info.go
tthvo: I was not sure about this so I replaced it for review. I could revert it if it is intentional here :D Also I wonder why we have Apache license here tho?

andrewazores:
Ah, the `**` syntax only works for the `-ignore` flag. Reading [the source](https://github.com/google/addlicense/blob/13e73a7f8fcb5696256b6a7b7addabf1070ad4b9/main.go#L210) makes this more clear. The actual check/add functionality only works on directories or filenames, but if you give it a directory then it does search it recursively.

```bash
make addlicense
./bin/addlicense -f LICENSE -check api internal
```

does what's expected. Those are the only two directories that contain `*.go` src files (`find . -type f -name "*.go"`).

tthvo:
> Ah, the `**` syntax only works for the `-ignore` flag. Reading [the source](https://github.com/google/addlicense/blob/13e73a7f8fcb5696256b6a7b7addabf1070ad4b9/main.go#L210) makes this more clear. The actual check/add functionality only works on directories or filenames, but if you give it a directory then it does search it recursively.
> 
> ```shell
> make addlicense
> ./bin/addlicense -f LICENSE -check api internal
> ```
> 
> does what's expected. Those are the only two directories that contain `*.go` src files (`find . -type f -name "*.go"`).

Oh yess! I just read the source and tested too! It did work! It's nice to remove the loop to `find`.

Makefile
```diff
@@ -164,6 +166,14 @@ generate: controller-gen
 	go generate ./...
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+# Check and add (if missing) license header
+LICENSE_FILE=$(shell pwd)/LICENSE
+.PHONY: add-license 
+add-license: addlicense
+	@echo "[RUNNING] Checking/Adding license..."
+	@$(ADDLICENSE) -v -f $(LICENSE_FILE) api internal 2>&1
```
andrewazores, line 174: I guess the downside of this approach is that this script now needs to be updated in the future if any other top-level directory is introduced and needs to be checked, other than `api internal`.

The other option is to go back to using `find` for this, unfortunately. But I think it can be done more cleanly than the first attempt was:

```bash
find . -type f -name "*.go" | xargs ./bin/addlicense -v -f LICENSE
```

@ebaron thoughts? I'd rather avoid the `find` and just let `addlicense` handle walking the tree, but I don't know how likely that `api internal` directory list is to change.

api/v1beta1/groupversion_info.go
ebaron: These files are generated by the Operator SDK. The `operator-sdk init` command has a `--license` parameter, which defaults to Apache 2.0:
```
--license string          license to use to boilerplate, may be one of 'apache2', 'none' (default "apache2")
```
We should be fine to change these headers then.

ebaron:
> > I wonder if we should also have a check in github workflow to check license too?
> 
> Yea, definitely. @ebaron maybe we just hook it in as a dependency for `fmt`? This way it gets run regularly for both GH CI and normal development builds.

I agree with adding it as a Makefile prerequisite to one of our existing targets. `fmt` seems fine.

Makefile
andrewazores: Agreed, `addlicense` doesn't look like it's doing anything like marking directories as already-walked if you pass it a directory as well as some of its subdirectories, so it would end up iterating over the same files multiple times.

```bash
$ go list -f '{{.Dir}}' ./... | sed -e "s|^$(pwd)||" | cut -d/ -f 2 | uniq
api
internal
```

so

```bash
$ go list -f '{{.Dir}}' ./... | sed -e "s|^$(pwd)||" | cut -d/ -f 2 | uniq | xargs ./bin/addlicense -f LICENSE
```

tthvo: Tested and worked for me! This is nicee!

Makefile
```diff
@@ -164,6 +166,15 @@ generate: controller-gen
 	go generate ./...
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+# Check and add (if missing) license header
+LICENSE_FILE=$(shell pwd)/LICENSE
+.PHONY: add-license 
+add-license: addlicense
+	@echo "Checking/Adding license..."
+	@$(ADDLICENSE) -v -f $(LICENSE_FILE) api internal 2>&1
```
andrewazores, line 174: This line can be removed now, right? The one below it is doing more or less the same thing.

tthvo: Oh that was a mistake :D That should be replaced with the new one...

Makefile
```diff
@@ -164,6 +166,14 @@ generate: controller-gen
 	go generate ./...
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+# Check and add (if missing) license header
+LICENSE_FILE=$(shell pwd)/LICENSE
+.PHONY: add-license 
+add-license: addlicense
+	@echo "Checking/Adding license..."
+	@go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | sort | uniq | xargs $(ADDLICENSE) -v -f $(LICENSE_FILE) 2>&1
```
tthvo, line 174: Do we need to sort the directory list tho?

ebaron: I don't think so

ebaron, line 174: A couple of suggestions here:
1. Remove the std error redirection here, unless it's needed for something
2. It would be nice to see how `addlicense` is invoked, so we can continue to verify it's working correctly. Could you extract the `go list ... | uniq` pipeline to a variable, and pass that to `addlicense` and remove the echo suppression?

Makefile
```diff
@@ -164,6 +166,14 @@ generate: controller-gen
...
+	@echo "All good"
```
ebaron, line 175: I think this `echo` can be removed

Makefile
tthvo: Yess thanks a lot! Just added new changes :D

tthvo: Output (if one is missing):

```
echo "Checking/Adding license..."
Checking/Adding license...
/home/thvo/workspace/cryostat-operator/bin/addlicense -v -f /home/thvo/workspace/cryostat-operator/LICENSE api internal
2022/07/04 13:53:50 internal/main.go modified
```

Makefile
```diff
@@ -164,6 +166,14 @@ generate: controller-gen
 	go generate ./...
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+# Check and add (if missing) license header
+LICENSE_FILE=$(shell pwd)/LICENSE
+.PHONY: add-license 
+add-license: addlicense
+	echo "Checking/Adding license..."
+	$(eval GO_PACKAGES  := $(shell go list -f '{{.Dir}}' ./... | sed -e "s|^$$(pwd)||" | cut -d/ -f2 | uniq))
```
ebaron, line 174: I think it's fine to declare this outside of the recipe, and remove the `eval` part. This should make it a little simpler.

Makefile
```diff
@@ -164,6 +166,14 @@ generate: controller-gen
 	go generate ./...
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+# Check and add (if missing) license header
+LICENSE_FILE=$(shell pwd)/LICENSE
+.PHONY: add-license 
+add-license: addlicense
+	echo "Checking/Adding license..."
```
ebaron, line 173: This `echo` should still have the `@` to prevent it printing twice.

Makefile
andrewazores: There should still be a `sort` in the pipeline before `uniq`, because `uniq` only detects repeated lines if they are adjacent to each other.

Or you can use `sort -u` instead of `sort | uniq`

andrewazores: Unless `go list` guarantees the output ordering will already be sorted, though adding `sort` to the pipeline is really low impact in the redundant case anyway.

andrewazores: `go help list` also explains:

```
The -find flag causes list to identify the named packages but not
resolve their dependencies: the Imports and Deps lists will be empty.

The -test flag causes list to report not only the named packages
but also their test binaries (for packages with tests), to convey to
source code analysis tools exactly how test binaries are constructed.
The reported import path for a test binary is the import path of
the package followed by a ".test" suffix, as in "math/rand.test".
When building a test, it is sometimes necessary to rebuild certain
dependencies specially for that test (most commonly the tested
package itself). The reported import path of a package recompiled
for a particular test binary is followed by a space and the name of
the test binary in brackets, as in "math/rand [math/rand.test]"
or "regexp [sort.test]". The ForTest field is also set to the name
of the package being tested ("math/rand" or "sort" in the previous
examples).
```

Adding `-find` speeds up the invocation.

```bash
$ time go list -f '{{.Dir}}' ./...
/home/andrew/workspace/cryostat-operator/api/v1beta1
/home/andrew/workspace/cryostat-operator/internal
/home/andrew/workspace/cryostat-operator/internal/controllers
/home/andrew/workspace/cryostat-operator/internal/controllers/client
/home/andrew/workspace/cryostat-operator/internal/controllers/common
/home/andrew/workspace/cryostat-operator/internal/controllers/common/resource_definitions
/home/andrew/workspace/cryostat-operator/internal/test

real	0m0.242s
user	0m0.906s
sys	0m0.223s
```

```bash
$ time go list -find -f '{{.Dir}}' ./...
/home/andrew/workspace/cryostat-operator/api/v1beta1
/home/andrew/workspace/cryostat-operator/internal
/home/andrew/workspace/cryostat-operator/internal/controllers
/home/andrew/workspace/cryostat-operator/internal/controllers/client
/home/andrew/workspace/cryostat-operator/internal/controllers/common
/home/andrew/workspace/cryostat-operator/internal/controllers/common/resource_definitions
/home/andrew/workspace/cryostat-operator/internal/test

real	0m0.117s
user	0m0.344s
sys	0m0.130s
```

Using `-test` finds some more (duplicated) source directories:

```bash
$ time go list -test -f '{{.Dir}}' ./...
/home/andrew/workspace/cryostat-operator/api/v1beta1
/home/andrew/workspace/cryostat-operator/internal
/home/andrew/workspace/cryostat-operator/internal/controllers
/home/andrew/workspace/cryostat-operator/internal/controllers/client
/home/andrew/workspace/cryostat-operator/internal/controllers/common
/home/andrew/workspace/cryostat-operator/internal/controllers/common/resource_definitions
/home/andrew/workspace/cryostat-operator/internal/test
/home/andrew/workspace/cryostat-operator/internal/controllers
/home/andrew/workspace/cryostat-operator/internal/controllers

real	0m0.305s
user	0m1.004s
sys	0m0.256s
```

These flags can't be used together, though.

I think using `-test` with a `sort | uniq` in the pipeline is probably the most correct and complete listing.

Makefile
tthvo: Oh I might have misunderstood it!

Makefile
tthvo: Cool thanks @andrewazores Just added the changes!

andrewazores (approved):
LGTM

ebaron (approved):
Nice job @tthvo!

Issue #410, opened 2022-06-13
Existing Services should be updated


ebaron:
Services do not currently have any update logic, the operator should continually update them to reflect the definitions in the Cryostat CR.

Issue #411, opened 2022-06-13
Existing Ingresses should be updated


ebaron:
While Routes are currently updated by the operator when required, Ingresses are not. The operator should update these as well in response to config changes.

PR #424, opened 2022-07-11
feat(network): update operator-controlled networking objects

ebaron:
This PR accomplishes a few things:
1. Services, Ingresses and Routes will be updated across operator versions, and reflect changes the user makes to the Cryostat CR.
2. Uses the same code paths ([CreateOrUpdate](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate)'s mutateFn) when creating or updating the object. Ensures that when the operator sets a new field in an object, we don't need to ensure that change is made in two places.
3. Begins refactoring the long cryostat_controller.go and resource_definitions.go into resource-specific source files. I plan to repeat this pattern with the other resource types. The Cryostat controller's Reconcile method should be much smaller and simpler by fanning out to resource-specific delegates once this is complete.

Apologies for the PR being large. I initially planned to break this into two or three smaller PRs, but these types depend on each other.

Fixes: #410, #411

### DIFF
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,9 +6,9 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=cryostat-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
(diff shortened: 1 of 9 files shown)


andrewazores (approved):
Built a bundle and deployed this on `crc` and tested various configuration changes (reports, core service config, storage config). Everything still looks like it's working. Fanning out to different sub reconcile loops makes a lot of sense - the main reconcile loop has grown a lot over time.

PR #425, opened 2022-07-18
feat(labels): allow users to define labels/annotations for routes

### LINKED ISSUE #407: Add ability to customize route labels
Users should be able to define labels in the Cryostat CRD that are automatically applied to the Routes that the operator creates. We may be able to use the existing `networkOptions.*.labels` properties for this.

tthvo:
Fixes #407 

Users can now define labels for routes in Cryostat CR.

### DIFF
--- a/internal/controllers/cryostat_controller_test.go
+++ b/internal/controllers/cryostat_controller_test.go
@@ -1291,45 +1291,6 @@ var _ = Describe("CryostatController", func() {
 					t.checkRoute(test.NewCustomGrafanaRoute(t.TLS))
 				})
 			})
-			Context("and existing routes", func() {
-				var cr *operatorv1beta1.Cryostat
-				BeforeEach(func() {
-					t.objs = append(t.objs, test.NewCryostat())
-				})
-				JustBeforeEach(func() {
-					// Fetch the current Cryostat CR
-					namespacedName := types.NamespacedName{Name: cr.Name, Namespace: cr.Namespace}
-					current := &operatorv1beta1.Cryostat{}
-					err := t.Client.Get(context.Background(), namespacedName, current)
-					Expect(err).ToNot(HaveOccurred())
-
-					// Customize it with network options from the test specs
-					current.Spec = cr.Spec
-					err = t.Client.Update(context.Background(), current)
-					Expect(err).ToNot(HaveOccurred())
-
-					// Reconcile again
-					result, err := t.controller.Reconcile(context.Background(), reconcile.Request{NamespacedName: namespacedName})
-					Expect(err).ToNot(HaveOccurred())
-					Expect(result).To(Equal(reconcile.Result{}))
-				})
-				Context("containing core config", func() {
-					BeforeEach(func() {
-						cr = test.NewCryostatWithCoreNetworkOptions()
-					})
-					It("should create the route as describe", func() {
-						t.checkRoute(test.NewCustomCoreRoute(t.TLS))
-					})
-				})
-				Context("containing grafana config", func() {
-					BeforeEach(func() {
-						cr = test.NewCryostatWithGrafanaNetworkOptions()
-					})
-					It("should create the route as describe", func() {
-						t.checkRoute(test.NewCustomGrafanaRoute(t.TLS))
-					})
-				})
-			})
 		})
 	})
 	Describe("reconciling a request in Kubernetes", func() {
--- a/internal/controllers/cryostat_controller_test.go
+++ b/internal/controllers/cryostat_controller_test.go
@@ -344,18 +344,14 @@ var _ = Describe("CryostatController", func() {
 				oldGrafanaRoute = test.OtherGrafanaRoute()
 				t.objs = append(t.objs, cr, oldCoreRoute, oldGrafanaRoute)
 			})
-			It("should update the Routes", func() {
+			It("should update Routes", func() {
 				t.reconcileCryostatFully()
 
-				// Routes should be replaced except for labels/annotations
+				// Routes should be replaced
 				expectedCoreRoute := test.NewCoreRoute(t.TLS)
-				expectedCoreRoute.Annotations = oldCoreRoute.Annotations
-				expectedCoreRoute.Labels = oldCoreRoute.Labels
 				t.checkRoute(expectedCoreRoute)
 
 				expectedGrafanaRoute := test.NewGrafanaRoute(t.TLS)
-				expectedGrafanaRoute.Annotations = oldGrafanaRoute.Annotations
-				expectedGrafanaRoute.Labels = oldGrafanaRoute.Labels
 				t.checkRoute(expectedGrafanaRoute)
 			})
 		})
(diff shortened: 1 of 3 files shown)


ebaron (commented):
Hi @tthvo, looks good for the most part. You're going to want to set the route's labels inside of the createOrUpdate callback (located in createOrUpdateRoute). This ensures that the labels continue to be updated after the Route has been created. See how it's done for Ingress as an example:
https://github.com/cryostatio/cryostat-operator/blob/5212264d9c61007472ddb599d042b665f8b4a3b9/internal/controllers/ingresses.go#L122-L133

I have a couple of additional requests:
1. Could you also do the same for Annotations? I should have mentioned this in the original issue.
2. Could you add some test(s) to verify that labels and annotations defined in the Cryostat custom resource are applied to the Routes? You could add a new function to return a Cryostat CR with customized labels and routes in internal/test/resources.go, similar to this: https://github.com/cryostatio/cryostat-operator/blob/5212264d9c61007472ddb599d042b665f8b4a3b9/internal/test/resources.go#L258-L281
Then add a test case in cryostat_controller_test.go that sets your customized custom resource in `t.objs` and add expectations on the route that the controller creates. The `checkRoute` method might be useful here. Let me know if you need some help with this, the test framework can be a bit confusing.

tthvo:
Yeh suree I am looking into the test framework and will work on it now! Sorry for PR not being completed yet :D

tthvo:
Hi @ebaron Is this still valid? Since we allow routes to receive custom labels.
https://github.com/cryostatio/cryostat-operator/blob/5212264d9c61007472ddb599d042b665f8b4a3b9/internal/controllers/cryostat_controller_test.go#L350

ebaron:
> Hi @ebaron Is this still valid? Since we allow routes to receive custom labels.
> 
> https://github.com/cryostatio/cryostat-operator/blob/5212264d9c61007472ddb599d042b665f8b4a3b9/internal/controllers/cryostat_controller_test.go#L350

Ah good catch. That part of the test will have to be modified since the labels and annotations are now managed by the operator.

tthvo:
https://github.com/cryostatio/cryostat-operator/blob/5212264d9c61007472ddb599d042b665f8b4a3b9/internal/controllers/cryostat_controller_test.go#L345

@ebaron when appending multiple objects as such, how are the states of route objects after running `reconcileCryostatFully`?

ebaron:
> https://github.com/cryostatio/cryostat-operator/blob/5212264d9c61007472ddb599d042b665f8b4a3b9/internal/controllers/cryostat_controller_test.go#L345
> 
> @ebaron when appending multiple objects as such, how are the states of route objects after running `reconcileCryostatFully`?

Any properties of the routes that are managed by the operator should be overwritten. You can see this inside of the createOrUpdateRoute:
https://github.com/cryostatio/cryostat-operator/blob/5212264d9c61007472ddb599d042b665f8b4a3b9/internal/controllers/routes.go#L133-L144

`Spec.To` is modified to point to the relevant Service (if it's not already), likewise `Spec.Port` is updated to point to the Service's port number. `Spec.TLS` is updated to use re-encryption or edge [TLS termination](https://docs.openshift.com/container-platform/4.10/networking/routes/secured-routes.html) depending on whether cert-manager integration is enabled (see Cryostat CRD: `Spec.EnableCertManager`). The Route's owner is set to the Cryostat custom resource. Finally, with your PR, the labels and annotations will be updated as well.

tthvo:
Ahh okayy got it! It all makes sense now! Any thoughts on new changes @ebaron?

ebaron:
> The new Cryostat CR will overwrite those old routes [1] or the other way around [2]? From the way the test presents, case [1] seems to be correct?

The former is correct. If a route exists already, the operator will change it as required to work with Cryostat's services and to reflect any user-provided customizations present in the Cryostat CR.

For your new test case, I think it's fine to not include these `oldCoreRoute` and `oldGrafanaRoute`. The updating case is covered by the existing test. Your test can instead just focus on the case where there are no existing routes, and you want to ensure that the routes are created with the labels and annotations provided in the Cryostat CR.

tthvo:
Ahh understood! It makes sense now! Thanks @ebaron! Any other suggestions I should be working on for this PR?

internal/controllers/routes.go
```diff
@@ -75,12 +76,12 @@ func (r *CryostatReconciler) reconcileGrafanaRoute(ctx context.Context, svc *cor
 			Namespace: cr.Namespace,
 		},
 	}
-
+	grafanaConfig := configureGrafanaRoute(cr)
```
ebaron, line 79: I would move this to after the following if-statement. This avoids calling this function if it's a minimal deployment.

internal/controllers/cryostat_controller_test.go
```diff
@@ -347,16 +347,9 @@ var _ = Describe("CryostatController", func() {
...
-				expectedCoreRoute.Labels = oldCoreRoute.Labels
-				t.checkRoute(expectedCoreRoute)
-
-				expectedGrafanaRoute := test.NewGrafanaRoute(t.TLS)
-				expectedGrafanaRoute.Annotations = oldGrafanaRoute.Annotations
-				expectedGrafanaRoute.Labels = oldGrafanaRoute.Labels
-				t.checkRoute(expectedGrafanaRoute)
+				// Routes should be replaced
+				t.checkRoute(test.NewCoreRoute(t.TLS))
+				t.checkRoute(test.NewGrafanaRoute(t.TLS))
```
ebaron, line 352: We can use `t.checkRoutes` to shorten this.

ebaron (commented):
This looks very good overall! I just have a few minor suggestions.

internal/controllers/routes.go
tthvo: Oh yess true!

internal/controllers/cryostat_controller_test.go
```diff
@@ -1278,6 +1270,27 @@ var _ = Describe("CryostatController", func() {
...
 		})
+		Context("with network options", func() {
+			JustBeforeEach(func() {
+				t.reconcileCryostatFully()
+			})
+			Context("containing core config", func() {
+				BeforeEach(func() {
+					t.objs = append(t.objs, test.NewCryostatWithCoreNetworkOptions())
+				})
+				It("should create the route as describe", func() {
```
ebaron, line 1281:
```suggestion
				It("should create the route as described", func() {
```

internal/controllers/cryostat_controller_test.go
```diff
@@ -1278,6 +1270,27 @@ var _ = Describe("CryostatController", func() {
...
+					t.checkRoute(test.NewCustomCoreRoute(t.TLS))
+				})
+			})
+			Context("containing grafana config", func() {
+				BeforeEach(func() {
+					t.objs = append(t.objs, test.NewCryostatWithGrafanaNetworkOptions())
+				})
+				It("should create the route as describe", func() {
```
ebaron, line 1289:
```suggestion
				It("should create the route as described", func() {
```

ebaron (commented):
One more thing. I just noticed this.

internal/controllers/cryostat_controller_test.go
tthvo: Oh always have some typos :D that is funny haha Fixed it!

ebaron (approved):
Looks great, excellent work @tthvo!

Issue #428, opened 2022-07-19
NetworkConfiguration docs should be more generic


ebaron:
Since we now use the NetworkConfiguration type to configure Route labels and annotations, they should not explicitly reference Ingresses.

Issue #403, opened 2022-06-06
[Story] Support Services without Ingresses or Routes


ebaron:
Our Helm chart currently allows users to install Cryostat using only Services for networking, with some additional post-install steps required. It would be useful if the operator could also support these use cases, without the post-install steps.

We would want to support `ClusterIP`, `NodePort` and `LoadBalancer` services. At least for `ClusterIP`, we would need to add some kind of "external URL" property to the Cryostat CRD. This would allow the user to inform the operator of what `host:port` the services are port-forwarded to. This is needed so Cryostat can set up the notifications WebSocket connection and also to launch Grafana.

ebaron:
We will need to be careful about `CRYOSTAT_SSL_PROXIED` and distinguishing between HTTP/HTTPS. See: https://github.com/cryostatio/cryostat-helm/pull/36

PR #429, opened 2022-07-19
docs(network): NetworkConfiguration is not Ingress-specific

ebaron:
This revises the API documentation to not be specific to Ingress, but also apply to Routes.

Fixes: #428


api/v1beta1/cryostat_types.go
```diff
@@ -68,8 +68,8 @@ type CryostatSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	ServiceOptions *ServiceConfigList `json:"serviceOptions,omitempty"`
-	// Options to control how the operator exposes the application outside of the cluster
-	// using an Ingress or Route.
+	// Options to control how the operator exposes the application outside of the cluster,
```
tthvo, line 71: Looks good! Just a quick question.. When we say `such as`, does that mean there are others objects besides Ingress/Route?

ebaron: I phrased it like this because we eventually want to give users the choice to expose Cryostat directly using Services instead of Ingress or Routes: https://github.com/cryostatio/cryostat-operator/issues/403

tthvo: Ahh got it! That makes sense!

tthvo (approved):

api/v1beta1/cryostat_types.go
ebaron: It's technically possible with all kinds of Services, but LoadBalancer is a good option for that.

tthvo: Nicee :)) Make sense! Thanks @ebaron!

andrewazores (approved):

PR #430, opened 2022-07-22
build(operator-sdk): upgrade operator sdk version to 1.22.2

### LINKED ISSUE #287: Upgrade Operator SDK to at least 1.8.2

### LINKED ISSUE #431: go-get-tool should install to $1 instead of $PROJECT_DIR/bin
> go-get-tool will 'go get' any package $2 and install it to $1.

`go-get-tool` defined in Makefile said to install the path specified in `$1` but eventually installs to `$PROJECT_DIR/bin` anyaway.

https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/Makefile#L224

tthvo:
Fixes #287 
Fixes #431 
Related to #433 

Upgraded operator-sdk to `v1.22.2`. (Actual) Changes for each version:

`1.6.1`: 
- Update tls configuration for metrics endpoint (though still disabled).
- Add patches to delete cert mount volume (still commented out but added for completeness).
- Add `build-catalog` target to build an index image.

`1.7.1`: 
- Add pipefail and non-zero status checks for `SHELL` in Makefile.
- Move lease to its own proxy-rule rule.
- Bump controller-runtime to `v0.8.3` and k8s deps to `v0.20.0`.

`1.8.0`: No migration

`1.9.0`: No migration

`1.10.0`: No migration

`1.11.0`: 
- Add protocol for `containerPort` field in manifest (i.e. setting to TCP).

`1.12.0`: No migration

`1.13.0`: No migration

`1.14.0`: 
- Bump `controller-gen` to `0.7.0`.
- Bump controller-runtime to `v0.10.0` and k8s deps to `v0.22.1`.
- Fix unit tests for FakeClient release [`0.9.0`](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.9.0)
- Use the new envtest tool (controller-runtime to `v0.10.0` removes `setup-envtest.sh`)
- Fixed kustomize target in Makefile since `go install` does not work with Kustomize. Issue is filed [here](https://github.com/kubernetes-sigs/kustomize/issues/3618).
- Use comment markers to mark deprecated API instead of patches since we upgrades past `1.10.*`.

`1.15.0`: No migration

`1.16.0`:
- Add annotation `kubectl.kubernetes.io/default-container` to specify default manager container.
- Add PHONY for all targets
- Add ignore-not-found flag for `undeploy/uninstall` targets.

`1.17.0`:
- Bump controller-gen to `0.8.0`.
- Bump k8s deps to `0.23.0` and controller-runtime to `0.11.0`.
- Add resource request/limit to kube-rbac-proxy container.
- Reduce debug log level of kube-rbac-proxy container from 10 to 0.
- Fix new struct names after release refactoring (i.e. Handler -> ProbeHandler)
- Fix unit tests after bumping to controller-runtime `0.11.0`.

`1.18.0`:
- Remove `replace` directive for gogo/protobuf since client-go new pulls `1.3.2`.
- Add support for bundle image digest instead of tags.

`1.19.0`:
- Manager deployment entry in CSV now has label field added.

`1.20.0`: No migration

`1.21.0`:
- Fix suite test to use global config.
- Bump controller-runtime to `0.11.2` and k8s deps to `0.23.5`.
- Bump `kube-rbac-proxy` image to `0.11.0`.

`1.22.0`:
- Bump go version to `1.18`.
- Bump controller-gen to `0.9.0`.
- Bump controller-runtime to `0.12.1` and k8s deps to `0.24.0`.

`1.20.1`: No migration

`1.20.2`: No migration

Others:
- Use `go install` instead of `go get` (deprecated) to download and build binaries.
- Bump `ginkgo` to `v1.16.5` (having deprecated warnings - currently add env var to silence) and `gomega` to `v1.18.1`.
- Since `1.22.0`, go version requirement is bump to `1.18`.
- Fix `spec.selector` should be updated only on creation.
- Update ci.yaml to use go `1.18.*` and operator-sdk `1.22.2`.


tthvo:
hi @ebaron  #356 mentions deprecated apis can be specified as markers. Would I start on fixing these too now that we have migrated to at least `v.0.13.0` now?

`v0.14.0` breaks tests as they upgrades controller-runtime to 0.10.0 so I will look into that.

ebaron:
> hi @ebaron #356 mentions deprecated apis can be specified as markers. Would I start on fixing these too now that we have migrated to at least `v.0.13.0` now?

That's a good idea, but we're planning to remove those APIs before the next release. So not a big deal.

tthvo:
Hi @ebaron, in our unit tests, when Cryostat CR is being deleted, we call `reconcileDeletedCryostat`:

https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/internal/controllers/cryostat_controller_test.go#L1540

which, I think, adds a delete-timestamp and calls `reconcile` again to resolve finalizers. 

Then, in test spec, we call check if finalizers are absent.

https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/internal/controllers/cryostat_controller_test.go#L1815

But in this case, would cryostat CR already be deleted since its finalizers are resolved so the finalizer checks failed to get cryostat CR as here?

```bash
should remove the finalizer [It]
          /home/thvo/workspace/cryostat-operator/internal/controllers/cryostat_controller_test.go:969

          Unexpected error:
              <*errors.StatusError | 0xc000178aa0>: {
                  ErrStatus: {
                      TypeMeta: {Kind: "", APIVersion: ""},
                      ListMeta: {
                          SelfLink: "",
                          ResourceVersion: "",
                          Continue: "",
                          RemainingItemCount: nil,
                      },
                      Status: "Failure",
                      Message: "cryostats.operator.cryostat.io \"cryostat\" not found",
                      Reason: "NotFound",
                      Details: {
                          Name: "cryostat",
                          Group: "operator.cryostat.io",
                          Kind: "cryostats",
                          UID: "",
                          Causes: nil,
                          RetryAfterSeconds: 0,
                      },
                      Code: 404,
                  },
              }
              cryostats.operator.cryostat.io "cryostat" not found
          occurred
```

tthvo:
Release [`0.9.0`](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.9.0) for controller-runtime said:

> Fakeclient: Handle Finalizers (https://github.com/kubernetes-sigs/controller-runtime/issues/869)
moderate impact: Finalizers in the fakeclient now behave similiar to the kube api, i.E. an object with a finalizer will not actually be deleted upon Delete and Updating an object with a DeletionTimestamp to remove the finalizer will result in deletion

ebaron:
> Release [`0.9.0`](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.9.0) for controller-runtime said:
> 
> > Fakeclient: Handle Finalizers ([kubernetes-sigs/controller-runtime#869](https://github.com/kubernetes-sigs/controller-runtime/issues/869))
> > moderate impact: Finalizers in the fakeclient now behave similiar to the kube api, i.E. an object with a finalizer will not actually be deleted upon Delete and Updating an object with a DeletionTimestamp to remove the finalizer will result in deletion

Ah, that explains the behaviour you saw then. In our existing tests, we would use the DeletionTimestamp to simulate the Cryostat CR being deleted, but nothing actually followed through with the deletion. I suppose now the tests should be updated to ensure the object is deleted instead of testing whether the finalizer is absent.

tthvo:
hi @ebaron, It looks like we update `template.spec` but not `templace.metadata`. Not sure if this is intentional?

https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/internal/controllers/cryostat_controller.go#L706

Fail test:

```
• Failure [0.005 seconds]
CryostatController
/home/thvo/workspace/cryostat-operator/internal/controllers/cryostat_controller_test.go:84
  reconciling a request in OpenShift
  /home/thvo/workspace/cryostat-operator/internal/controllers/cryostat_controller_test.go:121
    with an existing main Deployment
    /home/thvo/workspace/cryostat-operator/internal/controllers/cryostat_controller_test.go:272
      should update the Deployment [It]
      /home/thvo/workspace/cryostat-operator/internal/controllers/cryostat_controller_test.go:280

      Expected
          <string>: 
      to equal
          <string>: cryostat

      /home/thvo/workspace/cryostat-operator/internal/controllers/cryostat_controller_test.go:1939
```

tthvo:
It seems this only happens after bumping to `controller-runtime 0.11.0`. In case we expect the `$.spec.template.metadata` to not be updated, I add this to Deployment.Spec.Template in `OtherDeployment` method.

https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/internal/test/resources.go#L1684
```
	Name:      "cryostat",
	Namespace: "default",
	Labels: map[string]string{
		"app":       "cryostat",
		"kind":      "cryostat",
		"component": "cryostat",
	},
}
```

But I wonder if we need to update their labels/annotations in `$.spec.template.metadata` in our reconciler similar to `$.metadata` as here? Any suggestions @ebaron?

https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/internal/controllers/cryostat_controller.go#L700

tthvo:
With new upgrades, a `cache` directory is created when doing bundle deployment, not sure if we need to commit that so I add it into `.gitignore`.

andrewazores:
I did `make generate manifests manager oci-build bundle` and there are some changes to generated files:

```bash
$ git status
HEAD detached at thvo/upgrade-operator-sdk
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   bundle/manifests/operator.cryostat.io_cryostats.yaml
	modified:   bundle/manifests/operator.cryostat.io_flightrecorders.yaml
	modified:   bundle/manifests/operator.cryostat.io_recordings.yaml
	modified:   config/crd/bases/operator.cryostat.io_cryostats.yaml
	modified:   config/crd/bases/operator.cryostat.io_flightrecorders.yaml
	modified:   config/crd/bases/operator.cryostat.io_recordings.yaml
	modified:   config/rbac/role.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

`git diff` looks fine, so I think these are probably changes that should just be checked in and committed.

tthvo:
> 

Oh hmm that's weird. There appears no changes on my end. I am using operator-sdk `1.22.2` and go `1.18`. Any idea what happens?

andrewazores:
```bash
$ go version
go version go1.18.4 linux/amd64
$ operator-sdk version
operator-sdk version: "v1.22.2", commit: "da3346113a8a75e11225f586482934000504a60f", kubernetes version: "1.24.1", go version: "go1.18.4", GOOS: "linux", GOARCH: "amd64"
```

I'm in the middle of building and testing this PR as a bundle so I have some other changes that got made locally, modifying image tags and such. I'll post the diff of what I found earlier when this is out of the way.

tthvo:
Ahh would it be `controller-gen` out of date because I added a check in Makefile if `controller-gen` (same for other binaries) exists then no downloading executed. Right now, `controller-gen` is at `0.9.0`.

andrewazores:
Good hunch, but

```bash
$ ./bin/controller-gen --version
Version: v0.9.0
$ ./bin/kustomize version
{Version:kustomize/v3.8.7 GitCommit:ad092cc7a91c07fdf63a2e4b7f13fa588a39af4f BuildDate:2020-11-11T23:14:14Z GoOs:linux GoArch:amd64}
```

These were freshly retrieved since this is my first time doing an operator build on my new laptop.

tthvo:
Oh myy actually it is my setup that is out of date! Still on `0.7.0` when testing lower version! I will update and check in these changes now!

andrewazores:
Cool, just manually testing out deploying things using this new build looks good. I didn't go too in-depth but I was able to install the operator via bundle, create a Cryostat CR with a configuration for a report sidecar, and that got deployed as usual. Then I edited the CR to remove the report sidecar and the deployment was adjusted accordingly.

andrewazores:
> Oh myy actually it is my setup that is out of date! Still on `0.7.0` when testing lower version! I will update and check in these changes now!

Makes sense:

```diff
diff --git a/bundle/manifests/operator.cryostat.io_cryostats.yaml b/bundle/manifests/operator.cryostat.io_cryostats.yaml
index d2fa0f6..c55b204 100644
--- a/bundle/manifests/operator.cryostat.io_cryostats.yaml
+++ b/bundle/manifests/operator.cryostat.io_cryostats.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.9.0
```

andrewazores:
Nice, after your latest commit all the generated files appear to be up-to-date for me too.

ebaron:
> It seems this only happens after bumping to `controller-runtime 0.11.0`. In case we expect the `$.spec.template.metadata` to not be updated, I add this to Deployment.Spec.Template in `OtherDeployment` method since I guess metadata was missing and not updated in reconciler for new deployment so the check failed.
> 
> https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/internal/test/resources.go#L1684
> 
> ```
> ObjectMeta: metav1.ObjectMeta{
> 	Name:      "cryostat",
> 	Namespace: "default",
> 	Labels: map[string]string{
> 		"app":       "cryostat",
> 		"kind":      "cryostat",
> 		"component": "cryostat",
> 	},
> }
> ```
> 
> But I wonder if we need to update their labels/annotations in `$.spec.template.metadata` in our reconciler similar to `$.metadata` as here? Any suggestions @ebaron?
> 
> https://github.com/cryostatio/cryostat-operator/blob/5ceb7db194b765c07bf1e9270ea248da16cae581/internal/controllers/cryostat_controller.go#L700

I think this makes sense to do. In the future I would like the Deployment and Pod Template's labels and annotations to managed from the Cryostat CRD, but this seems like a good solution for now.

I'm still a bit confused what has changed in controller-runtime to cause this test failure.

tthvo:
Sure I will fix add the label and annotation merge for `spec.template.metatadata` now.

> I'm still a bit confused what has changed in controller-runtime to cause this test failure.

The [changelog](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.11.0) could not really help.

ebaron (commented):
> `1.6.1`:
> 
>     * Update tls configuration for metrics endpoint (though still disabled).

I think we might want this (commented out) for completeness: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.6.1/#manifestsv2-add-a-kustomize-patch-to-remove-the-cert-manager-volumevolumemount-from-your-csv

tthvo:
> 

Oh yes sure I might have been confused with our manually adding cert-manager. Will update now!

ebaron (commented):
This Makefile addition would be useful as well: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.6.1/#gov2-gov3-ansiblev1-helmv1-add-opm-and-catalog-build-makefile-targets

We can skip the `catalog-push` target since we don't have any other `*-push` targets in our Makefile. We'd also want to change `--container-tool docker` to `--container-tool podman`. We also don't seem to have an `IMAGE_TAG_BASE` in our Makefile, but you can substitute it with `$(IMAGE_NAMESPACE)/$(OPERATOR_NAME)`.

ebaron:
> This Makefile addition would be useful as well: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.6.1/#gov2-gov3-ansiblev1-helmv1-add-opm-and-catalog-build-makefile-targets
> 
> We can skip the `catalog-push` target since we don't have any other `*-push` targets in our Makefile. We'd also want to change `--container-tool docker` to `--container-tool podman`. We also don't seem to have an `IMAGE_TAG_BASE` in our Makefile, but you can substitute it with `$(IMAGE_NAMESPACE)/$(OPERATOR_NAME)`.

Ah, `IMAGE_TAG_BASE` is introduced in the next point: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.6.1/#gov2-gov3-ansiblev1-helmv1-changed-bundle_img-and-added-image_tag_base-makefile-variables

In our case, feel free to define: `IMAGE_TAG_BASE ?= $(IMAGE_NAMESPACE)/$(OPERATOR_NAME)` and replace instances of `$(IMAGE_NAMESPACE)/$(OPERATOR_NAME)` with `${IMAGE_TAG_BASE)`.

Makefile
```diff
@@ -1,4 +1,5 @@
-SHELL := /bin/bash
+SHELL := /bin/bash -o pipefail
+.SHELLFLAGS = -ec
```
ebaron, line 2: Might as well include the comment from here: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.7.1/#gov3-set-the-makefiles-shell-variable-to-the-systems-bash-binary

It will help us to remember what this is for.

ebaron:
> This Makefile addition would be useful as well: https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.6.1/#gov2-gov3-ansiblev1-helmv1-add-opm-and-catalog-build-makefile-targets
> 
> We can skip the `catalog-push` target since we don't have any other `*-push` targets in our Makefile. We'd also want to change `--container-tool docker` to `--container-tool podman`.

My mistake, we have an `IMAGE_BUILDER` variable, so this should be `--container-tool $(IMAGE_BUILDER)`.

tthvo:
Great I am going to wrap up these changes today or tomorrow :D

tthvo:
I also added a FIXME for `spec.selector` to be immutable following reference from the one the `controller-runtime` release: https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/example_test.go

```golang
// Immutable, only updated when the deployment is created
if deploy.ObjectMeta.CreationTimestamp.IsZero() {
	deploy.Spec.Selector = specCopy.Selector
}
```

tthvo:
> Sure I will fix add the label and annotation merge for `spec.template.metatadata` now.
> 
> > I'm still a bit confused what has changed in controller-runtime to cause this test failure.
> 
> The [changelog](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.11.0) could not really help :((

I tried looking into their source code after and before `0.11.0` release. It looks like they add a `zero` method that helps remove fields that should be removed during update patch. 

Here is the original issue: https://github.com/kubernetes-sigs/controller-runtime/issues/1639
Here is their fix: https://github.com/kubernetes-sigs/controller-runtime/pull/1651.

I guess it makes sense that `OtherDeployment` func does not define a `spec.template.metadata` but somehow receives them even though our reconciler does not update  `spec.template.metadata`. Looks reasonable to me but could be wrong haha :D

Makefile
```diff
@@ -197,41 +226,63 @@ check_cert_manager:
...
-rm -rf $$TMP_DIR ;\
-}
-endef
+ADDLICENSE = $(LOCALBIN)/addlicense
+.PHONY: addlicense
+addlicense: local-bin
+	[ -f $(ADDLICENSE) ] || GOBIN=$(LOCALBIN) go install github.com/google/addlicense@v$(ADDLICENSE_VERSION)
+
+# Download setup-envtest locally if necessary
+SETUP_ENVTEST = $(TESTBIN)/setup-envtest
```
ebaron, line 261: I noticed that the documentation uses `ENVTEST` for this variable name. Should we use that to minimize differences?

tthvo:
Since we upgraded our operator-sdk, not sure if we need to add `spec.minKubeVersion: some-version` as here?

https://olm.operatorframework.io/docs/best-practices/common/#provide-what-are-the-k8s-versions-supported-by-your-project

Makefile
```diff
@@ -197,41 +226,63 @@ check_cert_manager:
...
+	[ -f $(CONTROLLER_GEN) ] || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@v$(CONTROLLER_GEN_VERSION)
 
 # Download kustomize locally if necessary
-KUSTOMIZE = $(shell pwd)/bin/kustomize
-kustomize:
-	$(call go-get-tool,$(KUSTOMIZE),sigs.k8s.io/kustomize/kustomize/v3@v$(KUSTOMIZE_VERSION))
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
+KUSTOMIZE = $(LOCALBIN)/kustomize
+.PHONY: kustomize
+kustomize: local-bin
```
ebaron, line 251: I noticed in this guide, they use the path as the target (e.g. `$(KUSTOMIZE)` instead of `kustomize`): https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.21.0/#for-golang-based-operators--change-makefile-targets-to-replace-go-get-with-go-install

Is that something we can do, or did you run into a problem with it?

ebaron (commented):
Hey @tthvo, I just finished working through the migration guide myself. Very well done! I have just a couple of comments about the Makefile below.

Makefile
tthvo: Oh yes suree I will change this! Originally I used `SETUP_ENVTEST` to match its binary name :D

Makefile
tthvo: Oh no there is no issue at all. I was following our style for targets that fetch tools. Let me follow their current way then!

tthvo:
> Since we upgraded our operator-sdk, not sure if we need to add `spec.minKubeVersion: some-version` to our CSV as here?
> 
> https://olm.operatorframework.io/docs/best-practices/common/#provide-what-are-the-k8s-versions-supported-by-your-project

Any thoughts on this @ebaron?

ebaron:
> > Since we upgraded our operator-sdk, not sure if we need to add `spec.minKubeVersion: some-version` to our CSV as here?
> > https://olm.operatorframework.io/docs/best-practices/common/#provide-what-are-the-k8s-versions-supported-by-your-project
> 
> Any thoughts on this @ebaron?

This is something I've been meaning to add for a while. I noticed that now we get a bundle validation warning without it:
```
$ operator-sdk-v1.22.2_linux_amd64 bundle validate ./bundle --select-optional suite=operatorframework
WARN[0000] Warning: Value : (cryostat-operator.v2.2.0-dev) csv.Spec.minKubeVersion is not informed. It is recommended you provide this information. Otherwise, it would mean that your operator project can be distributed and installed in any cluster version available, which is not necessarily the case for all projects. 
INFO[0000] All validation tests have completed successfully 
```

Because of this warning, I think it's okay to add this with this PR. In our case, it should be set to `1.19.0`.

tthvo:
Yes sure! I will add those as a patch. Any guidance on how to determine that `0.19.0` is our choice?

ebaron:
> Yes sure! I will add those as a patch. Any guidance on how to determine that `0.19.0` is our choice?

We use some 1.19 features, such as the Ingress v1 API. Our [website](https://cryostat.io/) mentions support for Kubernetes 1.19 and OpenShift 4.6.

ebaron (commented):
I think you should just be able to set `minKubeVersion` in `config/manifests/bases/cryostat-operator.clusterserviceversion.yaml`. A patch should not be necessary.

tthvo:
Ahh makes sense :D Thanks a lottt!

ebaron (commented):
Oh, could you also update our CI config to use Go 1.18.* here:
https://github.com/cryostatio/cryostat-operator/blob/4e6f0c7516bd046169550346ebd088fa23a68ce3/.github/workflows/ci.yaml#L27

Makefile
```diff
@@ -197,41 +225,68 @@ check_cert_manager:
...
+addlicense: $(ADDLICENSE)
+$(ADDLICENSE): local-bin
+	[ -f $(ADDLICENSE) ] || GOBIN=$(LOCALBIN) go install github.com/google/addlicense@v$(ADDLICENSE_VERSION)
+
+# Download setup-envtest locally if necessary
+ENVTEST = $(TESTBIN)/setup-envtest
+.PHONY: setup-envtest
+setup-envtest: $(ENVTEST)
+$(ENVTEST): test-bin
+	[ -f $(ENVTEST) ] || GOBIN=$(TESTBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
ebaron, line 267: With the target changed to e.g. `$(KUSTOMIZE)` the `-f` check should no longer be necessary. This target will only run if the file does not exist. We should also take the same approach for the `local-bin` target. Replace the `local-bin` target with a `$(LOCALBIN)` target. We could get rid of `test-bin` as well and just use `$(LOCALBIN)` for everything.

This section should be very similar to the boilerplate here (with the addition of the `addlicense` tool):
https://github.com/operator-framework/operator-sdk/blob/da3346113a8a75e11225f586482934000504a60f/testdata/go/v3/memcached-operator/Makefile#L152-L180

Sorry for fussing over the Makefile syntax, but the closer we stay to the boilerplate in the Makefile, the easier future Operator SDK migrations will be.

tthvo: Oh yess suree! Let me change that! I didn't know using `$(KUSTOMIZE)` will also do the check! Good to learn that!

Makefile
```diff
@@ -1,4 +1,11 @@
-SHELL := /bin/bash
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL := /bin/bash -o pipefail
```
tthvo, line 3: Should we still enforce `/bin/bash` or use `/usr/bin/env bash` as in operator-sdk Makefile?

tthvo: https://github.com/operator-framework/operator-sdk/blob/0a7bfd8e41b8ac75e18d7138690ac9bddd6ea49a/testdata/go/v3/memcached-operator/Makefile#L64

andrewazores: `/usr/bin/env bash` is more portable, basically. It's usually fair to assume that `env` exists at `/usr/bin`, but in some distros or operating systems, `bash` may not be at `/usr/bin/bash`, perhaps at `/usr/local/bin/bash` and without any symlink for example. Same reason you'll often see `.py` files starting with ex. `#!/usr/bin/env python2`.

tthvo: Make sense! I will set it to use `env` for finding bash.

config/manifests/remove_cert_manager_volume_patch.yaml
```diff
@@ -0,0 +1,8 @@
+# Remove the manager container's "cert" volumeMount, since OLM will create and mount a set of certs.
+# Update the indices in this path if adding or removing containers/volumeMounts in the manager's Deployment.
+- op: remove
+  path: /spec/template/spec/containers/1/volumeMounts/0
+# Remove the "cert" volume, since OLM will create and mount a set of certs.
+# Update the indices in this path if adding or removing volumes in the manager's Deployment.
+- op: remove
+  path: /spec/template/spec/volumes/0
```
ebaron, line 8: In general, I think the approach you've taken here to include this patch in a separate file is cleaner. However, in the interest of making it easy to follow future migration steps, could you instead add this inline to `config/manifests/kustomization.yaml` like the example? https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.6.1/#manifestsv2-add-a-kustomize-patch-to-remove-the-cert-manager-volumevolumemount-from-your-csv

go.mod
```diff
@@ -1,27 +1,92 @@
...
+	k8s.io/kube-openapi v0.0.0-20220328201542-3ee0da9b0b42 // indirect
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
+	sigs.k8s.io/json v0.0.0-20211208200746-9f7c6b3444d2 // indirect
+	sigs.k8s.io/structured-merge-diff/v4 v4.2.1 // indirect
+	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
 replace github.com/openshift/api => github.com/openshift/api v0.0.0-20200618202633-7192180f496a
 
-// Fix for CVE-2020-26160, revisit when upgrading client-go
```
ebaron, line 19: Could we leave this comment in here for now? We might be able to remove this when upgrading other dependencies, but it's useful context until then.

internal/controllers/cryostat_controller.go
```diff
@@ -710,14 +710,18 @@ func (r *CryostatReconciler) createOrUpdateDeployment(ctx context.Context, deplo
 		if err := controllerutil.SetControllerReference(owner, deploy, r.Scheme); err != nil {
 			return err
 		}
-		// Update pod template spec and selector to propagate any changes from Cryostat CR
-		deploy.Spec.Template.Spec = specCopy.Template.Spec
-		// FIXME Immutable
-		deploy.Spec.Selector = specCopy.Selector
+		// Immutable, only updated when the deployment is created
+		if deploy.ObjectMeta.CreationTimestamp.IsZero() {
```
ebaron, line 714: I think you can do `deploy.CreationTimestamp.IsZero()` to make this shorter. I've also filed #433, which relates to this problem. This is part of the fix, but we should also detect when the selector needs to be updated and delete/recreate the deployment as a workaround.

Makefile
```diff
@@ -197,41 +225,61 @@ check_cert_manager:
...
+setup-envtest: $(ENVTEST)
+$(ENVTEST): $(LOCALBIN)
+	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+
+# Download opm locally if necessary
+OPM = $(LOCALBIN)/opm
+.PHONY: opm
+opm: $(OPM)
+$(OPM): $(LOCALBIN)
+	[ -f $(OPM) ] || { \
```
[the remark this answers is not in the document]
tthvo: Oh yess! Sorry I forgot to remove it last time!

go.mod
tthvo: Since it is fixed as `client-go` is upgraded, should I reword this comments tho?

ebaron: Sure, you can remove the `revisit ...` part. I think this can go away when we upgrade the OpenShift API, but need to make sure nothing else tries to pull in this old version.

tthvo: Right! Makes sense!

tthvo: Ahh I was actually looking at the wrong CVE number. Sorry for the mix-up :D

ebaron: No problem, looks good now. Just needs another `go mod tidy`

ebaron:
@tthvo I have to apologize. It seems the Makefile tool download targets are not working as I expected. Since they all have a prerequisite of `$(LOCALBIN)`, the modification time of `$(LOCALBIN)` is changed whenever a new tool is downloaded. This means that unless the same tool is being downloaded twice in succession, `$(LOCALBIN)` will always appear newer than the already downloaded tool, and Make will download it again. This can cause a problem with the Kustomize tool in particular, which is an open bug in the Operator SDK: https://github.com/operator-framework/operator-sdk/issues/5875

The fix coming upstream from Kubebuilder is very similar to what you had before, only using `test -s` instead: https://github.com/kubernetes-sigs/kubebuilder/commit/a9bf7b940fba55e47ed523c96e79d6e3f8f3d66e.

Could you add `test -s ...` to the Make recipes for controller-gen, kustomize, addlicense, envtest, and opm, as is done in the above Kubebuilder commit? Sorry again for getting this wrong.

ebaron:
> The fix coming upstream from Kubebuilder is very similar to what you had before, only using `test -s` instead: [kubernetes-sigs/kubebuilder@a9bf7b9](https://github.com/kubernetes-sigs/kubebuilder/commit/a9bf7b940fba55e47ed523c96e79d6e3f8f3d66e).
> 
> Could you add `test -s ...` to the Make recipes for controller-gen, kustomize, addlicense, envtest, and opm, as is done in the above Kubebuilder commit? Sorry again for getting this wrong.

Actually this Kubebuilder fix looks odd. It seems like instead of using `test -s $(LOCALBIN)/controller-gen` they should use `test -s $(CONTROLLER_GEN)`. This is because `CONTROLLER_GEN` and related variables can be overriden by the user.

tthvo:
> @tthvo I have to apologize. It seems the Makefile tool download targets are not working as I expected. Since they all have a prerequisite of `$(LOCALBIN)`, the modification time of `$(LOCALBIN)` is changed whenever a new tool is downloaded. This means that unless the same tool is being downloaded twice in succession, `$(LOCALBIN)` will always appear newer than the already downloaded tool, and Make will download it again. This can cause a problem with the Kustomize tool in particular, which is an open bug in the Operator SDK: [operator-framework/operator-sdk#5875](https://github.com/operator-framework/operator-sdk/issues/5875)
> 
> The fix coming upstream from Kubebuilder is very similar to what you had before, only using `test -s` instead: [kubernetes-sigs/kubebuilder@a9bf7b9](https://github.com/kubernetes-sigs/kubebuilder/commit/a9bf7b940fba55e47ed523c96e79d6e3f8f3d66e).
> 
> Could you add `test -s ...` to the Make recipes for controller-gen, kustomize, addlicense, envtest, and opm, as is done in the above Kubebuilder commit? Sorry again for getting this wrong.

Ohh no problem at all! I just saw the exact same behavior when running it too and about to ask!  I would just add it back with `test` then.

ebaron:
> Right! I guess we can just use `CONTROLLER_GEN` then.

I think so, unless I'm missing something. Same for the other binaries.

tthvo:
I am thinking whether only `LOCALBIN` variable could be overrided by user instead of the var for the binaries themselves since we set `GOBIN` anyway?

ebaron (approved):
Looks good! @tthvo thanks for taking on this substantial and overdue change!

tthvo:
Thanks a lot @ebaron for helping me in this! Its great to learn a lot more from this issue!

ebaron:
> Thanks a lot @ebaron for helping me in this! Its great to learn a lot more from this issue!

I was glad to help!

PR #432, opened 2022-07-22
feat(crd): add reference to Grafana secret in Cryostat CR

### LINKED ISSUE #391: Add Grafana Secret to CRD Status
I think it would be useful to add a reference to the Grafana Secret in the Cryostat CR. This would make it easier for a user to locate the the generated Grafana credentials.

maxcao13:
Fixes #391 

Thoughts? 

Now, to get the Grafana secrets, users can run
```
$ kubectl get cryostat -o jsonpath='{$.items[0].status.grafanaSecret.data.GF_SECURITY_ADMIN_USER}' | base64 -d
\\ admin
$ kubectl get cryostat -o jsonpath='{$.items[0].status.grafanaSecret.data.GF_SECURITY_ADMIN_PASSWORD}' | base64 -d
'\\ someRandomGeneratedPassword
```

1. I'm not sure the best way to test the password since it's generated randomly so I just tested if it was a string and non-empty. There must be better ways right?
2. Also, would it be better to _also_ have the secrets be revealed in ASCII form when doing 
`kubectl get cryostat` ?

Note: Thanks to @tthvo a lot for helping me understand a lot of the new concepts and helping me with this PR!


ebaron:
Hi @maxcao13, thanks for working on this! Sorry I wasn't clear in the issue, but I intended to just store the secret name in the Status and not the entire secret. Secrets have some security advantages over other Kubernetes objects [1], so storying the contents of the secret in our custom resource would lose those advantages. Your changes would look somewhat similar though.



[1] https://kubernetes.io/docs/concepts/configuration/secret/#information-security-for-secrets

> I'm not sure the best way to test the password since it's generated randomly so I just tested if it was a string and non-empty. There must be better ways right?

You won't need to worry about this with just the name being stored in the CR now.

api/v1beta1/cryostat_types.go
```diff
@@ -118,7 +118,7 @@ type CryostatStatus struct {
 	ApplicationURL string `json:"applicationUrl"`
 	// Grafana secret
 	// +operator-sdk:csv:customresourcedefinitions:type=status
-	GrafanaSecret *corev1.Secret `json:"grafanaSecret"`
+	GrafanaSecret string `json:"grafanaSecret"`
```
ebaron, line 121: Could you add the `// +optional` marker for this field, and also append `omitempty` to the JSON tag? Optional tells the Kubernetes API server that this field is not required, such as before the secret is created. `omitempty` is similar, but functions on the JSON level. If the field is missing, it won't be transmitted in the JSON body.

There's one more marker I'd like to add here. If you look at other fields, you'll see an `xDescriptors` marker. This controls some generated UI in the OpenShift Console. We can tell the Console that this field refers to the name of a Kubernetes Secret. When it gets rendered in the OpenShift Console, it should create a link to the named Secret that users can click on. The example here conveniently shows how this works for Secrets, but also works for other kinds of Kubernetes objects: https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md#3-k8sResourcePrefix

api/v1beta1/cryostat_types.go
```diff
@@ -116,6 +116,9 @@ type CryostatStatus struct {
 	// Address of the deployed Cryostat web application
 	// +operator-sdk:csv:customresourcedefinitions:type=status,xDescriptors={"urn:alm:descriptor:org.w3:link"}
 	ApplicationURL string `json:"applicationUrl"`
+	// Grafana secret
```
ebaron, line 119: This comment turns into the field's description, which is shown to the user. We might want to elaborate a bit on what this contains. Maybe something like: "Secret containing the generated Grafana credentials"

internal/controllers/cryostat_controller.go
```diff
@@ -316,6 +316,12 @@ func (r *CryostatReconciler) Reconcile(ctx context.Context, request ctrl.Request
 		}
 	}
 
+	instance.Status.GrafanaSecret = instance.Name + "-grafana-basic"
```
ebaron, line 319: Perhaps for this, we could use the value of `grafanaSecret.Name`? It'll be less error-prone if we decide to change the secret's name in the future.

internal/controllers/cryostat_controller.go
```diff
@@ -316,6 +316,12 @@ func (r *CryostatReconciler) Reconcile(ctx context.Context, request ctrl.Request
...
+	err = r.Client.Status().Update(context.Background(), instance)
```
ebaron, line 320: This should use `ctx` instead of `context.Background()`. We used `context.Background()` a lot before the Reconcile method provided a context to use, but I've been trying to replace this whenever making other changes to the affected code.

ebaron (commented):
Hey @maxcao13, nice work! I just have a few minor suggestions below.

ebaron (commented):
The code changes look good, but I'm having trouble pushing to Quay.io at the moment so I can't test it. Hopefully I'll have better luck tomorrow

ebaron (approved):
Hi Max, got Quay working. Your PR works as expected. Thanks for taking this on!

If you're curious, this is how the Grafana Secret looks in the OpenShift Console:
1. The Cryostat CR page has a clickable link to the Grafana Secret.
Screenshot 2022-07-27 at 14-49-46 Red Hat OpenShift Container Platform

2. Clicking on the Secret takes you to the Secret in the console, and you can scroll down to reveal the Grafana credentials.
Screenshot 2022-07-27 at 14-50-17 cryostat-sample-grafana-basic · Details · Red Hat OpenShift Container Platform
Screenshot 2022-07-27 at 14-55-49 cryostat-sample-grafana-basic · Details · Red Hat OpenShift Container Platform

maxcao13:
Thanks for helping Elliott!

For future reference, how are you able to install my build onto the OperatorHub as an "Installed Operator"? I have been testing using crc and have only been able to `make deploy` so I can only see the operator as part of the Deployments and Pods.

andrewazores:
> For future reference, how are you able to install my build onto the OperatorHub as an "Installed Operator"? I have been testing using crc and have only been able to `make deploy` so I can only see the operator as part of the Deployments and Pods.

`make bundle` and `make deploy_bundle`, essentially.

ebaron:
Right, you'll need to make sure that the bundle points to your custom operator image. Something like this:
```
export OPERATOR_IMG=quay.io/me/cryostat-operator:latest
export BUNDLE_IMG=quay.io/me/cryostat-operator-bundle:latest
make oci-build
podman push $OPERATOR_IMG
make bundle 
make bundle-build
podman push $BUNDLE_IMG
make deploy_bundle
```

Issue #434, opened 2022-08-05
Update dependencies that rely on Kubernetes


ebaron:
After #287 updated the version of Kubernetes API/client-go used by the operator, we can also upgrade other dependencies that use Kubernetes to a similar version. These dependencies include the OpenShift API and cert-manager.

PR #435, opened 2022-08-05
build(deps): update OpenShift and cert-manager

ebaron:
This PR updates the OpenShift API and cert-manager API to versions that target Kubernetes 1.24. The cert-manager import path changed beginning with 1.8.

Fixes: #434


tthvo (commented):
Looks great! I think we could also change this link.

https://github.com/cryostatio/cryostat-operator/blob/358805837b1df695756d3d543bb351c0c204ef25/README.md?plain=1#L56

Also, I wonder we need to update cert-manager version here too?
https://github.com/cryostatio/cryostat-operator/blob/358805837b1df695756d3d543bb351c0c204ef25/Makefile#L70

go.mod
```diff
@@ -3,22 +3,22 @@ module github.com/cryostatio/cryostat-operator
...
+	github.com/cert-manager/cert-manager v1.9.1
+	github.com/go-logr/logr v1.2.2
 	github.com/onsi/ginkgo v1.16.5
-	github.com/onsi/gomega v1.18.1
-	github.com/openshift/api v3.9.0+incompatible
-	k8s.io/api v0.24.0
-	k8s.io/apimachinery v0.24.0
-	k8s.io/client-go v0.24.0
+	github.com/onsi/gomega v1.19.0
+	github.com/openshift/api v0.0.0-20220712151050-2647eb31dee7
```
tthvo, line 10: Also, I wonder which source we can find this tag?

ebaron: This is from the `release-4.11` branch. I added a comment to indicate that. Go modules don't work with branches, only tags, and the OpenShift API repo doesn't tag their releases.

ebaron:
> Looks great! I think we could also change this link.
> 
> https://github.com/cryostatio/cryostat-operator/blob/358805837b1df695756d3d543bb351c0c204ef25/README.md?plain=1#L56
> 
> Also, I wonder we need to update cert-manager version here too?
> 
> https://github.com/cryostatio/cryostat-operator/blob/358805837b1df695756d3d543bb351c0c204ef25/Makefile#L70

I've updated these. I also used cert-manager 1.7.1 instead of the newest, since this is what is currently available with the Red Hat cert-manager operator: https://catalog.redhat.com/software/containers/cert-manager/cert-manager-operator-rhel8/61a60be7bfd4a5234d596292

go.mod
tthvo: Ahh that makes sense! Thanks a lot!

tthvo (approved):
Looks great!!

PR #437, opened 2022-08-10
fix(resources): set cryostat container resource request defaults

### LINKED ISSUE #406: Add default resource requests for Cryostat containers
While we allow the user to specify resource requirements for each Cryostat container, we currently default to no requirements. To follow operator best practices, we should at least provide default CPU/memory requests for each container [1].

[1] https://sdk.operatorframework.io/docs/best-practices/managing-resources/

tthvo:
Fixes #406 

Cryostat containers now have defaults for their resource requests.


tthvo:
Holding off to check if `jfr-datasource` could be updated to support filters.

tthvo:
Minimum resources needed for start up:

| Container | Cpu | Memory|
|---|---|---|
|Datasource| 2Mi | 16.25MiB |
|Report| 16Mi | 134.6MiB |

andrewazores:
Quite the difference, that's the native image magic at work.

tthvo:
Right! its much lower than when I checked with Datasource in JVM :D

I am thinking of these requests at least from current observations. 

**Datasource**

| Cpu | Memory|
|---|---|
|100Mi | 512 MiB |

It seems the datasource while in action uses lots of memory resources. A single 10MiB recording already requires 315MiB memory. So, I chose a fairly big number but not over 1GB -> half of it. A quick notice is that the datasource appears to hold on to memory resource between reloads (memory usage keeps increasing). 

Cpu usage makes sense and tends to linearly grow with file size as what it mostly does loading the recording content in memory and handle http requests. This could even do less work than Cryostat container. So, to best safeguard, I chose it to be the same as Cryostat -> 100Mi

**Reports**

| Cpu | Memory|
|---|---|
| 128 Mi | 256 MiB |

Reports resources usage appear linearly grows with file size. in comparison to datasource, its usage of cpu is higher for the same recordings but not much farther. With this and the fact that report is built in JVM and would have agent attached later, I chose it to be higher than datasource. Perhaps, 128Mi. For memory, it is pretty halved that of datasource for each recordings -> 256MiB.

These number will need refining overtime. For now, what do you think?

ebaron:
I'm fine with this. Our choices will be somewhat arbitrary no matter what, and you've done some great research. Users have the capabilities of adjusting the resource requirements quite easily with our CRD.

tthvo:
I think its ready for review now :D Just a bit complications with resource limit/request map since a nil map is treated as no-specification while empty map will give "0" as requested/limited.

For tests, I have to do manual testing for resource value to avoid this issue below (perhaps deployment is modified for cleaner view) and some hacks to check resources spec cases (i.e. none, fully, with small limits):

```log
[ERROR] ...
 <v1.ResourceList | len:2>: {
              "cpu": {
                  i: {value: 1, scale: 0},
                  d: {Dec: nil},
                  s: "1",
                  Format: "DecimalSI",
              },
              "memory": {
                  i: {value: 268435456, scale: 0},
                  d: {Dec: nil},
                  s: "",
                  Format: "BinarySI",
              },
          }
      to equal
          <v1.ResourceList | len:2>: {
              "cpu": {
                  i: {value: 1000, scale: -3},
                  d: {Dec: nil},
                  s: "",
                  Format: "DecimalSI",
              },
              "memory": {
                  i: {value: 268435456, scale: 0},
                  d: {Dec: nil},
                  s: "",
                  Format: "BinarySI",
              },
          }
```

ebaron (commented):
Hey @tthvo, looks good! I have a suggestion below and some questions about the tests.

Could we avoid the additional ResourceSpecCase by using the resource requirements in the CR as our expectation? The tests should either ensure that the container resources match what's in the CR, or use the defaults if nothing is provided in the CR, with the exception of the limit < request case.

Does that make sense to you? I'm just trying to avoid making the tests more complex than they already are.

ebaron:
> I think its ready for review now :D Just a bit complications with resource limit/request map since a nil map is treated as no-specification while empty map will give "0" as requested/limited.
> 
> For tests, I have to do manual testing for resource value to avoid this issue below (perhaps deployment is modified for cleaner view) and some hacks to check resources spec cases (i.e. none, fully, with small limits):
> 
> ```
> [ERROR] ...
>  <v1.ResourceList | len:2>: {
>               "cpu": {
>                   i: {value: 1, scale: 0},
>                   d: {Dec: nil},
>                   s: "1",
>                   Format: "DecimalSI",
>               },
>               "memory": {
>                   i: {value: 268435456, scale: 0},
>                   d: {Dec: nil},
>                   s: "",
>                   Format: "BinarySI",
>               },
>           }
>       to equal
>           <v1.ResourceList | len:2>: {
>               "cpu": {
>                   i: {value: 1000, scale: -3},
>                   d: {Dec: nil},
>                   s: "",
>                   Format: "DecimalSI",
>               },
>               "memory": {
>                   i: {value: 268435456, scale: 0},
>                   d: {Dec: nil},
>                   s: "",
>                   Format: "BinarySI",
>               },
>           }
> ```

This method might be helpful for testing: https://pkg.go.dev/k8s.io/apimachinery@v0.25.3/pkg/api/resource#Quantity.Equal
You could do `Expect(actualCPU.Equal(expectedCPU)).To(BeTrue())`

tthvo:
> Could we avoid the additional ResourceSpecCase by using the resource requirements in the CR as our expectation? The tests should either ensure that the container resources match what's in the CR, or use the defaults if nothing is provided in the CR, with the exception of the limit < request case.
> 
> Does that make sense to you? I'm just trying to avoid making the tests more complex than they already are.

Right! Does this mean in context "with resource requirements", we return the CR resource requirements while other we return defaults? And no tests for limit < request case? Sorry I might misunderstand your idea.

tthvo:
> This method might be helpful for testing: https://pkg.go.dev/k8s.io/apimachinery@v0.25.3/pkg/api/resource#Quantity.Equal You could do `Expect(actualCPU.Equal(expectedCPU)).To(BeTrue())`

Right! This is nice! I used `Equal` and `Cmp` for comparison now instead of `Value()` and `MiliValue()`.

ebaron:
> > Could we avoid the additional ResourceSpecCase by using the resource requirements in the CR as our expectation? The tests should either ensure that the container resources match what's in the CR, or use the defaults if nothing is provided in the CR, with the exception of the limit < request case.
> > Does that make sense to you? I'm just trying to avoid making the tests more complex than they already are.
> 
> Right! Does this mean in context "with resource requirements", we return the CR resource requirements while other we return defaults? And no tests for limit < request case? Sorry I might misunderstand your idea.

I just meant that instead of adding an additional test input containing the expectation for resources, we could just use the information that's already in the CR. Of course this would add a bit extra logic to the tests, but it shouldn't be too complex. Something like:
```
if requestDefinedInCR {
  request = requestFromCR
} else if limitDefinedInCR && limit < defaultRequest {
  request = limitFromCR
} else {
  request = defaultRequest
}
```

I didn't intend to remove any test cases though. Testing all the different code paths is ideal, so make sure to keep the "limit < request" case. I hope that clarifies it a bit.

tthvo:
> > > Could we avoid the additional ResourceSpecCase by using the resource requirements in the CR as our expectation? The tests should either ensure that the container resources match what's in the CR, or use the defaults if nothing is provided in the CR, with the exception of the limit < request case.
> > > Does that make sense to you? I'm just trying to avoid making the tests more complex than they already are.
> > 
> > 
> > Right! Does this mean in context "with resource requirements", we return the CR resource requirements while other we return defaults? And no tests for limit < request case? Sorry I might misunderstand your idea.
> 
> I just meant that instead of adding an additional test input containing the expectation for resources, we could just use the information that's already in the CR. Of course this would add a bit extra logic to the tests, but it shouldn't be too complex. Something like:
> 
> ```
> if requestDefinedInCR {
>   request = requestFromCR
> } else if limitDefinedInCR && limit < defaultRequest {
>   request = limitFromCR
> } else {
>   request = defaultRequest
> }
> ```
> 
> I didn't intend to remove any test cases though. Testing all the different code paths is ideal, so make sure to keep the "limit < request" case. I hope that clarifies it a bit.

Ohh right! That makes sense :D I will fix that.

tthvo:
To check `limit < defaultRequest ` in test, I made a copy of function `checkWithLimit` from controller since the check logic is pretty much the same.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -993,14 +1054,33 @@ func NewGrafanaContainer(cr *operatorv1beta1.Cryostat, imageTag string, tls *TLS
...
+	}
+	requests := resources.Requests
+	if _, found := requests[corev1.ResourceCPU]; !found {
+		requests[corev1.ResourceCPU] = resource.MustParse(defaultJfrDatasourceCpuRequest)
+	}
+	if _, found := requests[corev1.ResourceMemory]; !found {
+		requests[corev1.ResourceMemory] = resource.MustParse(defaultJfrDatasourceMemoryRequest)
+	}
+	checkResourceRequestWithLimit(requests, resources.Limits)
+	return resources
```
ebaron, line 1081: This block of code is repeated for each container. Is it possible to extract it to a helper function to reduce duplication?

tthvo: Righ, i will try to refactor this.

internal/controllers/cryostat_controller_test.go
```diff
@@ -2593,20 +2615,21 @@ func (t *cryostatTestInput) checkMainPodTemplate(deployment *appsv1.Deployment,
 	emptyDir := cr.Spec.StorageOptions != nil && cr.Spec.StorageOptions.EmptyDir != nil && cr.Spec.StorageOptions.EmptyDir.Enabled
 	builtInDiscoveryDisabled := cr.Spec.TargetDiscoveryOptions != nil && cr.Spec.TargetDiscoveryOptions.BuiltInDiscoveryDisabled
 	dbSecretProvided := cr.Spec.JmxCredentialsDatabaseOptions != nil && cr.Spec.JmxCredentialsDatabaseOptions.DatabaseSecretName != nil
+
 	checkCoreContainer(&coreContainer,
 		t.minimal, t.TLS, t.externalTLS, t.EnvCoreImageTag,
 		t.controller.IsOpenShift, ingress, reportsUrl,
 		cr.Spec.AuthProperties != nil, emptyDir, builtInDiscoveryDisabled, dbSecretProvided,
-		cr.Spec.Resources.CoreResources, test.NewCoreSecurityContext(cr))
+		*test.NewCoreContainerResource(cr), test.NewCoreSecurityContext(cr))
```
ebaron, line 2623: I would change the resources parameter for these functions to be a pointer instead of dereferencing when calling them.

internal/controllers/cryostat_controller_test.go
```diff
@@ -2797,21 +2818,55 @@ func checkReportsContainer(container *corev1.Container, tls bool, tag *string, r
...
+		limitCpu, limitCpuFound := containerResource.Limits[corev1.ResourceCPU]
+		expectedLimitCpu, expectedLimitCpuFound := expectedResource.Limits[corev1.ResourceCPU]
+
+		Expect(limitCpuFound).To(Equal(expectedLimitCpuFound))
+		if expectedLimitCpuFound {
+			Expect(limitCpu.Equal(expectedLimitCpu)).To(BeTrue())
+		}
+
+		limitMemory, limitMemoryFound := containerResource.Limits[corev1.ResourceCPU]
+		expectedlimitMemory, expectedLimitMemoryFound := expectedResource.Limits[corev1.ResourceCPU]
```
ebaron, line 2865: These should be `ResourceMemory`.

tthvo: Ahh! Must be copy-paste error :d

ebaron (approved):
Looks good! Thanks @tthvo!

PR #442, opened 2022-08-28
fix(auth-properties): update docs links and add missing custom cluster-role in tests

tthvo:
Related to #440 

Updated config docs with permanent links to permission mapping files.

Added missing custom cluster-role in unit tests for auth-properties.


ebaron (approved):
Seems good! Thanks for fixing this.

Issue #444, opened 2022-09-02
Add a custom scorecard test


ebaron:
We should start leveraging custom scorecard tests in order to write some integration tests against a running cluster [1]. An example of such tests is given here: https://github.com/operator-framework/tekton-scorecard-image. To begin with, we can simply verify that the operator deployment starts up.

[1] https://sdk.operatorframework.io/docs/testing-operators/scorecard/custom-tests/

PR #445, opened 2022-09-06
test(scorecard): add scorecard test for operator availability

ebaron:
This adds the first custom scorecard test for the operator. This test is pretty basic, it just verifies that the operator deployment becomes available. It should help to serve as a simple smoke test to ensure that the operator can at least start.

I followed the instructions in the Operator SDK guide on custom scorecard tests for the most part. [1] The sample test image for Tekton was also a useful resource [2]. The tests run in a separate namespace with the operator bundle installed and RBAC objects used by the test pod created.

To test it out:
1. Build the custom scorecard test image and push to Quay.
    ```
    export CUSTOM_SCORECARD_IMG="quay.io/<your name>/cryostat-operator-scorecard:some-unique-tag"
    make scorecard-build 
    podman push "$CUSTOM_SCORECARD_IMG"
    ```
2. Generate bundle and push to Quay (with CUSTOM_SCORECARD_IMG defined from step 1).
    ```
    export BUNDLE_IMG="quay.io/<your name>/cryostat-operator-bundle:latest"
    make bundle bundle-build
    podman push "$BUNDLE_IMG"
    ```
3. Run scorecard tests (with running cluster)
    ```
    make test-scorecard
    ```

You should see a result for the custom test like the following:
```
--------------------------------------------------------------------------------
Image:      quay.io/<your name>/cryostat-operator-scorecard:some-unique-tag
Entrypoint: [cryostat-scorecard-tests operator-install]
Labels:
	"test":"operator-install"
	"suite":"cryostat"
Results:
	Name: operator-install
	State: pass

	Log:
		deployment cryostat-operator-controller-manager is available


--------------------------------------------------------------------------------
```

Fixes: #444 

[1] https://sdk.operatorframework.io/docs/testing-operators/scorecard/custom-tests/
[2] https://github.com/operator-framework/tekton-scorecard-image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,9 @@ OPM_VERSION ?= 1.23.0
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION ?= 1.24 
 
+CUSTOM_SCORECARD_VERSION = $(IMAGE_VERSION)
+CUSTOM_SCORECARD_IMG ?= $(IMAGE_TAG_BASE)-scorecard:$(CUSTOM_SCORECARD_VERSION)
+
 DEPLOY_NAMESPACE ?= cryostat-operator-system
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
@@ -318,6 +321,18 @@ create_cryostat_cr: destroy_cryostat_cr
 .PHONY: destroy_cryostat_cr
 destroy_cryostat_cr:
 	- $(CLUSTER_CLIENT) delete -f config/samples/operator_v1beta1_cryostat.yaml
+
+# Build custom scorecard tests
+.PHONY: custom-scorecard-tests
+custom-scorecard-tests: fmt vet
+	cd internal/images/custom-scorecard-tests/ && \
+	go build -o bin/cryostat-scorecard-tests main.go
+
+# Build the custom scorecard OCI image
+.PHONY: scorecard-build
+scorecard-build: custom-scorecard-tests
+	cd internal/images/custom-scorecard-tests/ && \
+	BUILDAH_FORMAT=docker $(IMAGE_BUILDER) build -t $(CUSTOM_SCORECARD_IMG) .
 
 # Local development/testing helpers
 
--- a/Makefile
+++ b/Makefile
@@ -123,17 +123,13 @@ ifneq ($(SKIP_TESTS), true)
 endif
 
 .PHONY: test-scorecard
-test-scorecard: clean-scorecard custom-scorecard-patch
+test-scorecard: clean-scorecard
 ifneq ($(SKIP_TESTS), true)
 	$(CLUSTER_CLIENT) create namespace $(SCORECARD_NAMESPACE)
 	$(CLUSTER_CLIENT) -n $(SCORECARD_NAMESPACE) create -f internal/images/custom-scorecard-tests/rbac/
 	operator-sdk run bundle -n $(SCORECARD_NAMESPACE) $(BUNDLE_IMG)
 	operator-sdk scorecard -n $(SCORECARD_NAMESPACE) -s cryostat-scorecard -w 5m $(BUNDLE_IMG)
 endif
-
-.PHONY: custom-scorecard-patch
-custom-scorecard-patch:
-	envsubst < hack/custom.config.yaml.in > config/scorecard/patches/custom.config.yaml
 
 .PHONY: clean-scorecard
 clean-scorecard:
@@ -193,6 +189,7 @@ manifests: controller-gen
 	$(CONTROLLER_GEN) rbac:roleName=role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 	envsubst < hack/image_tag_patch.yaml.in > config/default/image_tag_patch.yaml
 	envsubst < hack/image_pull_patch.yaml.in > config/default/image_pull_patch.yaml
+	envsubst < hack/custom.config.yaml.in > config/scorecard/patches/custom.config.yaml
 
 # Run go fmt against code
 .PHONY: fmt
--- a/Makefile
+++ b/Makefile
@@ -342,8 +342,8 @@ custom-scorecard-tests: fmt vet
 # Build the custom scorecard OCI image
 .PHONY: scorecard-build
 scorecard-build: custom-scorecard-tests
-	cd internal/images/custom-scorecard-tests/ && \
-	BUILDAH_FORMAT=docker $(IMAGE_BUILDER) build -t $(CUSTOM_SCORECARD_IMG) .
+	BUILDAH_FORMAT=docker $(IMAGE_BUILDER) build -t $(CUSTOM_SCORECARD_IMG) \
+	-f internal/images/custom-scorecard-tests/Dockerfile .
 
 # Local development/testing helpers
 
--- a/Makefile
+++ b/Makefile
@@ -129,11 +129,12 @@ ifneq ($(SKIP_TESTS), true)
 	$(CLUSTER_CLIENT) -n $(SCORECARD_NAMESPACE) create -f internal/images/custom-scorecard-tests/rbac/
 	operator-sdk run bundle -n $(SCORECARD_NAMESPACE) $(BUNDLE_IMG)
 	operator-sdk scorecard -n $(SCORECARD_NAMESPACE) -s cryostat-scorecard -w 5m $(BUNDLE_IMG)
+	- $(CLUSTER_CLIENT) delete --ignore-not-found=$(ignore-not-found) namespace $(SCORECARD_NAMESPACE)
 endif
 
 .PHONY: clean-scorecard
 clean-scorecard:
-	- $(CLUSTER_CLIENT) delete namespace $(SCORECARD_NAMESPACE)
+	- $(CLUSTER_CLIENT) delete --ignore-not-found=$(ignore-not-found) namespace $(SCORECARD_NAMESPACE)
 
 # Build manager binary
 .PHONY: manager
--- a/Makefile
+++ b/Makefile
@@ -78,10 +78,13 @@ OPM_VERSION ?= 1.23.0
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION ?= 1.24 
 
-CUSTOM_SCORECARD_VERSION = $(IMAGE_VERSION)
-CUSTOM_SCORECARD_IMG ?= $(IMAGE_TAG_BASE)-scorecard:$(CUSTOM_SCORECARD_VERSION)
+# Scorecard ImagePullPolicy is hardcoded to IfNotPresent
+# See: https://github.com/operator-framework/operator-sdk/pull/4762
+CUSTOM_SCORECARD_VERSION = $(IMAGE_VERSION)00001
+export CUSTOM_SCORECARD_IMG ?= $(IMAGE_TAG_BASE)-scorecard:$(CUSTOM_SCORECARD_VERSION)
 
 DEPLOY_NAMESPACE ?= cryostat-operator-system
+SCORECARD_NAMESPACE ?= cryostat-operator-scorecard
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
@@ -120,10 +123,21 @@ ifneq ($(SKIP_TESTS), true)
 endif
 
 .PHONY: test-scorecard
-test-scorecard: destroy_cryostat_cr undeploy uninstall
+test-scorecard: clean-scorecard custom-scorecard-patch
 ifneq ($(SKIP_TESTS), true)
-	operator-sdk scorecard bundle
+	$(CLUSTER_CLIENT) create namespace $(SCORECARD_NAMESPACE)
+	$(CLUSTER_CLIENT) -n $(SCORECARD_NAMESPACE) create -f internal/images/custom-scorecard-tests/rbac/
+	operator-sdk run bundle -n $(SCORECARD_NAMESPACE) $(BUNDLE_IMG)
+	operator-sdk scorecard -n $(SCORECARD_NAMESPACE) -s cryostat-scorecard -w 5m $(BUNDLE_IMG)
 endif
+
+.PHONY: custom-scorecard-patch
+custom-scorecard-patch:
+	envsubst < hack/custom.config.yaml.in > config/scorecard/patches/custom.config.yaml
+
+.PHONY: clean-scorecard
+clean-scorecard:
+	- $(CLUSTER_CLIENT) delete namespace $(SCORECARD_NAMESPACE)
 
 # Build manager binary
 .PHONY: manager
--- a/Makefile
+++ b/Makefile
@@ -123,8 +123,10 @@ ifneq ($(SKIP_TESTS), true)
 endif
 
 .PHONY: test-scorecard
-test-scorecard: clean-scorecard
+test-scorecard:
 ifneq ($(SKIP_TESTS), true)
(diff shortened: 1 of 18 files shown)


andrewazores:
I get a test failure like this:

```
--------------------------------------------------------------------------------
Image:      quay.io/andrewazores/cryostat-operator-scorecard:1
Entrypoint: [cryostat-scorecard-tests operator-install]
Labels:
	"suite":"cryostat"
	"test":"operator-install"
Results:
	State: fail

	Errors:
		invalid character 'c' looking for beginning of value
	Log:
		cryostat-scorecard-tests: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by cryostat-scorecard-tests)
		cryostat-scorecard-tests: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by cryostat-scorecard-tests)
```

tthvo:
Oh somehow that failure does not appear for me (`crc` with Openshift 4.10.18):

```bash
Image:      quay.io/thvo/cryostat-operator-scorecard:0.0.0-custom-scorecard
Entrypoint: [cryostat-scorecard-tests operator-install]
Labels:
	"test":"operator-install"
	"suite":"cryostat"
Results:
	Name: operator-install
	State: pass

	Log:
		deployment cryostat-operator-controller-manager is available
```

andrewazores:
Not sure if this is relevant, but it's a difference:

```
$ crc version
CRC version: 2.8.0+217b3bd
OpenShift version: 4.11.1
Podman version: 4.1.1
```

tthvo:
Hmm still passed with 

```bash
CRC version: 2.8.0+217b3bd
OpenShift version: 4.11.1
Podman version: 4.1.1
```

Here is my log:

```bash
Image:      quay.io/thvo/cryostat-operator-scorecard:0.0.0-custom-scorecard
Entrypoint: [cryostat-scorecard-tests operator-install]
Labels:
	"suite":"cryostat"
	"test":"operator-install"
Results:
	Name: operator-install
	State: pass

	Log:
		deployment cryostat-operator-controller-manager is available
```

config/scorecard/patches/basic.config.yaml
```diff
@@ -4,7 +4,7 @@
     entrypoint:
     - scorecard-test
     - basic-check-spec
-    image: quay.io/operator-framework/scorecard-test:v1.5.0
+    image: quay.io/operator-framework/scorecard-test:v1.22.0
```
tthvo, line 7: Should we use tag `v1.22.2` to match our current `operator-sdk` version? 

https://quay.io/repository/operator-framework/scorecard-test?tab=tags&tag=v1.22.2

internal/images/custom-scorecard-tests/rbac/scorecard_role.yaml
```diff
@@ -0,0 +1,48 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# Permissions for Cryostat custom Scorecard tests
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: cryostat-scorecard
```
tthvo, line 41: I wonder without a namespace, the role will assume the namespace in the current context, right?

tthvo:
I also saw PodSecurity violation for scorecard pods:

```bash
W0908 14:00:00.662049   49213 warnings.go:70] would violate PodSecurity "restricted:v1.24": allowPrivilegeEscalation != false (containers "scorecard-untar", "scorecard-test" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (containers "scorecard-untar", "scorecard-test" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or containers "scorecard-untar", "scorecard-test" must set securityContext.runAsNonRoot=true), seccompProfile (pod or containers "scorecard-untar", "scorecard-test" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

I think this fix depends on OperatorSDK: https://github.com/operator-framework/operator-sdk/issues/5939?

Makefile
```diff
@@ -319,6 +333,18 @@ create_cryostat_cr: destroy_cryostat_cr
 destroy_cryostat_cr:
 	- $(CLUSTER_CLIENT) delete -f config/samples/operator_v1beta1_cryostat.yaml
 
+# Build custom scorecard tests
+.PHONY: custom-scorecard-tests
+custom-scorecard-tests: fmt vet
```
tthvo, line 338: I am wondering if we should build this binary as a step in setting up the scorecard image in its `internal/images/custom-scorecard-tests/Dockerfile`? I think it would be consistent across dev environment?

config/scorecard/kustomization.yaml
```diff
@@ -14,3 +14,9 @@ patchesJson6902:
     kind: Configuration
     name: config
 # +kubebuilder:scaffold:patchesJson6902
```
tthvo, line 16: I wonder if this marker should be at the end of the patch list? Not sure if it makes any changes?

config/scorecard/patches/basic.config.yaml
ebaron: Good catch.

internal/images/custom-scorecard-tests/rbac/scorecard_role.yaml
ebaron: This is used in the Makefile by `$(CLUSTER_CLIENT) -n $(SCORECARD_NAMESPACE) create -f internal/images/custom-scorecard-tests/rbac/`. So this will get created in the `$(SCORECARD_NAMESPACE)` namespace.

ebaron:
> I also saw PodSecurity violation for scorecard pods:
> 
> ```shell
> W0908 14:00:00.662049   49213 warnings.go:70] would violate PodSecurity "restricted:v1.24": allowPrivilegeEscalation != false (containers "scorecard-untar", "scorecard-test" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (containers "scorecard-untar", "scorecard-test" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or containers "scorecard-untar", "scorecard-test" must set securityContext.runAsNonRoot=true), seccompProfile (pod or containers "scorecard-untar", "scorecard-test" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
> ```
> 
> I think this fix depends on OperatorSDK: [operator-framework/operator-sdk#5939](https://github.com/operator-framework/operator-sdk/issues/5939)?

We should keep an eye out for this. There will probably be a lot of Pod Security Admission related fixes coming.

Makefile
ebaron: I'll change the Dockerfile to build the binary similar to the operator binary. Hopefully this will resolve Andrew's issue.

config/scorecard/kustomization.yaml
ebaron: I'm not sure what this is used for. It's generated by the Operator SDK/Kubebuilder. I left it where it was because everything prior is generated YAML.

internal/images/custom-scorecard-tests/rbac/scorecard_role.yaml
tthvo: Ahh right! Thanks alot for explaining :D

config/scorecard/kustomization.yaml
tthvo: OperatorSDK seems to lack documentations on `# +kubebuilder:scaffold:*` :D I think, as tests are working as expected, it is okay to leave till they release the documentations: https://github.com/kubernetes-sigs/kubebuilder/issues/1487

ebaron:
> I get a test failure like this:
> 
> ```
> --------------------------------------------------------------------------------
> Image:      quay.io/andrewazores/cryostat-operator-scorecard:1
> Entrypoint: [cryostat-scorecard-tests operator-install]
> Labels:
> 	"suite":"cryostat"
> 	"test":"operator-install"
> Results:
> 	State: fail
> 
> 	Errors:
> 		invalid character 'c' looking for beginning of value
> 	Log:
> 		cryostat-scorecard-tests: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by cryostat-scorecard-tests)
> 		cryostat-scorecard-tests: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by cryostat-scorecard-tests)
> ```

I'm hoping this is now fixed with the Dockerfile building the tests in a multi-stage build. Mind trying to build an image and test again?

Makefile
```diff
@@ -117,11 +123,18 @@ ifneq ($(SKIP_TESTS), true)
...
+test-scorecard: clean-scorecard
 ifneq ($(SKIP_TESTS), true)
-	operator-sdk scorecard bundle
+	$(CLUSTER_CLIENT) create namespace $(SCORECARD_NAMESPACE)
+	$(CLUSTER_CLIENT) -n $(SCORECARD_NAMESPACE) create -f internal/images/custom-scorecard-tests/rbac/
+	operator-sdk run bundle -n $(SCORECARD_NAMESPACE) $(BUNDLE_IMG)
+	operator-sdk scorecard -n $(SCORECARD_NAMESPACE) -s cryostat-scorecard -w 5m $(BUNDLE_IMG)
 endif
 
+.PHONY: clean-scorecard
```
tthvo, line 134: Our custom scorecard test resources appear to persist after tests. Should we remove them after tests like other scorecards?

internal/images/custom-scorecard-tests/main.go
```diff
@@ -0,0 +1,102 @@
...
+const podBundleRoot = "/bundle"
+
+func main() {
+	entrypoint := os.Args[1:]
+	if len(entrypoint) == 0 {
+		log.Fatal("specify one or more test name arguments")
+	}
+
+	// Get namespace from SCORECARD_NAMESPACE environment variable
+	namespace := os.Getenv("SCORECARD_NAMESPACE")
```
tthvo, line 61: This variable `SCORECARD_NAMESPACE` is set for our scorecard container automatically when the pod is created right?

Makefile
ebaron: Sure, I added a delete command after the tests run. Also included the `--ignore-not-found` argument used elsewhere.

internal/images/custom-scorecard-tests/main.go
ebaron: Right, this is automatically set by Scorecard.

Makefile
tthvo: Great!

andrewazores:
> > I get a test failure like this:
> > ```
> > --------------------------------------------------------------------------------
> > Image:      quay.io/andrewazores/cryostat-operator-scorecard:1
> > Entrypoint: [cryostat-scorecard-tests operator-install]
> > Labels:
> > 	"suite":"cryostat"
> > 	"test":"operator-install"
> > Results:
> > 	State: fail
> > 
> > 	Errors:
> > 		invalid character 'c' looking for beginning of value
> > 	Log:
> > 		cryostat-scorecard-tests: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by cryostat-scorecard-tests)
> > 		cryostat-scorecard-tests: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by cryostat-scorecard-tests)
> > ```
> 
> I'm hoping this is now fixed with the Dockerfile building the tests in a multi-stage build. Mind trying to build an image and test again?

Nice, worked well this time.

Makefile
```diff
@@ -117,11 +123,19 @@ ifneq ($(SKIP_TESTS), true)
 endif
 
 .PHONY: test-scorecard
-test-scorecard: destroy_cryostat_cr undeploy uninstall
+test-scorecard: clean-scorecard
```
andrewazores, line 126: I guess since the scorecard tests run in a separate namespace now, it's not necessary to clear out any previous installation? Would it make sense to fail early on this run if the namespace already exists, or even exists and is not empty?

internal/images/custom-scorecard-tests/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Read the pod's untar'd bundle from a well-known path.
+	bundle, err := apimanifests.GetBundleFromDir(podBundleRoot)
+	if err != nil {
+		log.Fatalf("failed to read bundle manifest: %s", err.Error())
+	}
+
+	var result scapiv1alpha3.TestStatus
+
+	switch entrypoint[0] {
```
andrewazores, line 74: It's fine for now since there is only one test that can be invoked, but I think we may as well prepare to handle the case where multiple tests exist and multiple can be specified. The `entrypoint := os,Args[1:]` already gets us halfway there - I think basically this `switch` just needs to be iterated over and the index used instead of `[0]`.

Maybe it's also worth doing two passes - the first to make sure all the named tests are known/valid, and the second to actually run them.

Makefile
ebaron: The namespace could be non-empty if any of the test steps encounter an error. In this case, the recipe would exit before running the `delete namespace` command. I suppose it's probably safer to make the user delete the namespace manually.

tthvo: I am thinking whether instead of deleting the namespace before tests, we could clean up only operator resources with `operator-sdk cleanup` and ignore if namespace does not exist?

internal/images/custom-scorecard-tests/main.go
tthvo: Great idea!! I just noticed the operator-sdk set up the scorecard tests like:

https://github.com/ebaron/cryostat-operator/blob/5b1501ac567c51b6a176f680985d0104026ddc21/config/scorecard/patches/olm.config.yaml#L1

It seems they launch a separate container (same scorecard image) for each test. Should we follow them too?

ebaron: This is actually what's being done already, I just followed the Operator SDK example. One test per container, which is why the boilerplate in main.go considers only `entrypoint[0]`. That said, there doesn't appear to be any reason in the API why the container can't run more than one test. I'll modify main.go to support this.

Makefile
ebaron: As we get more complicated tests, there could be additional objects that get created that aren't cleaned up by `operator-sdk cleanup`. In order to ensure the tests have a clean slate, it's ideal to wipe out the namespace. A fresh cluster is even better, but not the most practical.

tthvo: Ahh right! Thanks for explaining :D

ebaron: Added a check for the namespace which prompts the user to delete it, if it exists.

internal/images/custom-scorecard-tests/main.go
ebaron: I reworked the `main.go` to support multiple test arguments now.

internal/images/custom-scorecard-tests/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+func runTests(testNames []string, bundle *apimanifests.Bundle, namespace string) []scapiv1alpha3.TestResult {
+	results := []scapiv1alpha3.TestResult{}
+
+	// Run tests
+	for _, testName := range testNames {
+		switch testName {
+		case tests.OperatorInstallTestName:
+			results = append(results, tests.OperatorInstallTest(bundle, namespace))
+		default:
```
tthvo, line 118: Looks great!  I just noticed since we validateTests, this default case won't be reachable right?

ebaron: Not normally, it's just there as a sanity test if we add a new test to `validateTests` but forget to add it to `runTests`.

tthvo: Right! That is a good idea :D

tthvo (approved):
Looks great to me :D

Issue #447, opened 2022-09-09
Segfault possible if report options omitted


ebaron:
Our sample CR includes a default of setting `reportOptions.replicas` to 0.
```yaml
apiVersion: operator.cryostat.io/v1beta1
kind: Cryostat
metadata:
  name: cryostat-sample
spec:
  # ...
  reportOptions:
    replicas: 0
```
This should be equivalent to leaving out the reportOptions section altogether, but results in a segfault when doing so.
```
goroutine 403 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.12.1/pkg/internal/controller/controller.go:118 +0x1f4
panic({0x159cd00, 0x2521fc0})
	/usr/local/go/src/runtime/panic.go:838 +0x207
github.com/cryostatio/cryostat-operator/internal/controllers/common/resource_definitions.NewCoreContainer(_, _, {_, _}, _, _)
	/workspace/internal/controllers/common/resource_definitions/resource_definitions.go:640 +0x61d
github.com/cryostatio/cryostat-operator/internal/controllers/common/resource_definitions.NewPodForCR(0xc000414400, 0xc000627c50?, 0xc0006dcfc0, 0xc00046f1a0, 0x4844, 0x0)
	/workspace/internal/controllers/common/resource_definitions/resource_definitions.go:240 +0x199
github.com/cryostatio/cryostat-operator/internal/controllers/common/resource_definitions.NewDeploymentForCR(0xc000414400, 0xc0004bb3b0?, 0x19ea358?, 0xc00027a1b0?, 0x19ebb20?, 0xe0?)
	/workspace/internal/controllers/common/resource_definitions/resource_definitions.go:177 +0x3c5
github.com/cryostatio/cryostat-operator/internal/controllers.(*CryostatReconciler).Reconcile(0xc0004bb3b0, {0x19ea358, 0xc00027a1b0}, {{{0xc000620660?, 0x10?}, {0xc000661640?, 0x40d787?}}})
	/workspace/internal/controllers/cryostat_controller.go:305 +0xccd
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x19ea2b0?, {0x19ea358?, 0xc00027a1b0?}, {{{0xc000620660?, 0x16b6640?}, {0xc000661640?, 0x4041f4?}}})
[...]

Issue #448, opened 2022-09-09
Replace fake client with envtest in tests


ebaron:
There are some cases where the fake client doesn't reflect the behaviour of a real API server. This bug contains one example: https://github.com/cryostatio/cryostat-operator/issues/447#issuecomment-1242491587. We already spin up envtest as part of our test suite, we just don't use it. This is closer to a real API server than the fake client represents. We should switch to using it.

ebaron:
The Kubebuilder book has some good examples of testing controllers with envtest: https://book.kubebuilder.io/cronjob-tutorial/writing-tests.html

PR #449, opened 2022-09-09
fix(reports): add nil-check and remove side effects

ebaron:
Unfortunately it's not possible to add a regression test for this while using the fake client (see #448). Although after removing any side effects modifying `spec.reportOptions`, the segfault occurs as expected in the tests. Adding the nil-check in `NewCoreContainer` fixes it.

Fixes: #447


tthvo (approved):
Looks great to me! Thanks for fixing! It makes sense we shouldn't modify CR spec here. I could work on moving FakeClient to envtest later :D

Issue #404, opened 2022-06-06
Update workloads for "restricted" Pod Security Standard


ebaron:
Our operator and Cryostat workloads will require some minor tweaks to work with the `restricted` Pod Security Standard:
https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted

This tool can be used to verify the Pod Security level of the workloads: https://github.com/stlaz/psachecker

ebaron:
A useful resource on how operators should handle this: https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/1417

Issue #446, opened 2022-09-06
Security Context should be configurable with Cryostat CRD


ebaron:
With [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/), this is fairly important to ensure our workloads can run on a variety of clusters. It doesn't seem like one set of defaults will work for everybody. 

At the API level, I expect something like the following:

```go
type CryostatSpec struct {
// ...
	// Options to configure the Security Contexts for the Cryostat application.
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	SecurityOptions *SecurityOptions `json:"securityOptions,omitempty"`
}

type ReportConfiguration struct {
// ...
	// Options to configure the Security Contexts for the Cryostat report generator.
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	SecurityOptions *ReportsSecurityOptions `json:"securityOptions,omitempty"`
}

// SecurityOptions contains Security Context customizations for the
// main Cryostat application at both the pod and container level.
type SecurityOptions struct {
	// Security Context to apply to the Cryostat pod
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	PodSecurityContext *corev1.PodSecurityContext `json:"podSecurityContext,omitempty"`
	// Security Context to apply to the Cryostat application container
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	CoreSecurityContext *corev1.SecurityContext `json:"coreSecurityContext,omitempty"`
	// Security Context to apply to the JFR Data Source container
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	DataSourceSecurityContext *corev1.SecurityContext `json:"dataSourceSecurityContext,omitempty"`
	// Security Context to apply to the Grafana container
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	GrafanaSecurityContext *corev1.SecurityContext `json:"grafanaSecurityContext,omitempty"`
}

// ReportsSecurityOptions contains Security Context customizations for the
// Cryostat report generator at both the pod and container level.
type ReportsSecurityOptions struct {
	// Security Context to apply to the Cryostat report generator pod
	// +optional
	// +operator-sdk:csv:customresourcedefinitions:type=spec
	PodSecurityContext *corev1.PodSecurityContext `json:"podSecurityContext,omitempty"`
[...]

PR #450, opened 2022-09-12
fix(security): support restricted pod security standard

ebaron:
The primary difficulty with this fix is that OpenShift < 4.11 does not permit you to set `seccompProfile` by default. Following the discussion in https://github.com/redhat-openshift-ecosystem/community-operators-prod/discussions/1417, I decided to take the following approach with this PR.

For the operator deployment:
- Set `runAsNonRoot` and `seccompProfile` in the pod security context. This is not backwards compatible with OpenShift < 4.11. In the downstream release, we will have to remove `seccompProfile` to support these versions. The `restricted-v2` Security Context Constraint (SCC) will set the `seccompProfile` for us when it is necessary. The alternative is not being compliant on standard Kubernetes, which does not have a SCC mechanism. The end result will be upstream releases will continue to work on Kubernetes 1.19, and OpenShift >= 4.11. Downstream releases will continue to work on OpenShift >= 4.6.
- Drop all capabilities and disallow privilege escalation.

For the operand deployments:
- Set `runAsNonRoot` and if not running on OpenShift, set `seccompProfile`. This preserves backwards compatibility, and relies on the `restricted-v2` SCC to set `seccompProfile` in OpenShift >= 4.11.
- Drop all capabilities and disallow privilege escalation.

Fixes: #404

### DIFF
--- a/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
@@ -442,8 +442,13 @@ spec:
                     memory: 64Mi
                 securityContext:
                   allowPrivilegeEscalation: false
+                  capabilities:
+                    drop:
+                    - ALL
               securityContext:
                 runAsNonRoot: true
+                seccompProfile:
+                  type: RuntimeDefault
               serviceAccountName: cryostat-operator-service-account
               terminationGracePeriodSeconds: 10
       permissions:
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -29,6 +29,9 @@ spec:
         name: manager
         securityContext:
           allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
         livenessProbe:
           httpGet:
             path: /healthz
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -20,6 +20,8 @@ spec:
       serviceAccountName: service-account
       securityContext:
         runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       containers:
       - command:
         - /manager
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -374,9 +374,11 @@ func NewPodForCR(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTags *I
 		volumes = append(volumes, authResourceVolume)
 	}
 
-	// Ensure PV mounts are writable
+	nonRoot := true
 	sc := &corev1.PodSecurityContext{
-		FSGroup: &fsGroup,
+		// Ensure PV mounts are writable
+		FSGroup:      &fsGroup,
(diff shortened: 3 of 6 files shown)


tthvo:
Looks great! From some readings, I did manual runs on Openshift `4.11.1` & `4.10.18` and Minikube `1.26.0`. It worked well! I just have a couple questions:

1. Our sample apps seem not to deployed in `restricted-v2` (got an alert for violation but still allowed). I think the recipe still works because I logged in as kubeadmin. Maybe we could specify some field with `oc new-app`?
2. For our custom scorecard tests, we create a namespace. I wonder if we should also test in a namespace which enforce `restricted` PSA? (not sure I understand the context right)
3. Maybe we could start testing our Cryostat CR in scorecard too to check if pods are admitted?

tthvo:
I believe the security context set here for cryostat containers are default and will be overwritted from CR spec as in #446 ?

ebaron:
> Looks great! From some readings, I did manual runs on Openshift `4.11.1` & `4.10.18` and Minikube `1.26.0`. It worked well! I just have a couple questions:

Did this work on 4.10.18? I expected the operator deployment would not be accepted because of the `seccompProfile` field.

> 
>     1. Our sample apps seem not to deployed in `restricted-v2` (got an alert for violation but still allowed). I think the recipe still works because I logged in as kubeadmin. Maybe we could specify some field with `oc new-app` and update demo images?

I'm not sure if there's any parameters we could set for this in `new-app`. Another solution would be to add the deployment and service YAML manifests to our repo, and create them with `$(CLUSTER_CLIENT) create -f`. This would also avoid the dependency on `oc` instead of `kubectl`.

> 
>     2. For our custom scorecard tests, we create a namespace. I wonder if we should also test in a namespace which enforce `restricted` PSA? (not sure I understand the context right)

That seems like a good idea. We should do that once https://github.com/operator-framework/operator-sdk/issues/5939 is fixed.

> 
>     3. Maybe we could start testing our Cryostat CR in scorecard too to check if pods are admitted?

Yes absolutely. The next scorecard test I want to write is one that creates the Cryostat CR and ensures that Cryostat starts properly.

ebaron:
> I believe the security context set here for cryostat containers are default and will be overwritted from CR spec as in #446 ?

That's right.

ebaron:
@tthvo I created 3 new issues for each point in https://github.com/cryostatio/cryostat-operator/pull/450#issuecomment-1248301740

tthvo:
> Did this work on 4.10.18? I expected the operator deployment would not be accepted because of the seccompProfile field.

Hmm interestingly, now that I give a second thought. Operator pod is actually admitted and launched.

tthvo:
> Hmm right! I totally forgot that. Interestingly, Operator pod is actually admitted and launched in 4.10.18 with seccompProfile set.

**Editted**: The default namespace should not used as SCC does not apply.  Worked as expected on 4.10.18! Thanks for pointing me to this :D

https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html#role-based-access-to-ssc_configuring-internal-oauth

tthvo (approved):
Looks great to me! Worked as expected :D

PR #453, opened 2022-09-14
fix(makefile): use ignore-not-found flag when deleting resources

tthvo:
Related to #430 

Specified `--ignore-not-found` for any `$(KUBE_CLIENT) delete` calls.


ebaron (approved):
Looks good! Thanks for keeping it consistent.

PR #457, opened 2022-09-19
feat(security): security contexts are now configurable

### LINKED ISSUE #446: Security Context should be configurable with Cryostat CRD

tthvo:
Fixes #446 

Users can now specify security contexts in Cryostat CR to overwrite defaults.


tthvo:
I added the spec field under `Advanced`. Any thoughts?

Screenshot from 2022-09-19 14-14-25
Screenshot from 2022-09-19 14-14-47

docs/config.md
```diff
@@ -309,3 +309,53 @@ spec:
 Each `configMapName` must refer to the name of a Config Map in the same namespace as Cryostat. The corresponding `filename` must be a key within that Config Map containing resource mappings. The `clusterRoleName` must be a valid name of an existing Cluster Role.
 
 **Note:** If the mapping is updated, Cryostat must be manually restarted.
+
+
+### Security Context
+
+With [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/), pods that are not properly configured under the enforced security standards defined globally or on namspace level will be admitted to launch.
+
+The user is responsible for ensuring the security contexts of their workloads to meet these standards. The property `spec.securityOptions` can be set to define security contexts for Cryostat application and its report sidecar.
```
ebaron, line 318: You might want to reference `spec.reportOptions.securityOptions` for the latter.

ebaron: We might also want to mention the defaults we use. The defaults conform to the `restricted` Pod Security Standard [1]. For the Cryostat application pod, the operator selects an `fsGroup` to ensure that Cryostat can read and write files in its Persistent Volume.

On OpenShift, `spec.securityContext.seccompProfile` is left unset for backward compatibility. On versions of OpenShift supporting Pod Security Admission, the `restricted-v2` Security Context Constraint sets `seccompProfile` to `runtime/default` as required for the `restricted` Pod Security Standard. [2]

[1] https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
[2] https://docs.openshift.com/container-platform/4.11/authentication/managing-security-context-constraints.html#default-sccs_configuring-internal-oauth

ebaron (commented):
Hey @tthvo, looks good so far! I'm still working through it, but here are a couple suggestions for the documentation.

docs/config.md
tthvo: Right! Theses are really important! I will add these!

internal/test/resources.go
```diff
@@ -1007,7 +1027,11 @@ func NewCoreEnvironmentVariables(minimal bool, tls bool, externalTLS bool, opens
 				Value: "custom-auth-cluster-role",
 			})
 		}
+		envs = append(envs, NewCoreAndGrafanaURLEnvironementVariables(minimal, tls, externalTLS)...)
+	} else if ingress { // On Kubernetes
+		envs = append(envs, NewCoreAndGrafanaURLEnvironementVariables(minimal, tls, externalTLS)...)
```
ebaron, line 1032: Is this `ingress` branch useful? It seems like the same function call in both cases. Maybe I'm missing something though.

tthvo: Oh yes. I was adding this branch so that tests in k8s that do not have Ingress should not add external-url-related variable since the service won't be configured.

ebaron: Thanks, I understand now. I'm a bit torn about this. One one hand, Cryostat will not work without these environment variables. On the other, it's good to have one test that ensures that the operator doesn't crash if the Ingress configs are not specified on Kubernetes. I think this is fine for now, and we can revisit in #403.

Maybe calling the function `newNetworkEnvironmentVariables` might make it a bit clearer.

tthvo: Right! It would be better with that new name :D

tthvo: > think this is fine for now, and we can revisit in https://github.com/cryostatio/cryostat-operator/issues/403.

Oh yess! After that, I think I could remove this branch and these vars for both.

ebaron (approved):
Works as expected. Thanks again @tthvo!

Issue #409, opened 2022-06-13
Existing RBAC objects should be updated


ebaron:
Currently the Service Account is updated when it differs from the operator's spec. The Role, RoleBinding and ClusterRoleBinding also need to be updateable. The ClusterRole is a manifest provided by the operator and is unaffected.

PR #458, opened 2022-09-23
feat(rbac): update operator-controlled RBAC objects

ebaron:
This is mostly refactoring RBAC reconciling logic into a separate `rbac.go` file. This PR also ensures that Roles, RoleBindings and ClusterRoleBindings created by the operator are kept up-to-date across operator upgrades, and includes new test cases to exercise this. There are a couple of other changes here:
1. The reports pod now uses the "default" Service Account instead of the Cryostat one. The reports generator doesn't require access to the Kubernetes API, so the elevated permissions given to Cryostat aren't necessary here.
2. Instead of controlling the value of `AutomountServiceAccountToken` in the Service Account, we simply set this to true for the Cryostat pod template. Cryostat depends on its Service Account token being mounted upon startup. The definition in the Pod spec takes priority over any setting in the Service Account. [1]

Fixes: #409 

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server

### DIFF
--- a/internal/controllers/cryostat_controller_test.go
+++ b/internal/controllers/cryostat_controller_test.go
@@ -337,7 +337,83 @@ var _ = Describe("CryostatController", func() {
 
 				Expect(sa.ImagePullSecrets).To(Equal(oldSA.ImagePullSecrets))
 				Expect(sa.Secrets).To(Equal(oldSA.Secrets))
-				Expect(sa.AutomountServiceAccountToken).To(BeNil())
+				Expect(sa.AutomountServiceAccountToken).To(Equal(oldSA.AutomountServiceAccountToken))
+			})
+		})
+		Context("with an existing Role", func() {
+			var cr *operatorv1beta1.Cryostat
+			var oldRole *rbacv1.Role
+			BeforeEach(func() {
+				cr = test.NewCryostat()
+				oldRole = test.OtherRole()
+				t.objs = append(t.objs, cr, oldRole)
+			})
+			It("should update the Role", func() {
+				t.reconcileCryostatFully()
+
+				role := &rbacv1.Role{}
+				err := t.Client.Get(context.Background(), types.NamespacedName{Name: "cryostat", Namespace: "default"}, role)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(metav1.IsControlledBy(role, cr)).To(BeTrue())
+
+				// Labels are unaffected
+				Expect(role.Labels).To(Equal(oldRole.Labels))
+
+				// Rules should be fully replaced
+				Expect(role.Rules).To(Equal(test.NewRole().Rules))
+			})
+		})
+		Context("with an existing Role Binding", func() {
+			var cr *operatorv1beta1.Cryostat
+			var oldBinding *rbacv1.RoleBinding
+			BeforeEach(func() {
+				cr = test.NewCryostat()
+				oldBinding = test.OtherRoleBinding()
+				t.objs = append(t.objs, cr, oldBinding)
+			})
+			It("should update the Role Binding", func() {
+				t.reconcileCryostatFully()
+
+				binding := &rbacv1.RoleBinding{}
+				err := t.Client.Get(context.Background(), types.NamespacedName{Name: "cryostat", Namespace: "default"}, binding)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(metav1.IsControlledBy(binding, cr)).To(BeTrue())
+
+				// Labels are unaffected
+				Expect(binding.Labels).To(Equal(binding.Labels))
+
+				// Subjects and RoleRef should be fully replaced
+				expected := test.NewRoleBinding()
+				Expect(binding.Subjects).To(Equal(expected.Subjects))
+				Expect(binding.RoleRef).To(Equal(expected.RoleRef))
+			})
+		})
+		Context("with an existing Cluster Role Binding", func() {
+			var cr *operatorv1beta1.Cryostat
+			var oldBinding *rbacv1.ClusterRoleBinding
+			BeforeEach(func() {
+				cr = test.NewCryostat()
+				oldBinding = test.OtherClusterRoleBinding()
+				t.objs = append(t.objs, cr, oldBinding)
+			})
+			It("should update the Cluster Role Binding", func() {
+				t.reconcileCryostatFully()
+
+				binding := &rbacv1.ClusterRoleBinding{}
+				err := t.Client.Get(context.Background(), types.NamespacedName{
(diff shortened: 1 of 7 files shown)


andrewazores (commented):
LGTM

internal/controllers/rbac.go
```diff
@@ -0,0 +1,312 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	r.Log.Info(fmt.Sprintf("Cluster Role Binding %s", op), "name", binding.Name, "namespace", binding.Namespace)
+	return nil
+}
+
+func (r *CryostatReconciler) deleteClusterRoleBinding(ctx context.Context, cr *operatorv1beta1.Cryostat) error {
+	clusterBinding := newClusterRoleBinding(cr) // TODO move to finalizeRBAC, add tests
```
tthvo, line 300: Is this still a `TODO`?

internal/controllers/cryostat_controller_test.go
```diff
@@ -2152,8 +2228,9 @@ func (t *cryostatTestInput) checkReportsDeployment() {
 	}
 
 	checkReportsContainer(&template.Spec.Containers[0], t.TLS, t.EnvReportsImageTag, resources, test.NewReportSecurityContext(cr))
-	// Check that the proper Service Account is set
-	Expect(template.Spec.ServiceAccountName).To(Equal("cryostat"))
+	// Check that the default Service Account is used
+	Expect(template.Spec.ServiceAccountName).To(BeEmpty())
```
tthvo, line 2232: This is only specific to `FakeClient`, right? In actual cluster, it would be set to `default`?

internal/controllers/cryostat_controller_test.go
```diff
@@ -337,7 +337,83 @@ var _ = Describe("CryostatController", func() {
...
+				t.reconcileCryostatFully()
+
+				binding := &rbacv1.RoleBinding{}
+				err := t.Client.Get(context.Background(), types.NamespacedName{Name: "cryostat", Namespace: "default"}, binding)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(metav1.IsControlledBy(binding, cr)).To(BeTrue())
+
+				// Labels are unaffected
+				Expect(binding.Labels).To(Equal(binding.Labels))
```
tthvo, line 385: I think the second one should be `oldBinding`?

internal/controllers/cryostat_controller_test.go
```diff
@@ -337,7 +337,83 @@ var _ = Describe("CryostatController", func() {
...
+				t.reconcileCryostatFully()
+
+				binding := &rbacv1.ClusterRoleBinding{}
+				err := t.Client.Get(context.Background(), types.NamespacedName{
+					Name: "cryostat-9ecd5050500c2566765bc593edfcce12434283e5da32a27476bc4a1569304a02",
+				}, binding)
+				Expect(err).ToNot(HaveOccurred())
+
+				// Labels are unaffected
+				Expect(binding.Labels).To(Equal(binding.Labels))
```
tthvo, line 411: should be `oldBinding`?

internal/controllers/cryostat_controller_test.go
```diff
@@ -337,7 +337,83 @@ var _ = Describe("CryostatController", func() {
...
+				t.reconcileCryostatFully()
+
+				role := &rbacv1.Role{}
+				err := t.Client.Get(context.Background(), types.NamespacedName{Name: "cryostat", Namespace: "default"}, role)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(metav1.IsControlledBy(role, cr)).To(BeTrue())
+
+				// Labels are unaffected
+				Expect(role.Labels).To(Equal(oldRole.Labels))
```
tthvo, line 361: Should we check `annotations` too?

tthvo (commented):
Looks great and worked as expected! Just a couple of comments. Also, I wonder if these steps should always be applied to Kubernetes Context too?

internal/controllers/rbac.go
ebaron: Nope, thanks for noticing that. I've removed it.

internal/controllers/cryostat_controller_test.go
ebaron: This is for the real cluster too. When the `serviceAccountName` is left empty, the default service account is used: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server

internal/controllers/cryostat_controller_test.go
ebaron: Good catch! This would have tested nothing otherwise.

internal/controllers/cryostat_controller_test.go
ebaron: Doesn't hurt to include those as well.

ebaron:
@tthvo I addressed your feedback and added the tests to the Kubernetes branch as well. I have mostly avoided duplicating the tests if there are no logic changes between OpenShift and Kubernetes, but there actually is one in this case. The OAuth annotation is only present on OpenShift. I'm hoping after the release we can tidy up this test class a bit, it's very long. Perhaps we can avoid some copy-paste by using this `DescribeTable` Ginkgo feature: https://onsi.github.io/ginkgo/#table-specs?

internal/controllers/cryostat_controller_test.go
tthvo: Ahh right! I just realized golang string default-value is empty. Thanks for explaining that :D

tthvo:
> @tthvo I addressed your feedback and added the tests to the Kubernetes branch as well. I have mostly avoided duplicating the tests if there are no logic changes between OpenShift and Kubernetes, but there actually is one in this case. The OAuth annotation is only present on OpenShift. I'm hoping after the release we can tidy up this test class a bit, it's very long. Perhaps we can avoid some copy-paste by using this `DescribeTable` Ginkgo feature: https://onsi.github.io/ginkgo/#table-specs?

Right! I was wondering that too. This Ginkgo feature would help a lot! I think we could file an issue for this?

ebaron:
> > @tthvo I addressed your feedback and added the tests to the Kubernetes branch as well. I have mostly avoided duplicating the tests if there are no logic changes between OpenShift and Kubernetes, but there actually is one in this case. The OAuth annotation is only present on OpenShift. I'm hoping after the release we can tidy up this test class a bit, it's very long. Perhaps we can avoid some copy-paste by using this `DescribeTable` Ginkgo feature: https://onsi.github.io/ginkgo/#table-specs?
> 
> Right! I was wondering that too. This Ginkgo feature would help a lot! I think we could file an issue for this?

Sure, I filed #461. Any other concerns with this PR?

tthvo (approved):
Looks great to me :D

Issue #412, opened 2022-06-13
Existing PersistentVolumeClaim should be updated


ebaron:
Existing PVCs should be updated in response to config changes within the CR or changed definitions in the operator. We need to ensure we only attempt to modify fields that can be modified.

There are some fields that cannot be modified, e.g. with the PVC:
```
Forbidden: spec is immutable after creation except resources.requests for bound claims
```

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/issues/294#issuecomment-1081138341_

PR #462, opened 2022-09-30
feat(pvc): update existing PVCs 

ebaron:
This PR causes the operator to update Persistent Volume Claims when the CR spec is modified. My initial draft of this PR attempted to check against PVC spec modifications that would be rejected by the API server. It turns out the validation logic for PVC updates is fairly complex, and can depend on the Storage Class used as well [1][2]. Because of this, it's not the most practical to consider all invalid cases and we might end up preventing a valid update from succeeding.

I reworked this PR to let the user attempt to update the PVC spec as they see fit, and intercept a validation error from the API server and emit a warning event for the Cryostat CR when this update is invalid. To simulate this invalid update with the tests, I added a client wrapper that injects an error when attempting to update a specified object.

I also included a minor unrelated change to use `DeepCopy` instead of modifying the service configs stored within the CR. This is to avoid potential side-effects from modifying the CR. This is the same approach I added in this PR for the PVC spec stored in the CR.

Fixes: #412 

[1] https://github.com/kubernetes/kubernetes/blob/0fff0242763c29d09b2e830a5013087eb74fb325/pkg/apis/core/validation/validation.go#L2161-L2220
[2] https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#persistentvolumeclaimresize


ebaron:
There are still some conflicts with the Persistent Volume controller, converting this to a draft.

ebaron:
This should be ready for review now.

I ran into cases where the operator was fighting against two different Kubernetes controllers. Once for the `VolumeName` field and annotations, and again for the `StorageClassName` when using dynamic provisioning. Our operator and the controllers would infinitely override each other and this ended up blocking our controller from proceeding further in installing/updating Cryostat. 

Since this is pretty severe and depends on how the cluster is configured, I decided to only allow one `.spec` field to be updated after creation and that's the resource requests [1]. Typically the resource requests can only be increased, but there are corner cases where they can be decreaesd. We'll accept any resource request modification and let Kubernetes determine if it's valid. If not, we emit a warning Event.

For the annotations, there are quite a few prefixes currently used, and these could differ among Kubernetes versions [2]. To avoid conflicts, we merge the existing annotations and labels with those in the CR. If the user needs to delete an annotation or label, they'll have to remove it from the PVC directly.

[1] https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims
[2] https://github.com/kubernetes/kubernetes/blob/a330d3b9721ff2a9c78554d390a134c905c32c99/staging/src/k8s.io/component-helpers/storage/volume/pv_helpers.go#L33-L82

internal/controllers/pvc.go
```diff
@@ -0,0 +1,151 @@
...
+
+// Event type to inform users of invalid PVC specs
+const eventPersistentVolumeClaimInvalidType = "PersistentVolumeClaimInvalid"
+
+func (r *CryostatReconciler) reconcilePVC(ctx context.Context, cr *operatorv1beta1.Cryostat) error {
+	emptyDir := cr.Spec.StorageOptions != nil && cr.Spec.StorageOptions.EmptyDir != nil && cr.Spec.StorageOptions.EmptyDir.Enabled
+	if emptyDir {
+		// If user requested an emptyDir volume, then do nothing.
+		// Don't delete the PVC to prevent accidental data loss
+		// depending on the reclaim policy.
```
tthvo, line 59: Hi, I got a question about this part :D

If previously a persistent volume is claimed and then switched to emptyDir, this still means the `PV` is still owned by `Cryostat` pod and Cryostat can still put data on the PV?

internal/controllers/cryostat_controller_test.go
```diff
@@ -831,6 +836,67 @@ var _ = Describe("CryostatController", func() {
...
+					Expect(err).ToNot(HaveOccurred())
+
+					// Expect an Invalid status error after reconciling
+					req := reconcile.Request{NamespacedName: types.NamespacedName{Name: "cryostat", Namespace: "default"}}
+					_, err = t.controller.Reconcile(context.Background(), req)
+					Expect(err).To(HaveOccurred())
+					Expect(kerrors.IsInvalid(err)).To(BeTrue())
+				})
+				It("should emit a PersistentVolumeClaimInvalid event", func() {
+					recorder := t.controller.EventRecorder.(*record.FakeRecorder)
```
tthvo, line 893: Could you explain this syntax here? Sorry, I couldn't understand it (first time seeing this).

tthvo (commented):
Looks good and worked as expected!

Just a question about how storage options are shown on OpenShift. I think the below settings are for emptyDir, but shown under `Storage Options`. Probably, that's why I got bit confused above.

Screenshot from 2022-10-07 14-05-53

tthvo:
I could not fully test the Cryostat since we needs #465. I will get it done soon.

internal/controllers/cryostat_controller_test.go
```diff
@@ -831,6 +836,67 @@ var _ = Describe("CryostatController", func() {
 				t.expectPVC(test.NewDefaultPVCWithLabel())
 			})
 		})
+		Context("with an existing PVC", func() {
```
tthvo, line 839: Just a thought: 

Should we refactor this test to assume exising PVC is valid, then we add two case:

1. Valid -> reconcile -> successfuly updates
2. Invalid -> Inject client -> reconcile -> Then, check recorder.

Just to be consistent flow between cases in this `existing PVC` context and closer to PVC is updated with our controller. What do you think?

internal/controllers/cryostat_controller_test.go
ebaron: This is a [type assertion[(https://go.dev/ref/spec#Type_assertions). It's similar to a cast. In this case the controller's `EventRecorder` field is of the `record.EventRecorder` interface type. Instead of the real EventRecorder, which talks to the Kubernetes API server, the tests use the controller-runtime project's `FakeRecorder`. This also implements the `record.EventRecorder` interface, but only simulates talking to the API server. Because the test case needs to access the `Events` channel field of the `FakeRecorder` type, we use a type assertion to make our `recorder` variable of type `record.FakeRecorder` instead of the `record.EventRecorder` interface it implements.

tthvo: Oh right! Make sense! Thanks a lot for explaining this! I probably should go back to revise some Golang basics :D

ebaron:
> Looks good and worked as expected!
> 
> Just a question about how storage options are shown on OpenShift. I think the below settings are for emptyDir, but shown under `Storage Options`. Maybe, we should also add some xDesc for PVC so that they are also shown.
> 
> Screenshot from 2022-10-07 14-05-53

That would be nice, but it we wouldn't be able to do this using marker comments in the code. Since we embedded an existing type `corev1.PersistentVolumeClaimSpec`, which we don't control, we can't add x-descriptors through marker comments. We would have to add them through a Kustomize patch like this: https://github.com/cryostatio/cryostat-operator/blob/def0b0df06fa05fca03e36ed9fe21b2d1eca3818/config/manifests/xdescriptors_patch.yaml.

So, this is likely doable for the field types that have existing x-descriptors, but not as straightforward as it seems.

internal/controllers/cryostat_controller_test.go
ebaron: Perhaps I am misunderstanding, but I think what you're describing is what the tests currently do. These new test cases create a valid PVC as a precondition to the test. The success case checks whether this PVC is updated (where possible) to match the config in the CR. The failure case injects the client which rejects the PVC update and checks the recorder.

tthvo: Oh wait, right! Sorry, I was looking at the overall structure. But the better phrasing would be: for the second `invalid` case, should we add an existing PVC in default object list in `BeforeEach`?

tthvo: https://github.com/ebaron/cryostat-operator/blob/86b057dce4ad7439bd15fd15968a8d60cfa47f90/internal/controllers/cryostat_controller_test.go#L868

ebaron: Ah, I see what you mean. In the failure case, the existing PVC is created by the controller. I'll see if I can rework it a bit so the success and failure cases have a more similar flow.

ebaron: I've revised the tests, let me know what you think. While it appears simpler, one drawback is it relies on the ordering that the controller reconciles object types. PVCs are one of the first now, but if they got moved after the TLS setup for example, once call of Reconcile wouldn't be sufficient because we need to update Certificate statuses.

I have an idea to not rely on this `reconcileCryostatFully` to do mock status updates in a particular order. This causes flexibility problems when writing tests, like this one. I'll save that for another PR though.

tthvo (approved):
Looks good to me! About the x-descriptor issue, I have been playing around with some but i could not find ones that are suitable to display. I will look into it later :D

Issue #413, opened 2022-06-13
Existing Secrets should updated


ebaron:
Since the JMX and Grafana auth, and Keystore secrets contain generated passwords, we should not update those. Otherwise the password would change at each reconcile. We should however ensure that the Secrets remain owned by the Cryostat CR.

PR #466, opened 2022-10-06
feat(secrets): update operator-controlled secrets

ebaron:
Depends on: #412 
Fixes: #413


internal/controllers/secrets.go
```diff
@@ -0,0 +1,134 @@
...
+
+	operatorv1beta1 "github.com/cryostatio/cryostat-operator/api/v1beta1"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (r *CryostatReconciler) reconcileSecrets(ctx context.Context, cr *operatorv1beta1.Cryostat) error {
+	err := r.reconcileGrafanaSecret(ctx, cr)
```
tthvo, line 51: Should we generate Grafana Secret if deployment is minimal? Also, any clean up if deployment is switched to minimal from non-minimal?

ebaron: Ideally no, but there is no harm in doing so. I'll see if I can change it quickly.

tthvo (commented):
Looks good! I just have the question about the minimal deployment :D I will start on adding `JMX_CREDENTIALS_DB_PASSWORD ` after this PR.

internal/controllers/secrets.go
ebaron: I've updated this PR to only create/update the secret for the non-minimal configuration, and delete the secret for the minimal configuration. This also removes the Secret link from the CR Status when deleting the secret.

I had to revise the password generation in the tests to accommodate this by adding a slice of generated passwords to the test input. Seems to work well.

ebaron:
> Looks good! I just have the question about the minimal deployment :D I will start on adding `JMX_CREDENTIALS_DB_PASSWORD ` after this PR.

I build a core image which is one commit before the introduction of `JMX_CREDENTIALS_DB_PASSWORD`. That should aid in testing the PR.
`make deploy OPERATOR_IMG=<your build> CORE_IMG=quay.io/ebaron/cryostat:before-db`

andrewazores:
Looks good from a reading and it seems to work fine from a basic build and deploy test on CRC. Is there anything I should do to manually test and exercise this?

internal/controllers/secrets.go
tthvo: Great! Thanks for making edits... :D

ebaron:
One way to test it is to modify the username within the secret. The operator should reset the change. LIkewise, you can modify the password within the secret. In this case, the operator should not reset the change.

internal/controllers/secrets.go
ebaron: Right, that would make sense to do. I think that might be getting a bit beyond the scope of this PR since that's handled by the certificates. I filed #471 for this.

andrewazores (approved):

internal/controllers/secrets.go
tthvo: Right make sense! Thanks :D

tthvo (approved):
Looks great :D

Issue #414, opened 2022-06-13
Existing ConsoleLinks should be updated


ebaron:
The ConsoleLink spec should be managed by the operator and updated when necessary.

PR #467, opened 2022-10-06
feat(openshift): update existing console links

ebaron:
This is a fairly straightforward refactoring of the console links and API server config modification into a new `openshift.go` file. The new feature here is just a small change to update console link specs after they have been created.

Fixes: #414


internal/controllers/rbac.go
```diff
@@ -292,7 +292,7 @@ func (r *CryostatReconciler) createOrUpdateClusterRoleBinding(ctx context.Contex
 	if err != nil {
 		return err
 	}
-	r.Log.Info(fmt.Sprintf("Cluster Role Binding %s", op), "name", binding.Name, "namespace", binding.Namespace)
+	r.Log.Info(fmt.Sprintf("Cluster Role Binding %s", op), "name", binding.Name)
```
ebaron, line 295: Just a heads up, this is an extra bit of cleanup since Cluster Role Bindings do not have a namespace.

internal/controllers/openshift.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+		r.Log.Error(err, "failed to delete ConsoleLink", "name", link.Name)
+		return err
+	}
+	r.Log.Info("deleted ConsoleLink", "name", link.Name)
+	return nil
+}
+
+func (r *CryostatReconciler) addCorsAllowedOriginIfNotPresent(ctx context.Context, cr *operatorv1beta1.Cryostat) error {
+	reqLogger := r.Log.WithValues("Request.Namespace", cr.Namespace, "Request.Name", cr.Name)
```
tthvo, line 130: Can we use `CryostatReconciler.Log` here instead?

tthvo: Same for `deleteCorsAllowedOrigins`?

internal/controllers/openshift.go
```diff
@@ -0,0 +1,199 @@
...
+		allowedOriginAsRegex,
+	)
+
+	err = r.Client.Update(ctx, apiServer)
+	if err != nil {
+		reqLogger.Error(err, "Failed to update APIServer CORS allowed origins")
+		return err
+	}
+
+	return nil
```
tthvo, line 163: Should have some logs for this operations? Same for `deleteCorsAllowedOrigins`?

internal/controllers/openshift.go
```diff
@@ -0,0 +1,199 @@
...
+		// Nothing to link to, so remove the link if it exists
+		err := r.deleteConsoleLink(ctx, link)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, link, func() error {
+		link.Spec.Link = consolev1.Link{
+			Text: "Cryostat",
```
tthvo, line 99: This will patched downstream to something like `Red Hat build of Cryostat`?

tthvo (commented):
Looks great! Just a couple of questions :D

internal/controllers/openshift.go
```diff
@@ -0,0 +1,199 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	consolev1 "github.com/openshift/api/console/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+func (r *CryostatReconciler) reconcileOpenShift(ctx context.Context, cr *operatorv1beta1.Cryostat) error {
+	if !r.IsOpenShift {
```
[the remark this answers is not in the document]
ebaron: I'd prefer to keep this here. My goal is to simplify the main CryostatReconciler by offloading as much logic to the helpers as possible.

internal/controllers/openshift.go
ebaron: This is using that log, but configuring it so the request name and namespace are added to each log message. This is useful when logging things to do with cluster-scoped resources where it's not obvious which CR caused the log message.

internal/controllers/openshift.go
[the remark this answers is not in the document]
tthvo: Right make sense!

internal/controllers/openshift.go
tthvo: Oh yehh from a second look, I understand now. Thanks for explaining :D

internal/controllers/openshift.go
ebaron: Good point. I've modified our generated code to include a constant for the application name that's configurable at build time.

internal/controllers/openshift.go
ebaron: Good idea. I've updated the logging for this file.

Issue #415, opened 2022-06-13
Update list of watched objects


ebaron:
This list is out of date:
https://github.com/cryostatio/cryostat-operator/blob/126e3b5459692a6f104cb660455c8de598691113/internal/controllers/cryostat_controller.go#L438-L439

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/issues/294#issuecomment-1139955770_

We should add:
- ServiceAccount
- Role
- RoleBinding
- Ingress
- Certificate (can these last two be included if cert-manager is absent from cluster?)
- Issuer

We should also use the `Owns` method instead of `Watches` for these, as suggested by the godocs.

PR #468, opened 2022-10-07
fix(watches): update object types watched by controller

ebaron:
This updates the kinds of objects that the controller watches to ensure that any change to an owned object triggers a reconcile. The cert-manager APIs pose a slight complication since if the resource types do not exist in the Kubernetes API server, the controller will fail to start.

I modified the OpenShift detection logic to also test for cert-manager. Unfortunately, there's no way yet to dynamically update the list of watched object types (https://github.com/kubernetes-sigs/controller-runtime/issues/1884). This means that if the user installs cert-manager after the controller has started, they will have to redeploy the operator to pick up that change.

I changed the setup from using the `Watches` method to the newer `Owns` which is recommended for this purpose.

Fixes: #415

### DIFF
--- a/internal/controllers/cryostat_controller.go
+++ b/internal/controllers/cryostat_controller.go
@@ -53,18 +53,19 @@ import (
 
 	"github.com/cryostatio/cryostat-operator/internal/controllers/common"
 	resources "github.com/cryostatio/cryostat-operator/internal/controllers/common/resource_definitions"
+	certv1 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1"
 	openshiftv1 "github.com/openshift/api/route/v1"
 	securityv1 "github.com/openshift/api/security/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	netv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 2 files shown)


internal/controllers/cryostat_controller.go
```diff
@@ -315,17 +317,18 @@ func (r *CryostatReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&operatorv1beta1.Cryostat{})
 
 	// Watch for changes to secondary resources and requeue the owner Cryostat
-	resources := []client.Object{&appsv1.Deployment{}, &corev1.Service{}, &corev1.Secret{}, &corev1.PersistentVolumeClaim{}}
+	resources := []client.Object{&appsv1.Deployment{}, &corev1.Service{}, &corev1.Secret{}, &corev1.PersistentVolumeClaim{},
+		&corev1.ServiceAccount{}, &rbacv1.Role{}, &rbacv1.RoleBinding{}, &netv1.Ingress{}}
```
tthvo, line 321: Should `ClusterRoleBinding` be included here too? Though, I am a bit unsure since our operator only works under single namespace now.

ebaron: Unfortunately it can't be because namespace-scoped objects (our CR) can't own cluster-scoped objects,

ebaron: > Also, I suppose these objects are watched only within the namespace of the operator?

Right, same as the CR itself. That will likely change in the future though.

tthvo: Right! That makes sense :D

tthvo (approved):
Looks good to me! I did some edits on owned objects and the reconciler is called as expected :D

Issue #463, opened 2022-10-03
Document scheduling config options


ebaron:
#340 introduced new options to affect pod scheduling onto nodes. We should add document these options in `docs/config.md`

PR #469, opened 2022-10-07
docs(scheduling): document scheduling options

ebaron:
Fixes: #463


andrewazores:
Seems to make sense, though I don't know if I have any way to test out this config.

ebaron:
> Seems to make sense, though I don't know if I have any way to test out this config.

It's not the easiest thing to test since you need a multi-node cluster. As long as the scheduling options are passed exactly as-is to the deployments, then you can infer correctness from that.

andrewazores (approved):

tthvo (approved):
Looks good :D

Issue #340, opened 2022-02-07
Add options to control scheduling on nodes


andrewazores:
Original comment: http://disq.us/p/2mek3sz

> Thanks for this demo and for developing Cryostat. It looks like a really good tool. On the demo, I'm having an issue completing it on a local multi-node Minikube cluster. There seems to be a [known issue](https://disq.us/url?url=https%3A%2F%2Fgithub.com%2Fkubernetes%2Fminikube%2Fissues%2F11765%3A8vN0X3Hm6-jUyrDbdK3c4mDWAAA&cuid=5264065) with multi-node Minikube and hostPath volumes. It seems that hostPath volumes always get created on the primary node (minikube). When running this demo my Cryostat Pod gets scheduled onto, for example, minikube-m02 node. It can't find the required hostPath volume there and fails with a CreateContainerConfigError status.
> Apart from just running it on a single node Minikube cluster, is there a workaround for this? I've tried to edit the Cryostat deployment and add a nodeSelector to force scheduling on the mimikube node, but perhaps because it is controlled by the operator, any changes I make don't take effect

CJT1:
Thanks for creating this @andrewazores . I think the issue is more to do with persistent volumes running on multi-node Minikube than anything specific to Cryostat. But if it is helpful, this is the setup I used when I experienced the problem:
3 Node Minikube cluster running:

- Minikube: v1.25.1
- VirtualBox: 6.1.32r149290
- Kubernetes: v1.22.5 

The k8s control plane is on a node called minkube

I got Cryostat working by ssh'ing onto the two worker nodes (minikube-m02, minikube-m03) and manually creating the directories Cryostat was expecting under _/tmp/hostpath-provisioner_. So I ended up with something like:
_/tmp/hostpath-provisioner/<cryostat namespace>/<cryostat name>_

The Cryostat Pod started up ok once the directory was created on the node

ebaron:
We should add the following options to the Cryostat CRD to allow the user to control which nodes the Cryostat and report deployments are scheduled to:
* [Node Selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector)
* [Affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity)
* [Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)

[...]

PR #470, opened 2022-10-07
fix(reports): fix panic in reports scheduling

ebaron:
This fixes a panic in the reports scheduling options when the main Cryostat scheduling options is not also set. This is caused by an incorrect nil-check. This also fixes a few problems in the reports tests where they did not test what they appeared to be testing.

Related to: #340


andrewazores (approved):

tthvo (approved):
Looks good :D

Issue #417, opened 2022-06-14
Increase Operator Capability Level


ebaron:
Once #294 is complete, I believe we qualify for bumping our Operator Capability Level from "Basic Install" to "Seamless Upgrades":
https://sdk.operatorframework.io/docs/overview/operator-capabilities/#level-2---seamless-upgrades

Issue #433, opened 2022-08-03
Deployment pod selector is immutable


ebaron:
We should avoid changing this as much as possible, but if we need to change the selector across upgrades, we'll need to delete and recreate the deployment.

Issue #294, opened 2021-11-10
[Epic] Modification of Cryostat CR spec should always update affected objects


ebaron:
The operator updates some objects when the Cryostat CR is modified, the Deployment spec is one example. This behaviour should be expanded to all objects managed by the operator. When reconciling the Cryostat CR, this can be done using `controllerutil.CreateOrUpdate` instead of checking for existence and using `client.Create`.

Tasks:
- [x] #408 
- [x] #409
- [x] #410
- [x] #411
- [x] #412
- [x] #413
- [x] #414
- [x] #415

ebaron:
There are some fields that cannot be modified, e.g. with the PVC:
```
Forbidden: spec is immutable after creation except resources.requests for bound claims
```

ebaron:
This should also update any annotations and labels required for Cryostat to function, such as labels used by the selectors for services. For example, we currently update the deployment's pod template, but not its labels or annotations.

ebaron:
We should also update the list of object kinds that we watch:
https://github.com/cryostatio/cryostat-operator/blob/126e3b5459692a6f104cb660455c8de598691113/internal/controllers/cryostat_controller.go#L438-L439

tthvo:
I guess we can close this issue too :D

PR #473, opened 2022-10-12
build(olm): move capability level to Seamless Upgrades

ebaron:
After the fixes for #294 and #433, I believe we should qualify for the "Seamless Upgrades" [operator capability level](https://sdk.operatorframework.io/docs/overview/operator-capabilities/). This PR updates our CSV to use that instead of "Basic Install". This is presented to the user as part of the install process on OperatorHub.io and OpenShift's OperatorHub.

Screenshot 2022-10-12 at 11-19-44 OperatorHub · Red Hat OpenShift Container Platform

Depends on: #433 
Fixes: #417


andrewazores:
I think it would qualify for the base line of a Seamless Upgrade-capable Operator, but there's probably also more we can do to better support future Cryostat minor semver upgrades. The Cryostat database schema could change for example, or more of our persistent data could be moved from JSON filesystem storage into the database, both of which would entail some kind of database or data migration scripts/actions that ideally the Operator would run, rather than that code being embedded into Cryostat itself.

tthvo:
Just a question on this: This capability specification is likely self specified by the operator right? There is not really any test suite/checks for this?

ebaron:
> Just a question on this: This capability specification is likely self specified by the operator right? There is not really any test suite/checks for this?

Right, I'm not aware of any test for this yet. There is mention of adding this to Scorecard, but nothing yet: https://github.com/operator-framework/operator-sdk/issues/5339

ebaron:
> Yea, an init container was along what I was thinking, but using Cryostat's entrypoint.bash could also work. Easy enough to add a hook there to run any scripts found in some `migrations.d` before doing anything else. The operator could then add scripts and mount the `migrations.d` volume. Those same scripts would be published upstream so the Operator can just package them, and end users (including helm users?) can figure their own way to run them either directly or by hooking into the entrypoint.

Do you expect that these scripts would vary based on the environment Cryostat is running in? I'm just having a hard time understanding why this wouldn't ship as part of Cryostat itself. I was thinking this would be something like checking if the database schema version is out of date, and if so, run the necessary SQL commands to update it. Cryostat would need to do this no matter where it's running wouldn't it? Maybe I'm not thinking about it the right way.

andrewazores:
Also, more specifically to Postgres in the future, that Postgres instance that the user has configured Cryostat to use may not be used exclusively by Cryostat - it could be a shared database that the user's own workloads are already using. In that case we shouldn't try to run any manual migrations and should leave that up to the user, so now the entrypoint or the migration script need some configuration to prevent themselves from running even if the entrypoint does detect that Cryostat is configured to talk to a Postgres instance.

And in that scenario, the user should really have Postgres deployed with an admin account for themselves, and separate DB user accounts for their workloads and for Cryostat - and Cryostat's account probably shouldn't even have permissions to modify the DB schema. It would be a migration for the admin user to run separately before doing the Cryostat version upgrade.

ebaron:
These are fair points. We'll have to discuss in more detail when planning the next release. Considering there are no database migrations required yet, is it acceptable to upgrade our capability level yet?

andrewazores:
I think so. It does mean we're committing to having that automatic migration ability built in to Cryostat and/or the Operator by the time we have a release that might need to use such migrations, but I think that's fair. Having some automatic system for migrations to be applied on upgrade should be part of the planning process for any changes that would require such a migration.

ebaron:
> I think so. It does mean we're committing to having that automatic migration ability built in to Cryostat and/or the Operator by the time we have a release that might need to use such migrations, but I think that's fair. Having some automatic system for migrations to be applied on upgrade should be part of the planning process for any changes that would require such a migration.

Makes sense. Even if it slows down feature progression, I think it's worthwhile to ensure that users can upgrade.

andrewazores (approved):

Issue #465, opened 2022-10-05
Generate credentials DB password


ebaron:
After https://github.com/cryostatio/cryostat/pull/1083, Cryostat requires `CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD` to be set. For now, we should generate this in the operator and store it as a secret. Ideally, we make this configurable later.

Issue #451, opened 2022-09-13
Cryostat should be deployed in `h2:file` mode


andrewazores:
These two environment variables should be set for the Cryostat container:

https://github.com/cryostatio/cryostat/blob/58bc83dd1e0145bf34cd3289f908a929ef664d6a/smoketest.sh#L22

If not set then the container defaults to running in `h2:mem` mode, which does not provide an actual persistent database. The discovery API would be not durable across reboots, which also means custom target definitions would not survive.

tthvo:
What if we allow customization for this behaviour? Maybe a spec field to set which type of storage to use?

Also, just wondering how `postgres` is etup our deployment on k8s/OCP context :D

andrewazores:
Yea, eventually we should make it configurable. Right now the change is too fresh and likely to change heavily for the next release too so I don't want to try to support a Postgres setup in the official distribution just yet.

I guess if the Cryostat CR has no storage allocated (no PVC), then the Operator could automatically configure Cryostat to use `h2:mem`. If there is a PVC then use `h2:file` with the file located somewhere on that PVC.

PR #474, opened 2022-10-12
feat(discovery): allow configurations for discovery service

### LINKED ISSUE #452: Add CR property to disable Cryostat built-in discovery mechanisms
The environment variable `CRYOSTAT_DISABLE_BUILTIN_DISCOVERY` can be set to `true` on the Cryostat container to disable the built-in discovery mechanisms. In an Operator context this would mean disabling Cryostat's `Endpoints` querying/watch. This would be used in cases where the end user only intends to have applications discovered via the Discovery Plugin API.

tthvo:
This PR includes configurations to discovery service:

- The operator now manages a Secret that will provide generated password for `CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD` variable.
- Built-in Discovery Service can now be disabled via Cryostat CR spec.
- Cryostat is deployed with `h2:file` if PVC is configured/defaulted. Otherwise, `h2:mem` is used.

I also added missing nil-checks in test resources.

Fixes: #452, #451, #465


internal/test/resources.go
```diff
@@ -2150,38 +2216,39 @@ func OtherServiceAccount() *corev1.ServiceAccount {
 }
 
 func NewRole() *rbacv1.Role {
+	rules := []rbacv1.PolicyRule{
```
tthvo, line 2219: In case built-in discovery is disabled, I believe we can remove these rules (or at least `endpoints`. But I am a bit unsure if these rules use cases are beyond built-in discovery mechanism. Right now, I don't remove any, but any suggestions on these?

andrewazores: True, I think `endpoints` is only used for the old-style discovery system. If the new `BuiltInDiscoveryDisabled` property is `true` then there should be nothing making k8s API requests for that specific resource type.

The rest look like permissions that the Operator needs to be able to set up the deployments and routes for things.

I don't remember exactly - are there two separate service accounts for the Operator and for Cryostat itself? Cryostat would only need `endpoints` from this list, and the Operator would need everything but `endpoints` I think (which used to be used for the FlightRecorder CR).

docs/config.md
```diff
@@ -436,3 +436,17 @@ spec:
         value: ok
         effect: NoExecute
 ```
+
+### Target Discovery Options
+
+If you wish to use only Cryostat's Discovery Plugin API, the property `spec.targetDiscoveryOptions.builtInDiscoveryDisabled` to `true` to disable Cryostat's built-in discovery mechanisms.
```
andrewazores, line 442: Probably worth linking to this doc: https://github.com/cryostatio/cryostat/blob/main/docs/DISCOVERY_PLUGINS.md

internal/controllers/secrets.go
```diff
@@ -127,6 +129,33 @@ func (r *CryostatReconciler) reconcileJMXSecret(ctx context.Context, cr *operato
...
+		},
+	}
+	return r.createOrUpdateSecret(ctx, secret, cr, func() error {
+		if secret.StringData == nil {
+			secret.StringData = map[string]string{}
+		}
+
+		// Password is generated, so don't regenerate it when updating
+		if secret.CreationTimestamp.IsZero() {
+			secret.StringData[databaseSecretPassKey] = r.GenPasswd(32)
```
andrewazores, line 153: What is the lifecycle of this secret? Would it be generated once when the Operator is installed or is it dropped and regenerated for each Cryostat CR instance?

tthvo: Oh, this secret is generated once (i.e. its password) for each Cryostat CR and garbage collected if CR is deleted.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -739,6 +739,35 @@ func NewCoreContainer(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTa
...
+			Name:  "CRYOSTAT_DISABLE_BUILTIN_DISCOVERY",
+			Value: fmt.Sprintf("%v", disableBuiltInDiscovery),
+		})
+	}
+
+	useEmptyDir := cr.Spec.StorageOptions != nil && cr.Spec.StorageOptions.EmptyDir != nil && cr.Spec.StorageOptions.EmptyDir.Enabled
+	if !useEmptyDir {
+		envs = append(envs, corev1.EnvVar{
+			Name:  "JDBC_URL",
+			Value: "jdbc:h2:file:/opt/cryostat.d/conf.d/h2;INIT=create domain if not exists jsonb as other",
```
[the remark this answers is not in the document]
tthvo: Right! I will fix this :D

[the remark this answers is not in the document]
tthvo: Should we also need to set `JDBC` name and password?

internal/test/resources.go
tthvo: Right, there are 2 separate SA for operator and Cryostat. I think the operator still needs `endpoints` access in order to gain that to Cryostat SA. Do you think these permissions are somewhat used for Automated Rules?

internal/controllers/secrets.go
tthvo: Hmm it's possible that Cryostat will re-gain access to old persistent volume (i.e. maybe after re-creation). Then, the new password can't be used. So, the user has to clean up the volume first?

internal/test/resources.go
andrewazores: Automated Rules goes through the same target discovery code as everything else within Cryostat, so Automated Rules only needs Endpoints indirectly if the KubeApiPlatformClient is in use (via the BuiltInDiscovery hooking it into the DiscoveryStorage). If the property/envvar is set and built-in discovery is disabled then Automated Rules, and other things, will continue just querying the DiscoveryStorage for what it does know about, and in this scenario it would not be populated by KubeApiPlatformClient so no Endpoints queries would ever be made.

andrewazores: It makes sense that the Operator might need to have a strict superset of the permissions it might ever grant to Cryostat. But in the case where this property is set and disables built-in discovery, the Cryostat SA shouldn't need that permission anymore and it should probably not be granted.

internal/controllers/common/resource_definitions/resource_definitions.go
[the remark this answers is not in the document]
andrewazores: For `h2:mem` the user/pass are irrelevant. This database is in-memory only so when the Cryostat JVM shuts down the database also disappears entirely.

For `h2:file`, if the username or password changes then Cryostat can't open the database file for access. If the database file is expected to be garbage collected and deleted along with Cryostat being undeployed then in effect it is also irrelevant. If a database file is expected to be available to reuse between different Cryostat instances (delete and recreate CR) then both of those Cryostat instances need to use the same username and password. In this case the JMX credentials keyring passphrase would also need to be the same between these two Cryostat instances.

When we add Postgres support later it'll have the same requirements as `h2:file`, plus the user/pass need to be configurable by the user along with the JDBC URL pointing at that database so that the user can set up their own DB instance and configure it and give Cryostat access.

internal/controllers/secrets.go
andrewazores: Either the user has to clean up the volume or the user can have the ability to specify their own password. I guess that would work by the user creating a Secret manually and linking to that in the CR. This way the Secret contains the password long-term, outside of the Cryostat/Operator lifecycles, so it can be consistent for different Cryostat deployments and instances if the user needs that.

tthvo: Right, this is most ideal. I think we can a spec in Cryostat CR for this and remove the managed secret now. Probably, have to document the required key name.

internal/controllers/secrets.go
```diff
@@ -127,6 +129,33 @@ func (r *CryostatReconciler) reconcileJMXSecret(ctx context.Context, cr *operato
 	})
 }
 
+// databaseSecretNameSuffix is the suffix to be appended to the name of a
+// Cryostat CR to name its JMX credentials Database secret
```
ebaron, line 133:
```suggestion
// Cryostat CR to name its JMX credentials database secret
```

internal/controllers/secrets.go
```diff
@@ -127,6 +129,33 @@ func (r *CryostatReconciler) reconcileJMXSecret(ctx context.Context, cr *operato
...
+const databaseSecretNameSuffix = "-jmx-credentials-db"
+
+// dbSecretUserKey indexes the password within the Cryostat JMX credentials Database Secret
```
ebaron, line 136:
```suggestion
// dbSecretUserKey indexes the password within the Cryostat JMX credentials database Secret
```

internal/controllers/common/resource_definitions/resource_definitions.go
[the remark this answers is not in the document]
tthvo: Right, I think it would make sense to mount these env vars too so later with Postgres, we can just override them with specs in Cryostat CR.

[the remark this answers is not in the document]
tthvo: Should I use the values in test setup? Seems like its `sa` for username and empty for password.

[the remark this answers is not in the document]
andrewazores: To clarify:

> if the username or password changes then Cryostat can't open the database file for access

When Cryostat creates the `h2:file` database it uses the JDBC user/pass it has at that time. If a later Cryostat instance comes along after a restart etc. then it must have he same JDBC user/pass settings in order to access that same database file.

The JDBC user/pass do not need to be the ones currently specified as defaults in StorageModule or as overrides in smoketest.sh, those can be considered as totally arbitrary.

internal/test/resources.go
tthvo: Right, make sense to remove those accesses if not needed. I can remove the `endpoints` if disabling built-in mechanism. Just a bit unsure what are other permissions used for so I can remove them too. Any lead for this too @ebaron?

internal/controllers/common/resource_definitions/resource_definitions.go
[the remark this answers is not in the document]
andrewazores: For now when they are simply hardcoded, maybe just the CR name as the username or something like that. The password can probably just be empty. There is nothing sensitive in the database file other than the JMX credentials, which are encrypted within the file anyway, so I don't think this is too critical at the moment.

ebaron (commented):
Hey @tthvo looks good! Just a few style comments at first glance.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -739,6 +739,41 @@ func NewCoreContainer(cr *operatorv1beta1.Cryostat, specs *ServiceSpecs, imageTa
 		}
 	}
 
+	disableBuiltInDiscovery := cr.Spec.TargetDiscoveryOptions != nil && cr.Spec.TargetDiscoveryOptions.BuiltInDiscoveryDisabled
+	if disableBuiltInDiscovery {
+		envs = append(envs, corev1.EnvVar{
+			Name:  "CRYOSTAT_DISABLE_BUILTIN_DISCOVERY",
+			Value: fmt.Sprintf("%v", disableBuiltInDiscovery),
```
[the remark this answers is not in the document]
andrewazores: As mentioned elsewhere, if this env var is set with any non-empty value then it causes the built-ins to be disabled.

I would suggest actually hardcoding a value like `"true"` since this is wrapped in `if disableBuiltInDiscovery`. The way it's written now looks like an easy refactoring for someone to come "clean up" later on by removing the `if`, which would actually break the behaviour and always disable built-ins.

internal/controllers/common/resource_definitions/resource_definitions.go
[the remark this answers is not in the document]
tthvo: Okayy that makes sense! I could use the same value for both.

internal/controllers/common/resource_definitions/resource_definitions.go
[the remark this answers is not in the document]
tthvo: Right, I could set it as `strconv.FormatBool(true)`, then if the way this var is checked changed later, we could replace with variable `disableBuiltInDiscovery `.

internal/test/resources.go
tthvo: Right...For now (or this PR), I think we can avoid modifying roles :D

tthvo:
With the latest commits, I made the following changes: 
- The credentials database password must be configured via a spec field to point to an existing secret. The secret must have the key `CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD` with password. This secret (or specifically the password) must not be editted.
- Operator only accesses the key `CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD` in the secret instead of reading every keys.
- For test, I add a test secret for all test cases since it is always required for Cryostat to work.

internal/controllers/secrets.go
ebaron: Perhaps we should still generate the secret if not provided. Adding a new mandatory user-supplied field to the CRD would break the upgrade process.

tthvo: Oh yes I was concerned about that. Okayyy let me add back the secret reconciler for this password.

tthvo: Hmm, this means updates to Cryostat CR must reserve the option to whether specify secret or auto-generated? Switching between them across updates creates a mismatch in password for the database.

ebaron: That likely would not be easy to do. I think it's fine to not allow switching for now. If users are upgrading from 2.1, they'll get the auto-generated password. If they're creating a fresh CR, they can provide their own password.

tthvo:
I added new changes from last time:

- Database Secret Name spec is now optional. However, it must be specified or not at all for the entire lifetime of the Cryostat CR. Switching is not allowed to avoid failed access due to password mismatch.
- The Operator still manages the changes due to switching but failure in Cryostat due to this change is not handled.

docs/config.md
```diff
@@ -269,6 +269,34 @@ spec:
     targetCacheTTL: 10
 ```
 
+### JMX Credentials Database
+
+The Cryostat application must be provided with a password to encrypt saved JMX credentials in database. The user can specify the a secret containing the password entry with key `CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD`. The Cryostat application will use this password to encrypt saved JMX credentials in database.
```
andrewazores, line 274: "the a secret"

docs/config.md
```diff
@@ -269,6 +269,34 @@ spec:
...
+apiVersion: operator.cryostat.io/v1beta1
+kind: Cryostat
+metadata:
+  name: cryostat-sample
+spec:
+  jmxCredentialsDatabaseOptions:
+    databaseSecretName: credentials-database-secret
+```
+
+**Note**: If the secret is not provided, a default one is generated for this purpose. However, switching between using provided and generated secret is not allowed to avoid password mismatch that causes the Cryostat application's failure to operator the credentials database.
```
andrewazores, line 299: "failure to access" the credentials database, maybe?

andrewazores:
Other than docs typos above the code changes make sense to me. I'll build and test this out shortly.

docs/config.md
tthvo: ohhh that was a pretty odd type XD let me fix it

andrewazores (commented):
LGTM, will leave it for ebaron to approve

ebaron (commented):
How about this idea to guard against the user breaking their credentials DB by destroying the password?
1. If the CR doesn't have a defined `DatabaseSecretName`, create the default secret with a generated password
2. When creating the deployment, check whether the default secret exists.
    - If it does, then the CR initially had no custom secret and the credentials DB would have been created with the generated password.
    - If it doesn't, then check if `DatabaseSecretName` has been set (it should), and use that.
    - If for some reason the `DatabaseSecretName` is not set, then the password has been lost. Likely due to the user deleting the default secret. We could emit an Event here.

ebaron:
> How about this idea to guard against the user breaking their credentials DB by destroying the password?
> 
>     1. If the CR doesn't have a defined `DatabaseSecretName`, create the default secret with a generated password
> 
>     2. When creating the deployment, check whether the default secret exists.
>        
>        * If it does, then the CR initially had no custom secret and the credentials DB would have been created with the generated password.
>        * If it doesn't, then check if `DatabaseSecretName` has been set (it should), and use that.
>        * If for some reason the `DatabaseSecretName` is not set, then the password has been lost. Likely due to the user deleting the default secret. We could emit an Event here.

@tthvo This can be in a follow-up PR I think.

tthvo:
> 

Right, thats a good idea. This means we don't delete the default secret if switching from default to provided secret option so that we can check when creating the deployment?

ebaron:
> 

Right, it may be a good idea as a temporary measure to just not delete the generated secret. I don't think that will ever result in a working deployment. At least with it still around, users can revert their changes.

tthvo:
Just a question: What about the case where the user switch from custom secret to default one? The default secret will take place and still make Cryostat fail right?

tthvo:
I opened the issue for the follow up at #475 :D

ebaron:
> Just a question: What about the case where the user switch from custom secret to default one? The default secret will take place and still make Cryostat fail right?

True, that would be harder to detect. How about this idea instead? We add a `Status.DatabaseSecret` to the CRD. We write to this field once, which will be the canonical database secret for the lifetime of that Cryostat. We could do this by comparing against the existing value of the field and only setting it when nil.

tthvo:
Right that would be a better safe guarding against it. Would you prefer in a separate PR later or in this PR too?

internal/controllers/common/resource_definitions/resource_definitions.go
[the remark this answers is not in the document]
ebaron: This currently reads `fmt.Sprintf("%v", strconv.FormatBool(true))`. The `Sprintf` is unnecessary in this case because `strconv.FormatBool(true)` returns the string `"true"`. I'm with Andrew on this though, since this is always true, there's no need to call a function, `Value: "true"` makes most sense.

ebaron:
> Right that would be a better safe guarding against it. Would you prefer in a separate PR later or in this PR too?

Separate PR is fine. That would actually be the fix for #475.

internal/controllers/common/resource_definitions/resource_definitions.go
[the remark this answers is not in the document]
tthvo: oh right! I will remove it and just use `true` literal.

tthvo:
@ebaron When I run `go mod tidy`, the module `require github.com/google/go-cmp v0.5.6` seems to changed to be a direct required instead of indirect. Is this change desired?

```diff
diff --git a/go.mod b/go.mod
index 80dfb7f..1695415 100644
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,8 @@ require (
        sigs.k8s.io/controller-runtime v0.12.1
 )
        sigs.k8s.io/controller-runtime v0.12.1
 )
 
+require github.com/google/go-cmp v0.5.6
+
 require (
        cloud.google.com/go v0.90.0 // indirect
        github.com/Azure/go-autorest v14.2.0+incompatible // indirect
@@ -41,7 +43,6 @@ require (
        github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
        github.com/golang/protobuf v1.5.2 // indirect
        github.com/google/gnostic v0.5.7-v3refs // indirect
-       github.com/google/go-cmp v0.5.6 // indirect
        github.com/google/gofuzz v1.2.0 // indirect
```

ebaron:
> @ebaron When I run `go mod tidy`, the module `require github.com/google/go-cmp v0.5.6` seems to changed to be a direct required instead of indirect. Is this change desired?
> 
> ```diff
> diff --git a/go.mod b/go.mod
> index 80dfb7f..1695415 100644
> --- a/go.mod
> +++ b/go.mod
> @@ -15,6 +15,8 @@ require (
>         sigs.k8s.io/controller-runtime v0.12.1
>  )
>         sigs.k8s.io/controller-runtime v0.12.1
>  )
>  
> +require github.com/google/go-cmp v0.5.6
> +
>  require (
>         cloud.google.com/go v0.90.0 // indirect
>         github.com/Azure/go-autorest v14.2.0+incompatible // indirect
> @@ -41,7 +43,6 @@ require (
>         github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
>         github.com/golang/protobuf v1.5.2 // indirect
>         github.com/google/gnostic v0.5.7-v3refs // indirect
> -       github.com/google/go-cmp v0.5.6 // indirect
>         github.com/google/gofuzz v1.2.0 // indirect
> ```

Oh, that must have been caused by #433 where I used `cmp` to compare selectors. This change is wanted, but I would add it into the existing `require ( ... ) block`.

ebaron (approved):
Everything seems to work as expected. Nice work @tthvo!

PR #478, opened 2022-10-19
docs(resources): document default resource requests

tthvo:
Related to #437 

Updated config docs for default resource requests. And fix minor typos.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -156,8 +156,8 @@ If the sidecar's resource requests are not specified, they are set with the foll
 
 | Request | Quantity |
 |---------|----------|
-| defaultReportCpuRequest | 128m |
-| defaultReportMemoryRequest | 256Mi |
+| Reports Container CPU | 128m |
+| Reports Container Memory | 256Mi |
 
 
 ### Resource Requirements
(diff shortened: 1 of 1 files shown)


tthvo:
Sorry, just realized I haven't updated docs after #437, so I did it now.

docs/config.md
```diff
@@ -188,6 +188,8 @@ spec:
 ```
 This example sets CPU and memory requests and limits for each container, but you may choose to define any combination of requests and limits that suits your use case.
 
+Note that if you define limits lower than default requests, the limits will be used in place of requests instead.
```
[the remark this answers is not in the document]
tthvo: Oh yeh this is clearer to understand :D I will fix this.

docs/config.md
```diff
@@ -152,8 +152,27 @@ spec:
     subProcessMaxHeapSize: 200
 ```
 
+If the sidecar's resource requests are not specified, they are set with the following defaults:
+
+| Request | Quantity |
+|---------|----------|
+| defaultReportCpuRequest | 128m |
```
ebaron, line 159: I think it's probably better to specify this in plain English instead of using the constant names. Something like "Reports Container CPU/Memory".

docs/config.md
```diff
@@ -152,8 +152,27 @@ spec:
...
+By default, the operator deploys Cryostat with pre-configured resource requests:
+
+| Request | Quantity |
+|---------|----------|
+| defaultCoreCpuRequest | 100m |
+| defaultCoreMemoryRequest | 384Mi |
+| defaultJfrDatasourceCpuRequest | 100m |
+|	defaultJfrDatasourceMemoryRequest | 512Mi |
+|	defaultGrafanaCpuRequest | 1000m |
+|	defaultGrafanaMemoryRequest | 256Mi |
```
ebaron, line 173: Same thing here. "Cryostat container CPU/memory", "JFR Data Source container CPU/memory", "Grafana container CPU/memory".

docs/config.md
tthvo: Right! I will change that!

ebaron (approved):
Looks good!

PR #480, opened 2022-10-25
chore(versions): bump image versions for 2.2.0 release point

andrewazores:
Related to cryostatio/cryostat#1118


andrewazores:
Here and in https://github.com/cryostatio/cryostat-helm/pull/42 I've used `(v)2.2` tags, leaving off the bugfix version. Right now those would be manual tags on quay.io pointing to the same image as `(v)2.2.0(-beta1)` etc., but I think it would be even better if we developed some automation to move an `x.y` tag to any more recent `x.y.z-a` images across our various deliverables. Something else to look into separately.

ebaron (commented):
Hey Andrew, sorry this PR slipped out of my view. Should we use this PR to bump versions post-release?

andrewazores:
I could regear it for that, or else I can just close it and one of us can open a new one to do that.

ebaron:
> I could regear it for that, or else I can just close it and one of us can open a new one to do that.

I'll open a new PR for this. Easier than trying to rebase.

Issue #481, opened 2022-10-31
CSV description points to outdated documentation


ebaron:
The CSV description directs users to the 2.0.0 documentation for Network Options. This has changed somewhat due to the deprecation of the command Ingress and the addition of using this field for route labels/annotations.

PR #482, opened 2022-10-31
build(csv): update link in CSV description

ebaron:
This link doesn't exist yet, but will when 2.2.0 is released. I also updated the date as well.

Fixes: #481


tthvo (approved):
Nice, looks good :D

Issue #454, opened 2022-09-15
Create sample apps from deployment/service manifests


ebaron:
>     1. Our sample apps seem not to deployed in `restricted-v2` (got an alert for violation but still allowed). I think the recipe still works because I logged in as kubeadmin. Maybe we could specify some field with `oc new-app` and update demo images?

I'm not sure if there's any parameters we could set for this in `new-app`. Another solution would be to add the deployment and service YAML manifests to our repo, and create them with `$(CLUSTER_CLIENT) create -f`. This would also avoid the dependency on `oc` instead of `kubectl`.

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/issues/450#issuecomment-1248301740_

PR #486, opened 2022-11-09
test(sample): update sample apps

andrewazores:
Related to #485

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -364,8 +364,8 @@ undeploy_sample_app:
 
 .PHONY: sample_app_quarkus
 sample_app_quarkus: undeploy_sample_app_quarkus
-	$(call new-sample-app,quay.io/andrewazores/quarkus-test:0.0.2)
-	$(CLUSTER_CLIENT) patch svc/quarkus-test -p '{"spec":{"$setElementOrder/ports":[{"port":9096},{"port":9999}],"ports":[{"name":"jfr-jmx","port":9096}]}}'
+	$(call new-sample-app,quay.io/andrewazores/quarkus-test:0.0.10)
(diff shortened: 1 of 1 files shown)


ebaron (approved):

tthvo:
Sorry a bit late on this, but just noticed `quarkus-test` is not using JMX auth while checking on Cryostat web console?

andrewazores:
We could add it in there, but it would need another update to the app container to add the `jmxremote.password` and `jmxremote.access` files like what `vertx-fib-demo` has.

tthvo:
Ah right! Are we planning to add them in or Is this quarkus-test meant to have SSL enabled while vertx has Auth for this sample app setup?

andrewazores:
`vertx-fib-demo` can do both auth and SSL, it just needs env vars set to enable them (https://github.com/cryostatio/cryostat/blob/e97ecb1915af7218c3a3d3f9ed784e0520caaba2/smoketest.sh#L82). `quarkus-test` isn't built/configured with support for either of those things, it's just there as a placeholder for a second testing application right now. It can also be configured to act as a discovery plugin though, and it also has an early build of the Cryostat Agent (discovery plugin impl only) baked in, so it's useful for testing those components.

Adding more make recipes to deploy different instances of each app with each of these different configs would be cool.

tthvo:
Right that makes sense! Also, wondering what is keeping #485 open?

> Adding more make recipes to deploy different instances of each app with each of these different configs would be cool.

That would very nice! Just have to see how #454 turns out :D

andrewazores:
#485 is still open because there's only JMX auth on the one sample app after this PR, but there is still not test case for SSL. I think that's important enough to have the issue remain open.

Issue #488, opened 2022-11-15
Scorecard tag in Makefile does not work well with releases


ebaron:
The image tag for the custom scorecard image is as follows:
```
CUSTOM_SCORECARD_VERSION = $(IMAGE_VERSION)00001
```
For development builds this has worked out to 2.2.0-dev00001, for a release build this changes to 2.2.000001, which is less desirable. Scorecard images are not particularly intended to follow the release cycle. I suggest we change this scheme to `MAJOR.MINOR.PATCH-TIMESTAMP`. This makes it clear which release the scorecard is intended to test, and also the age of the build.

PR #489, opened 2022-11-15
build(scorecard): use date for scorecard tag

ebaron:
Fixes: #488


Makefile
```diff
@@ -82,7 +82,9 @@ ENVTEST_K8S_VERSION ?= 1.24
 
 # Scorecard ImagePullPolicy is hardcoded to IfNotPresent
 # See: https://github.com/operator-framework/operator-sdk/pull/4762
-CUSTOM_SCORECARD_VERSION = $(IMAGE_VERSION)00001
+#
+# Suffix is the timestamp of the image build, compute with: date -u '+%Y%m%d%H%M%S'
+CUSTOM_SCORECARD_VERSION ?= 2.2.0-20220829183117
```
andrewazores, line 87: Could this value be generated within `scorecard-build`? Nothing else seems to use that as a dependency, so I think it would only happen when the developer is intentionally building a new scorecard test image, and this way it isn't just a hardcoded value that needs manuak updating along with the rebuild.

I guess this would also mean the scorecard config patch needs to be updated with something like a `hack/scorecard-patch.yaml.in` too.

tthvo (commented):
Looks good! Just the same question with Andrew above :D

Makefile
ebaron: How does this revision look? The version will be generated at build time, and the scorecard Kustomize patch will only get updated when the user runs `make scorecard-build`.

andrewazores: Nice, I think that should work well enough.

tthvo (approved):

PR #495, opened 2022-11-16
build(version): update versions for 2.2.0

ebaron:


tthvo (approved):

tthvo:
From Andrew's comment for #496, looks like we might have missed too?

https://github.com/cryostatio/cryostat-operator/blob/978e04a8b98cd4fa4e4eb843b1dd196b3d666271/config/manifests/bases/cryostat-operator.clusterserviceversion.yaml#L460

ebaron:
> From Andrew's comment for #496, looks like we might have missed too? Seems to not affect the bundle anyways. Just in base manifests. A little odd?
> 
> https://github.com/cryostatio/cryostat-operator/blob/978e04a8b98cd4fa4e4eb843b1dd196b3d666271/config/manifests/bases/cryostat-operator.clusterserviceversion.yaml#L460

I've changed this to a placeholder in `main`, so we won't have to manually update it anymore. The generated bundle has the correct version regardless, so no harm done.

tthvo:
Nice thanks for explaining that!

PR #496, opened 2022-11-16
build(version): bump versions for 2.3.0-dev

ebaron:

### DIFF
--- a/config/manifests/bases/cryostat-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/cryostat-operator.clusterserviceversion.yaml
@@ -18,8 +18,8 @@ metadata:
       ]
     capabilities: Seamless Upgrades
     categories: Monitoring, Developer Tools
-    containerImage: quay.io/cryostat/cryostat-operator:2.2.0-dev
-    createdAt: "2022-10-31 00:00:00"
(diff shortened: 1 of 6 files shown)


config/manifests/bases/cryostat-operator.clusterserviceversion.yaml
```diff
@@ -18,8 +18,8 @@ metadata:
       ]
     capabilities: Seamless Upgrades
     categories: Monitoring, Developer Tools
-    containerImage: quay.io/cryostat/cryostat-operator:2.2.0-dev
-    createdAt: "2022-10-31 00:00:00"
+    containerImage: quay.io/cryostat/cryostat-operator:2.3.0-dev
+    createdAt: "2022-11-16 00:00:00"
```
andrewazores, line 22: https://github.com/cryostatio/cryostat-operator/blob/4fde1a073e5abf9336532c156d1e0ddaea245b8b/config/manifests/bases/cryostat-operator.clusterserviceversion.yaml#L460

Does this need bumping too?

ebaron: Good catch. This version gets replaced with `make bundle`, so I've changed it to a 0.0.0 placeholder. I ran `make bundle` again and the version is set as expected.

tthvo (approved):
Looks good :D

PR #498, opened 2022-11-17
build(version): update versions for 2.2.1-dev

ebaron:


andrewazores (commented):
lgtm

tthvo (approved):
Looks good to me! Just wondering about the choice of image tags here? Why `snapshot` and `cryostat-` (only Grafana?)?

tthvo:
Right that makes sense! Thanks for explaining :D

ebaron:
> Right that makes sense! Thanks for explaining :D

No problem, I filed https://github.com/cryostatio/cryostat-grafana-dashboard/issues/19 in response to this discussion.

Issue #491, opened 2022-11-15
Default PVC can cause rolling updates to fail


ebaron:
Depending on the storage implementation, you may encounter errors of the form: `Multi-Attach error for volume "..." Volume is already used by pod(s) ...` when trying to do a rolling update. This occurs when the PV doesn't support `ReadWriteMany`, and the replicas end up on different nodes [1].

[1] https://stackoverflow.com/a/63524679

PR #499, opened 2022-11-21
fix(deploy): use recreate strategy and 1 replica

ebaron:
There are two bugs that occur due to replicas contending for access to the credentials database, and replicas on different nodes competing for the PVC itself. Cryostat is currently meant to work as a single replica, but during rolling updates there could be two simultaneously.

To avoid these issues, we're changing the deployment strategy from the default of `RollingUpdate` to `Recreate`. As a result there will be some brief downtime in response to upgrades or configuration changes. The other change is the operator managing the `Replicas` field of the main Cryostat deployment. This is to ensure that the replica count stays at 1 to avoid the above bugs.

To test:
1. `make deploy create_cryostat_cr OPERATOR_IMG=....`
2. `kubectl edit cryostat cryostat-sample`. Change some field such as setting `reportOptions.replicas` to 1.
3. The Cryostat deployment should scale down to 0 and then back up to 1, reflecting this new configuration.
4. Try manually scaling up/down the Cryostat deployment. After a brief time, the operator should revert the deployment to 1 replica.

Fixes: #487, #491

### DIFF
--- a/internal/controllers/cryostat_controller_test.go
+++ b/internal/controllers/cryostat_controller_test.go
@@ -309,7 +309,8 @@ var _ = Describe("CryostatController", func() {
 
 				// Deployment Selector is immutable
 				Expect(deploy.Spec.Selector).To(Equal(oldDeploy.Spec.Selector))
-				Expect(deploy.Spec.Replicas).To(Equal(oldDeploy.Spec.Replicas))
+				Expect(deploy.Spec.Replicas).To(Equal(&[]int32{1}[0]))
+				Expect(deploy.Spec.Strategy).To(Equal(test.NewMainDeploymentStrategy()))
 			})
 			Context("with a different selector", func() {
 				BeforeEach(func() {
@@ -2578,6 +2579,9 @@ func (t *cryostatTestInput) checkMainDeployment() {
 	}))
 	Expect(metav1.IsControlledBy(deployment, cr)).To(BeTrue())
 	Expect(deployment.Spec.Selector).To(Equal(test.NewMainDeploymentSelector()))
+	Expect(deployment.Spec.Replicas).ToNot(BeNil())
+	Expect(*deployment.Spec.Replicas).To(Equal(int32(1)))
+	Expect(deployment.Spec.Strategy).To(Equal(test.NewMainDeploymentStrategy()))
 
 	// compare Pod template
 	t.checkMainPodTemplate(deployment, cr)
@@ -2669,7 +2673,9 @@ func (t *cryostatTestInput) checkReportsDeployment() {
 	}))
 	Expect(metav1.IsControlledBy(deployment, cr)).To(BeTrue())
 	Expect(deployment.Spec.Selector).To(Equal(test.NewReportsDeploymentSelector()))
+	Expect(deployment.Spec.Replicas).ToNot(BeNil())
 	Expect(*deployment.Spec.Replicas).To(Equal(t.reportReplicas))
+	Expect(deployment.Spec.Strategy).To(BeZero())
 
 	// compare Pod template
 	template := deployment.Spec.Template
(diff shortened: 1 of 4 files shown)


internal/controllers/cryostat_controller_test.go
```diff
@@ -2669,7 +2673,9 @@ func (t *cryostatTestInput) checkReportsDeployment() {
 	}))
 	Expect(metav1.IsControlledBy(deployment, cr)).To(BeTrue())
 	Expect(deployment.Spec.Selector).To(Equal(test.NewReportsDeploymentSelector()))
+	Expect(deployment.Spec.Replicas).ToNot(BeNil())
 	Expect(*deployment.Spec.Replicas).To(Equal(t.reportReplicas))
+	Expect(deployment.Spec.Strategy).To(BeZero())
```
tthvo, line 2678: Not sure if I asked somewhere already, but is this fakeClient limitations because I suppose it would default to ''rollingUpdate" on actual cluster?

tthvo:
> kubectl edit cryostat cryostat-sample. Change some field such as setting reportOptions.replicas to 1.

About this, I am bit unsure what causes the main deployment to rollout new pods if this does not change its spec? Setting `reportOptions.replicas` just change the report deployment, right?

internal/controllers/cryostat_controller_test.go
ebaron: Right, if left unset on a real cluster, Kubernetes will populate the Strategy field as `RollingUpdate`: https://github.com/kubernetes/kubernetes/blob/520b991347b9c77635c0e4555f1703a86dcdd4ff/pkg/apis/apps/v1/defaults.go#L46-L64

tthvo:
I also got a bit of an issue when creating a report replica as pod seems to in `CrashLoopBackOff` with logs show:

```bash
exec /deployments/run-java.sh: permission denied
```

ebaron:
> > kubectl edit cryostat cryostat-sample. Change some field such as setting reportOptions.replicas to 1.
> 
> About this, I am bit unsure what causes the main deployment to rollout new pods if this does not change its spec? Setting `reportOptions.replicas` just change the report deployment, right?

This ends up setting the `CRYOSTAT_REPORT_GENERATOR` environment variable with the URL to the reports service: https://github.com/cryostatio/cryostat-operator/blob/23cfadd771a85cf681ab9790f6aba86a8efac14b/internal/controllers/common/resource_definitions/resource_definitions.go#L661-L668.

So there is a subtle config change on the main deployment as well.

ebaron:
> I also got a bit of an issue when creating a report replica as pod seems to in `CrashLoopBackOff` with logs show:
> 
> ```shell
> exec /deployments/run-java.sh: permission denied
> ```
> 
> I am using Cluster Bot with OCP 4.11.1

Oh this again. We've encountered this with some of our sample apps. It's a matter of updating our Dockerfiles. You should still be able to test this PR though. Instead of the reports replicas, you could try switching `enableCertManager`.

tthvo:
> > > kubectl edit cryostat cryostat-sample. Change some field such as setting reportOptions.replicas to 1.
> > 
> > 
> > About this, I am bit unsure what causes the main deployment to rollout new pods if this does not change its spec? Setting `reportOptions.replicas` just change the report deployment, right?
> 
> This ends up setting the `CRYOSTAT_REPORT_GENERATOR` environment variable with the URL to the reports service:
> 
> https://github.com/cryostatio/cryostat-operator/blob/23cfadd771a85cf681ab9790f6aba86a8efac14b/internal/controllers/common/resource_definitions/resource_definitions.go#L661-L668
> .
> 
> So there is a subtle config change on the main deployment as well.

Ah niceee! I didnt notice before. Make sense! Thanks :D

internal/controllers/cryostat_controller_test.go
tthvo: Right! Make sense! Thanks :D

tthvo (approved):
> Oh this again. We've encountered this with some of our sample apps. It's a matter of updating our Dockerfiles. You should still be able to test this PR though. Instead of the reports replicas, you could try switching enableCertManager.

Nice, so its unrelated then. And looks great to me and deployments are recreated :D

Issue #506, opened 2022-12-09
Tests must support multiple namespaces


ebaron:
Currently, our test objects have a hardcoded namespace. In order to add tests for more than one Cryostat, we should make the namespace of these test objects configurable.

PR #507, opened 2022-12-09
test(resources): multi-namespace support for test resources

ebaron:
This PR is long, but a lot of the changes are a result of changing functions to methods. I've added a `TestResources` struct that is embedded into the `cryostatTestInput` used by the controller tests. Several fields have moved to the `TestResources` type where it makes sense: `minimal`, `tls`, etc. These would be fields that have some effect on the creation of the test objects. In addition to this, a new `namespace` field is added. Modifying this field affects the namespace of all subsequently created test objects.

By changing the test object creation from functions to methods, I've been able to reduce the parameter bloat, which hopefully results in more readable tests. I've tried quite a few iterations of this refactoring, before settling on this approach. Let me know what you think.

Fixes: #506

### DIFF
--- a/internal/controllers/cryostat_controller_test.go
+++ b/internal/controllers/cryostat_controller_test.go
@@ -419,9 +419,10 @@ var _ = Describe("CryostatController", func() {
 			It("should update the Cluster Role Binding", func() {
 				t.reconcileCryostatFully()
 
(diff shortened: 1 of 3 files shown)


andrewazores (approved):

tthvo (approved):
Looks good to me :D

Issue #509, opened 2022-12-21
Reconciling in tests should be more flexible


ebaron:
Currently the `reconcileCryostatFully` method depends on the strict ordering of how the `Reconcile` method is implemented. This gives it the opportunity to mock the behaviour of cert-manager and the OpenShift router. This can be a problem though if the order of this implementation changes, and doesn't allow for subsequent reconciles to use this method.

PR #510, opened 2022-12-21
test(reconcile): simulate controllers with custom client

ebaron:
This PR moves the logic for updating certificate and route statuses from `reconcileCryostatFully` to a custom client implementation. This client triggers the update upon the first successful `Get` call. This happens after the object has been created, so we're still exercising the `RequeueAfter` code paths.

The advantage here is that we can simplify how we do reconciles in the tests themselves. I've changed `reconcileCryostatFully` to use Gomega's `Eventually` to call `Reconcile` asynchronously until it no longer re-queues or times out. This allows us to use `reconcileCryostatFully` in more places.

I also removed an unnecessary `Get` call for the certificate in `common/tls.go`. We already have the certificate in memory, and this was triggering the certificate update logic earlier than desired.

Fixes: #509

### DIFF
--- a/internal/controllers/certmanager.go
+++ b/internal/controllers/certmanager.go
@@ -123,18 +123,25 @@ func (r *CryostatReconciler) setupTLS(ctx context.Context, cr *operatorv1beta1.C
 		return nil, err
 	}
 
+	// Get the Cryostat CA certificate bytes from certificate secret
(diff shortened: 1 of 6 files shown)


andrewazores (approved):
Looks fine to me.

Issue #525, opened 2023-02-13
Version of `go-restful` is vulnerable to CVE-2022-1996


ebaron:
Until we upgrade to a newer Operator SDK and client-go, we should replace this transitive dependency in go.mod.

PR #526, opened 2023-02-13
build(deps): replace go-restful with newer version

ebaron:
Fixes: #525

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -95,3 +95,6 @@ replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.1
 
 // Fix for CVE-2021-3121
 replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+
+// Fix for CVE-2022-1996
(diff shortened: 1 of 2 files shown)


tthvo:
I am reading from this https://bugzilla.redhat.com/show_bug.cgi?id=2094982 (right source?)

> Authorization Bypass Through User-Controlled Key in GitHub repository emicklei/go-restful prior to v3.8.0.

Does that mean also `2.16.0` is affected or is it already back-ported for `2.16.0`?

ebaron:
> I am reading from this https://bugzilla.redhat.com/show_bug.cgi?id=2094982 (right source?)
> 
> > Authorization Bypass Through User-Controlled Key in GitHub repository emicklei/go-restful prior to v3.8.0.
> 
> Does that mean also `2.16.0` is affected or is it already back-ported for `2.16.0`?

It was also backported for 2.16.0, as shown here: https://nvd.nist.gov/vuln/detail/CVE-2022-1996
Doesn't seem like this is mentioned on the RHBZ bug though.

tthvo (approved):
Looks good to me! Thanks for explaining^^

Issue #529, opened 2023-02-23
Make cert-manager/OpenShift checks more lenient


ebaron:
Sometimes we may encounter an issue with a particular APIResource when gathering them all with `DiscoveryClient.ServerGroupsAndResources`, such as:
`error: unable to retrieve the complete list of server APIs: example.com/v1: the server is currently unable to handle the request`.

We should make the operator more resilient against these unrelated APIResource failures.

PR #530, opened 2023-02-23
fix(discovery): look up API resources individually

ebaron:
This changes the discovery approach from looking up all APIResources to just looking up those we need to detect OpenShift and cert-manager.

Fixes: #529


tthvo (approved):
Looks good to me!! Worked very well :D

PR #543, opened 2023-03-28
build(operator-sdk): upgrade operator sdk version to 1.28.0

### LINKED ISSUE #455: Label the scorecard namespace to enforce restricted PSA profile
>     2. For our custom scorecard tests, we create a namespace. I wonder if we should also test in a namespace which enforce `restricted` PSA? (not sure I understand the context right)

That seems like a good idea. We should do that once https://github.com/operator-framework/operator-sdk/issues/5939 is fixed.

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/issues/450#issuecomment-1248301740_

### LINKED ISSUE #536: Upgrade Operator SDK to >= 1.28.0
We should upgrade the operator to be compatible with the latest Operator SDK to bring in the latest features and fixes.

Migration guide: https://master.sdk.operatorframework.io/docs/upgrading-sdk-version/

tthvo:
Fixes #536 
Fixes #455 

Upgrade to use operator sdk `1.28.0`. Step by step changes:

References:

- https://master.sdk.operatorframework.io/docs/upgrading-sdk-version/v1.28.0/
- https://github.com/operator-framework/operator-sdk/blob/484013d1865c35df2bc5dfea0ab6ea6b434adefa/testdata/go/v3/memcached-operator/Makefile

### 1.23.0
- [x] Update controller-gen from `0.9.0` to `0.9.2`. Update to use variable `CONTROLLER_TOOLS_VERSION` for controller-gen version (i.e. sync with operator-sdk go template).
- [x] Update controller-runtime from `0.12.1` to `0.12.2` and Kubernetes dependencies from `0.24.0` to `0.24.2`.
- [x] Upgrade kube proxy image gcr.io/kubebuilder/kube-rbac-proxy from `0.11.0` to `0.13.0`. 

### 1.24.0

Nothing to change. There are some `sed` ops on `ARCH` var for arm support but no mentions of Go operators.

### 1.25.0

- [x] Bump k8s runtime dependencies to `0.25.0` and controller-runtime to `0.13.0`. 
  - [x] Upgrade go version to `1.19`.
  - [x] Update test client Get func signature.
- [x] Bump test dependencies (mainly to Gingko v2).
  - [x] Remove custom report (deprecated) for adding newline.
  - [x] Remove `ACK_GINKGO_DEPRECATIONS` as Gingko is upgrade to v2.
- [x] Multi-platform build for operator image. `docker` only and need to install `buildx`.

### 1.26.0

Nothing to change. 

### 1.27.0

Nothing to change. 

### 1.28.0
- [x] Makefile update
  - [x] Bump `CONTROLLER_TOOLS_VERSION` to `0.11.1` (migration guide said `0.11.3` but scaffold specifies at `0.11.1`).
  - [x] Bump `ENVTEST_K8S_VERSION` to `1.26`.
  - [x] Update prerequisites of `manager` target to include `manifests` target. Update image build targets to inlude `fmt` and `vet`.
  - [x] `controller-gen` and `kustomize` is added with a check for version (i.e. download if version is mismatched).
- [x] Bump k8s runtime dependencies to `0.26.2` and controller-runtime to `0.14.5`.
- [x] Bump `kube-rbac-proxy` image to `0.13.1`.
- [x] Set `--pod-security=restricted` when running `operator-sdk scorecard`. This launches scorecard pods with restricted PSA.

### Tests

- Reference bundle image: `quay.io/thvo/cryostat-operator-bundle:latest`.
- Passed scorecards.


Makefile
```diff
@@ -136,7 +135,7 @@ ifneq ($(SKIP_TESTS), true)
 	$(CLUSTER_CLIENT) create namespace $(SCORECARD_NAMESPACE)
 	$(CLUSTER_CLIENT) -n $(SCORECARD_NAMESPACE) create -f internal/images/custom-scorecard-tests/rbac/
 	operator-sdk run bundle -n $(SCORECARD_NAMESPACE) $(BUNDLE_IMG)
-	operator-sdk scorecard -n $(SCORECARD_NAMESPACE) -s cryostat-scorecard -w 5m $(BUNDLE_IMG)
+	operator-sdk scorecard -n $(SCORECARD_NAMESPACE) -s cryostat-scorecard -w 5m $(BUNDLE_IMG) --pod-security=restricted
```
tthvo, line 138: Seems this fixes #455?

Makefile
```diff
@@ -228,9 +227,25 @@ add-license: addlicense
...
+# - have enable BuildKit, More info: https://docs.docker.com/develop/develop-images/build_enhancements/
+# To properly provided solutions that supports more than one platform you should use this option.
+PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
+.PHONY: docker-buildx
+docker-buildx: manifests generate fmt vet test-envtest ## Build OCI image for the manager for cross-platform support
+	# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
+	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
+	- docker buildx create --name project-v3-builder
+	docker buildx use project-v3-builder
+	- docker buildx build --platform=$(PLATFORMS) --tag $(OPERATOR_IMG) -f Dockerfile.cross .
```
andrewazores, line 246:
```bash
$ podman buildx --help
Build images

Description:
  Build images

Usage:
  podman buildx [command]

Aliases:
  buildx, builder

Available Commands:
  build       Build an image using instructions from Containerfiles
  prune       Remove unused images
```

There is a `podman buildx` for cross-compilation too - is it not sufficient for this case? I'm not sure what this `create` and `use` is for, but building for a different platform with a specific tag on a given Dockerfile should all work.

tthvo: yehh those `create` and `use` is the headache cuz podman does not support that. But from reading this doc: https://docs.docker.com/engine/reference/commandline/buildx_create/,

> All Docker contexts also get the default builder instance.

I think we can just remove them to be compatible with `podman`?

Makefile
```diff
@@ -228,9 +227,25 @@ add-license: addlicense
...
+	- docker buildx rm project-v3-builder
+	rm Dockerfile.cross
```
ebaron, line 248: So far, I've just tested this portion. I had to make some edits to bring it in line with what's shown here:
https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.25.0/#kubebuilder-bump-brought-changes-that-modified-the-makefile-scaffolding-and-also-multiarch-changes-are-also-included

Basically, we need to add the `--push` argument and also the final bullet point in the comments about needing push access. I wasn't able to make it work locally without pushing for some reason. Here's the resulting image:
https://quay.io/repository/ebaron/cryostat-operator/manifest/sha256:5aaba17aa8cd2deaa80143f9c39cba5e00942f8d17b965baca7ef99c94884039

Once we start publishing multi-arch images, I think it would be good to add a podman/buildah equivalent like this:
https://sdk.operatorframework.io/docs/advanced-topics/multi-arch/
Having the Docker option is good for completeness though.

Makefile
ebaron: Looks like it!

Makefile
andrewazores: I think so. Worth trying it out, anyway. In my experience with doing cross-platform builds with podman, all I needed to do was use `--platform` and install the `qemu-user-static` RPM package. I didn't need to create any temporary build contexts etc.

Makefile
ebaron: I tried with `-o type=image` without `--push`, I could not find the resulting image/manifest list under `docker images`. I'm not sure why.

ebaron:
> For multi-arch build, it needs official `docker` tool with buildx plugin to build. Took me a while to realize the requirement for official docker includes `Fedora 36`, but I am still on `Fedora 35` and using `moby-engine`. Not sure there are enough time to upgrade to test. Could you help check it out instead?
> 
> Also, wondering if we need to add multi-arch build for scorecard images?

I was using `moby-engine` with the buildx plugin installed into `~/.docker/cli-plugins/`: https://github.com/docker/buildx#manual-download

Makefile
```diff
@@ -254,20 +269,25 @@ LOCALBIN ?= $(shell pwd)/bin
...
 	mkdir -p $(LOCALBIN)
 
-# Download controller-gen locally if necessary
+# Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
 CONTROLLER_GEN = $(LOCALBIN)/controller-gen
 .PHONY: controller-gen
 controller-gen: $(CONTROLLER_GEN)
 $(CONTROLLER_GEN): $(LOCALBIN)
-	test -s $(CONTROLLER_GEN) || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@v$(CONTROLLER_GEN_VERSION)
+	test -s $(CONTROLLER_GEN) && $(CONTROLLER_GEN) --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
```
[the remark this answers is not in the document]
ebaron: I think what's happening is that the `$(LOCALBIN)` directory has the same modification time as the most recently updated binary within it. Because `$(LOCALBIN)` is used as a prerequisite for targets like controller-gen, Make sees that the prerequisite is not newer than the target, so there's nothing to do.

One fix could be to use a phony target for `$(LOCALBIN)` and not use the actual directory as a prerequisite anywhere. This would ensure the directory exists, but not considering its modification time when running dependent targets.

Something like this:
```patch
diff --git a/Makefile b/Makefile
index 4ae65c9..788c668 100644
--- a/Makefile
+++ b/Makefile
@@ -266,14 +266,15 @@ check_cert_manager:
 
 # Location to install dependencies
 LOCALBIN ?= $(shell pwd)/bin
-$(LOCALBIN):
-	mkdir -p $(LOCALBIN)
+.PHONY: local-bin
+local-bin:
+	test -d $(LOCALBIN) || mkdir -p $(LOCALBIN)
 
 # Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
 CONTROLLER_GEN = $(LOCALBIN)/controller-gen
 .PHONY: controller-gen
 controller-gen: $(CONTROLLER_GEN)
-$(CONTROLLER_GEN): $(LOCALBIN)
+$(CONTROLLER_GEN): local-bin
 	test -s $(CONTROLLER_GEN) && $(CONTROLLER_GEN) --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
 	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@v$(CONTROLLER_TOOLS_VERSION)
 
@@ -282,7 +283,7 @@ KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/k
 KUSTOMIZE = $(LOCALBIN)/kustomize
 .PHONY: kustomize
 kustomize: $(KUSTOMIZE)
-$(KUSTOMIZE): $(LOCALBIN)
+$(KUSTOMIZE): local-bin
 	@if test -x $(LOCALBIN)/kustomize && ! $(LOCALBIN)/kustomize version | grep -q $(KUSTOMIZE_VERSION); then \
 		echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
 		rm -rf $(LOCALBIN)/kustomize; \
@@ -293,21 +294,21 @@ $(KUSTOMIZE): $(LOCALBIN)
 ADDLICENSE = $(LOCALBIN)/addlicense
 .PHONY: addlicense
 addlicense: $(ADDLICENSE)
-$(ADDLICENSE): $(LOCALBIN)
+$(ADDLICENSE): local-bin
 	test -s $(ADDLICENSE) || GOBIN=$(LOCALBIN) go install github.com/google/addlicense@v$(ADDLICENSE_VERSION)
 
 # Download setup-envtest locally if necessary
 ENVTEST = $(LOCALBIN)/setup-envtest
 .PHONY: setup-envtest
 setup-envtest: $(ENVTEST)
-$(ENVTEST): $(LOCALBIN)
+$(ENVTEST): local-bin
 	test -s $(ENVTEST) || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
 # Download opm locally if necessary
 OPM = $(LOCALBIN)/opm
 .PHONY: opm
 opm: $(OPM)
-$(OPM): $(LOCALBIN)
+$(OPM): local-bin
 	test -s $(OPM) || \
 	{ \
 	set -e ;\
```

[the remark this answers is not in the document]
ebaron: The `test -d` can be removed since we already use `mkdir -p`.

Makefile
tthvo: I found this: https://github.com/docker/roadmap/issues/371. Seems like its not possible yet (available in some beta version tho). I will add `--push` back then.

tthvo:
> I was using moby-engine with the buildx plugin installed into ~/.docker/cli-plugins/: https://github.com/docker/buildx#manual-download

Oh thanks! I will try that (saw the word `not recommend` last time so I quickly skipped it :D ).

Makefile
ebaron: Good find!

tthvo:
We also need to specify supported OS and ARCH in CSV? https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/

README.md
```diff
@@ -49,7 +49,7 @@ kubectl get secret ${CRYOSTAT_NAME}-jmx-auth -o jsonpath='{$.data.CRYOSTAT_RJMX_
 
 # Building
 ## Requirements
-- `go` v1.18
+- `go` v1.19
 - [`operator-sdk`](https://github.com/operator-framework/operator-sdk) v1.22.2
```
ebaron, line 53: This should be updated to v1.28.0

andrewazores:
> EDIT: Nice got it to use buildx with moby engine now. Thanks alot!

Any way around the `create`/`use` temporary build contexts? If so then this would also be podman compatible, which would be nice...

ebaron:
> We also need to specify supported OS and ARCH in CSV? https://olm.operatorframework.io/docs/advanced-tasks/ship-operator-supporting-multiarch/

We do yes, but this may be a bit premature until the operator and all the operand images have published multi-arch images.

Makefile
tthvo: I am seeing this without switching to another builder:

```
ERROR: multiple platforms feature is currently not supported for docker driver. 
Please switch to a different driver (eg. "docker buildx create --use")
make: *** [Makefile:245: oci-buildx] Error 1
```
We really need to create one then.

ebaron:
> How about split cases for docker and podman so we can specified command separately?

That seems fine to me, just wrap the extra Docker stuff in `ifeq($(IMAGE_BUILDER),docker)`

Dockerfile
```diff
@@ -15,7 +17,11 @@ COPY api/ api/
...
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager internal/main.go
+# the GOARCH has not a default value to allow the binary be built according to the host where the command
+# was called. For example, if we call make oci-build in a local env which has the Apple Silicon M1 SO
+# the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
+# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} GO111MODULE=on go build -a -o manager internal/main.go
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
tthvo, line 26: Without specifying any platform, it should automatically assume ones from `--platform` passed to docker/podman build right?

andrewazores: That's what seemed to happen when I was working on https://github.com/cryostatio/cryostat/pull/1352 . The correct base image got automatically pulled and built on top of, both when I did it manually with podman as well as when I did it through the jib plugin.

tthvo: Right thanks! No need any updates here then.

ebaron: I was able to confirm this by adding `-o type=tar,dest=./test.tar` to the `docker buildx` command, then extracting the manager binary from the tarball:
```
$ file manager 
manager: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, Go BuildID=s9qHdPIc1o_3I_4lPFO_/Y6beLdq3XQXAY2UZSJzf/gltfY-CbCYZBYIrTAmmM/E5RDVvqMGdi8cBTOUIyX, with debug_info, not stripped
```

tthvo: Niceee! Thanks again!

tthvo:
I am having a bit of issue with `oc-build`, where the command 

```bash
[1/2] STEP 11/11: RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} GO111MODULE=on go build -a -o manager internal/main.go
```

takes forever to run. Wondering if its just my setup or the same for everyone?

ebaron:
> I am having a bit of issue with `oc-build`, where the command
> 
> ```shell
> [1/2] STEP 11/11: RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} GO111MODULE=on go build -a -o manager internal/main.go
> ```
> 
> takes forever to run. Tho, it eventually finishes (happens on first build after bumping go and deps). Wondering if its just my setup or the same for everyone? Or was it because builder container was cached?

It doesn't take long for me. Not sure why there's a difference.

andrewazores:
I find that takes quite a while too, and subsequent runs are a lot faster after that container layer is cached. `podman image prune -f` deletes it because the intermediate is not tagged, and then the next invocation takes a long time again.

That's the case on `main` though, not just for this PR (which I haven't tested yet)

tthvo:
Oh yeh exactly! I have been running prune after every build. That makes sense thanks a lot!

tthvo:
Added the handing for podman case. Seems to build as expected (only included 2 amd64 and arm64, the last 2 takes so long..):

https://quay.io/repository/thvo/cryostat-operator/manifest/sha256:bdda2f96fd1d1c885616ea8eef18a724a8dcf1e73ef8ff5ac799231434d9dc65

Makefile
```diff
@@ -228,24 +228,36 @@ add-license: addlicense
...
 	docker buildx use project-v3-builder
-	- docker buildx build --platform=$(PLATFORMS) --tag $(OPERATOR_IMG) -f Dockerfile.cross .
+	- docker buildx build --push --platform=$(PLATFORMS) --tag $(OPERATOR_IMG) -f Dockerfile.cross .
 	- docker buildx rm project-v3-builder
 	rm Dockerfile.cross
+else
+	for platform in $$(echo $(PLATFORMS) | sed "s/,/ /g"); do \
+		os=$$(echo $${platform} | cut -d/ -f 1); \
+		arch=$$(echo $${platform} | cut -d/ -f 2); \
+		BUILDAH_FORMAT=docker podman buildx build --manifest $(OPERATOR_IMG) --platform $${platform} --build-arg TARGETOS=$${os} --build-arg TARGETARCH=$${arch} . ; \
```
ebaron, line 257: Do we need to pass the `TARGETOS` and `TARGETARCH` to `podman buildx`, or does it work implicitly like the `docker buildx` case?

tthvo: Seems like it does not for some reasons. I receive warning about missing `TARGETOS` and `TARGETARCH` if we don't. Do you see the same issue?

tthvo: https://sdk.operatorframework.io/docs/advanced-topics/multi-arch/

At the bottom, it said:

> Caveats: the Dockerfile generated by the SDK for the operator explicitly references GOARCH=amd64 for go build. This can be amended to GOARCH=$TARGETARCH. Docker will automatically set the environment variable to the value specified by –platform. With buildah –build-arg will need to be used for the purpose.

ebaron: Thanks for clearing that up for me

Makefile
```diff
@@ -228,9 +227,37 @@ add-license: addlicense
...
+oci-buildx: # manifests generate fmt vet test-envtest ## Build OCI image for the manager for cross-platform support
+ifeq ($(IMAGE_BUILDER), docker)
+# copy existing Dockerfile and insert --platform=${BUILDPLATFORM} into Dockerfile.cross, and preserve the original Dockerfile
+	sed -e '1 s/\(^FROM\)/FROM --platform=\$$\{BUILDPLATFORM\}/; t' -e ' 1,// s//FROM --platform=\$$\{BUILDPLATFORM\}/' Dockerfile > Dockerfile.cross
+	- docker buildx create --name project-v3-builder
+	docker buildx use project-v3-builder
+	- docker buildx build --push --platform=$(PLATFORMS) --tag $(OPERATOR_IMG) -f Dockerfile.cross .
+	- docker buildx rm project-v3-builder
+	rm Dockerfile.cross
+else
```
ebaron, line 253: Maybe we should explicitly check for podman here. There could be an else branch that errors out with something like "Unsupported IMAGE_BUILDER: foo".

Makefile
```diff
@@ -228,9 +227,37 @@ add-license: addlicense
...
+# You need to be able to push the image for your registry (i.e. if you do not inform a valid value via OPERATOR_IMG=<myregistry/image:<tag>> than the export will fail)
+# If IMAGE_BUILDER is docker, you need to:
+# - able to use docker buildx. More info: https://docs.docker.com/build/buildx/
+# - have enable BuildKit, More info: https://docs.docker.com/develop/develop-images/build_enhancements/
+# If IMAGE_BUILDER is podman, you need to:
+# - install qemu-user-static.
+# To properly provided solutions that supports more than one platform you should use this option.
+PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
+.PHONY: oci-buildx
+oci-buildx: # manifests generate fmt vet test-envtest ## Build OCI image for the manager for cross-platform support
```
ebaron, line 244: Should these prerequisites be uncommented now?

tthvo: Ohh sorry totally forgot. I commented them out to skip steps.

Makefile
ebaron: The podman check looks good, could we still add the `else` branch with something like `$(error unsupported IMAGE_BUILDER: $(IMAGE_BUILDER))`?

tthvo: Oops must have slipped out from this commit. Added now.

ebaron (approved):
Looks good to me! Thanks @tthvo!

Issue #521, opened 2023-02-08
Document ClusterCryostat API


ebaron:
Following the new API to be added in #513, we should update `docs/config.md` and the README to describe this new API and how it differs from the single namespace Cryostat CRD.

PR #544, opened 2023-04-04
docs(api): document ClusterCryostat API

ebaron:
This PR adds a separate `docs/multi-namespace.md` that only highlights the differences compared to the existing Cryostat CRD. It includes references to `docs/config.md` that point out that all existing configuration can be used with the ClusterCryostat CRD as well.

Fixes: #521

### DIFF
--- a/docs/multi-namespace.md
+++ b/docs/multi-namespace.md
@@ -1,5 +1,5 @@
 ## Configuring a multi-namespace Cryostat
-In addition to installing [Cryostat](https://github.com/cryostatio/cryostat) into a single namespace, the Cryostat Operator also allows you to create a Cryostat installation that can work across multiple namespaces. This can be done using the `ClusterCryostat` API. The `ClusterCryostat` API contains all the same [configuration properties](docs/config.md) as the `Cryostat` API does, but has some key differences to enable multi-namespace support that we'll outline below.
+In addition to installing [Cryostat](https://github.com/cryostatio/cryostat) into a single namespace, the Cryostat Operator also allows you to create a Cryostat installation that can work across multiple namespaces. This can be done using the `ClusterCryostat` API. The `ClusterCryostat` API contains all the same [configuration properties](config.md) as the `Cryostat` API does, but has some key differences to enable multi-namespace support that we'll outline below.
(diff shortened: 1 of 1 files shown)


ebaron:
@andrewazores What do you think? Should we add something here about the multi-tenancy concerns in https://github.com/cryostatio/cryostat/issues/1409?

andrewazores:
> @andrewazores What do you think? Should we add something here about the multi-tenancy concerns in [cryostatio/cryostat#1409](https://github.com/cryostatio/cryostat/issues/1409)?

Yes, I think so. Maybe between the "cluster scoped" and "installation namespace" sections.

```markdown
### Data Isolation
When installed in a cluster-wide, multi-namespace manner, all users with access to a Cryostat instance have the same visibility and privileges to all data available to that Cryostat instance. Administrators deploying Cryostat instances must ensure that the users who have access to a Cryostat instance also have equivalent access to all the applications that can be monitored by that Cryostat instance. Otherwise, underprivileged users may use Cryostat to escalate permissions to start recordings and collect JFR data from applications that they do not otherwise have access to.
```

ebaron:
> > @andrewazores What do you think? Should we add something here about the multi-tenancy concerns in [cryostatio/cryostat#1409](https://github.com/cryostatio/cryostat/issues/1409)?
> 
> Yes, I think so. Maybe between the "cluster scoped" and "installation namespace" sections.
> 
> ```gfm
> ### Data Isolation
> When installed in a cluster-wide, multi-namespace manner, all users with access to a Cryostat instance have the same visibility and privileges to all data available to that Cryostat instance. Administrators deploying Cryostat instances must ensure that the users who have access to a Cryostat instance also have equivalent access to all the applications that can be monitored by that Cryostat instance. Otherwise, underprivileged users may use Cryostat to escalate permissions to start recordings and collect JFR data from applications that they do not otherwise have access to.
> ```

Looks good, I've added this section now.

ebaron:
I've added another sentence to that section that mentions how the authorization checks are done against the install namespace instead of the target namespaces.

andrewazores (approved):

Issue #577, opened 2023-05-04
[Request] Sample app using Cryostat Agent


andrewazores:
### Describe the feature

There should be a `make` recipe for deploying a sample app equipped with the Cryostat Agent so that this is easily accessible for testing during development.

PR #578, opened 2023-05-04
feat(sample): add quarkus-test-agent sample app

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #577

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -436,7 +436,7 @@ undeploy_sample_app_quarkus:
 sample_app_agent: undeploy_sample_app_agent
 	$(CLUSTER_CLIENT) create -f config/samples/sample-app-agent.yml
 	@if [ ! "$(CLUSTER_CLIENT)" = "oc" ]; then \
-		@echo "You must now 'oc edit deployment/quarkus-test-agent' and set CRYOSTAT_AGENT_AUTHORIZATION to a suitable auth token value, ex 'oc whoami -t | base64'."; \
+		@echo "You must now 'oc edit deployment/quarkus-test-agent' and set CRYOSTAT_AGENT_AUTHORIZATION to a suitable auth token value."; \
 	else \
 		oc set env deployment/quarkus-test-agent CRYOSTAT_AGENT_AUTHORIZATION="Bearer $(shell oc whoami -t | base64)"; \
 	fi
--- a/Makefile
+++ b/Makefile
@@ -434,12 +434,16 @@ undeploy_sample_app_quarkus:
 
 .PHONY: sample_app_agent
 sample_app_agent: undeploy_sample_app_agent
-	$(CLUSTER_CLIENT) create -f config/samples/sample-app-agent.yml
-	@if [ ! "$(CLUSTER_CLIENT)" = "oc" ]; then \
-		@echo "You must now 'oc edit deployment/quarkus-test-agent' and set CRYOSTAT_AGENT_AUTHORIZATION to a suitable auth token value."; \
-	else \
-		oc set env deployment/quarkus-test-agent CRYOSTAT_AGENT_AUTHORIZATION="Bearer $(shell oc whoami -t | base64)"; \
-	fi
+	@if [ -z "${AUTH_TOKEN}" ]; then \
+		if [ "${CLUSTER_CLIENT}" = "oc" ]; then\
+			AUTH_TOKEN=`oc whoami -t | base64`; \
+		else \
+			echo "'AUTH_TOKEN' must be specified."; \
+			exit 1; \
+		fi; \
+	fi; \
+	$(CLUSTER_CLIENT) create -f config/samples/sample-app-agent.yml; \
+	$(CLUSTER_CLIENT) set env deployment/quarkus-test-agent CRYOSTAT_AGENT_AUTHORIZATION="Bearer $(AUTH_TOKEN)"
 
 .PHONY: undeploy_sample_app_agent
 undeploy_sample_app_agent:
--- a/Makefile
+++ b/Makefile
@@ -434,14 +434,16 @@ undeploy_sample_app_quarkus:
 
 .PHONY: sample_app_agent
 sample_app_agent: undeploy_sample_app_agent
-	$(call new-sample-app,quay.io/andrewazores/quarkus-test:0.0.12,quarkus-test-agent)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -436,6 +432,19 @@ sample_app_quarkus: undeploy_sample_app_quarkus
 undeploy_sample_app_quarkus:
 	- $(CLUSTER_CLIENT) delete all -l app=quarkus-test
 
+.PHONY: sample_app_agent
+sample_app_agent: undeploy_sample_app_agent
+	$(CLUSTER_CLIENT) create -f config/samples/sample-app-agent.yml
+	@if [ ! "$(CLUSTER_CLIENT)" = "oc" ]; then \
+		@echo "You must now 'oc edit deployment/quarkus-test-agent' and set CRYOSTAT_AGENT_AUTHORIZATION to a suitable auth token value."; \
```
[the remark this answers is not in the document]
tthvo: I suppose the token is the headache. Maybe we can let the users specify them when invoking `make`?

config/samples/sample-app-agent.yml
```diff
@@ -0,0 +1,146 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    deployment.kubernetes.io/revision: "3"
+    image.openshift.io/triggers: '[{"from":{"kind":"ImageStreamTag","name":"quarkus-test-agent:0.0.12"},"fieldPath":"spec.template.spec.containers[?(@.name==\"quarkus-test-agent\")].image"}]'
```
tthvo, line 7: We should leave out these k8s annotations  (same for Service) as they are set by the k8s controller?

config/samples/sample-app-agent.yml
```diff
@@ -0,0 +1,146 @@
...
+      restartPolicy: Always
+      schedulerName: default-scheduler
+      securityContext:
+        runAsNonRoot: true
+      terminationGracePeriodSeconds: 30
+      volumes:
+      - name: truststore
+        secret:
+          defaultMode: 420
+          secretName: cryostat-sample-tls
```
tthvo, line 101: Some places are using suffix `cryostat-sample-`. This is the name of the Cryostat CR, that could take other names? Maybe we should call out/warn about this kinda of usage?

Makefile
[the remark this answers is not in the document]
andrewazores: Yea, it's the token which is the problem here, `kubectl whoami` is not there by default (it's an available plugin). Passing it as a variable and failing if it isn't supplied seems to make sense.

config/samples/sample-app-agent.yml
andrewazores: It's the same name as in the sample CR (`make create_cryostat_cr`). IMO it's not a big deal since these are both really developer conveniences. In particular this sample app thing is not intended for use by end users.

config/samples/sample-app-agent.yml
```diff
@@ -0,0 +1,132 @@
...
+      deployment: quarkus-test-agent
+  strategy:
+    rollingUpdate:
+      maxSurge: 25%
+      maxUnavailable: 25%
+    type: RollingUpdate
+  template:
+    metadata:
+      annotations:
+        openshift.io/generated-by: OpenShiftNewApp
```
tthvo, line 25: I think we can remove this annotation. 

How about this? This should give a minimal specs for the deployment and service objects.

```diff
diff --git a/config/samples/sample-app-agent.yml b/config/samples/sample-app-agent.yml
index 140ba5a..5e2acd3 100644
--- a/config/samples/sample-app-agent.yml
+++ b/config/samples/sample-app-agent.yml
@@ -4,28 +4,16 @@ kind: Deployment
 metadata:
   labels:
     app: quarkus-test-agent
-    app.kubernetes.io/component: quarkus-test-agent
-    app.kubernetes.io/instance: quarkus-test-agent
   name: quarkus-test-agent
 spec:
-  progressDeadlineSeconds: 600
   replicas: 1
-  revisionHistoryLimit: 10
   selector:
     matchLabels:
-      deployment: quarkus-test-agent
-  strategy:
-    rollingUpdate:
-      maxSurge: 25%
-      maxUnavailable: 25%
-    type: RollingUpdate
+      app: quarkus-test-agent
   template:
     metadata:
-      annotations:
-        openshift.io/generated-by: OpenShiftNewApp
-      creationTimestamp: null
       labels:
-        deployment: quarkus-test-agent
+        app: quarkus-test-agent
     spec:
       containers:
       - env:
@@ -74,18 +62,13 @@ spec:
           capabilities:
             drop:
             - ALL
-        terminationMessagePath: /dev/termination-log
-        terminationMessagePolicy: File
         volumeMounts:
         - mountPath: /var/run/secrets/myapp/truststore.p12
           name: truststore
           subPath: truststore.p12
-      dnsPolicy: ClusterFirst
       restartPolicy: Always
-      schedulerName: default-scheduler
       securityContext:
         runAsNonRoot: true
-      terminationGracePeriodSeconds: 30
       volumes:
       - name: truststore
         secret:
@@ -97,14 +80,10 @@ kind: Service
 metadata:
   labels:
     app: quarkus-test-agent
-    app.kubernetes.io/component: quarkus-test-agent
-    app.kubernetes.io/instance: quarkus-test-agent
   name: quarkus-test-agent
 spec:
-  internalTrafficPolicy: Cluster
-  ipFamilies:
-  - IPv4
-  ipFamilyPolicy: SingleStack
+  selector:
+    app: quarkus-test-agent
   ports:
   - name: jfr-jmx
     port: 9097
@@ -126,7 +105,3 @@ spec:
     port: 8443
     protocol: TCP
     targetPort: 8443
-  selector:
-    deployment: quarkus-test-agent
-  sessionAffinity: None
-  type: ClusterIP

```

config/samples/sample-app-agent.yml
```diff
@@ -0,0 +1,132 @@
...
+      dnsPolicy: ClusterFirst
+      restartPolicy: Always
+      schedulerName: default-scheduler
+      securityContext:
+        runAsNonRoot: true
+      terminationGracePeriodSeconds: 30
+      volumes:
+      - name: truststore
+        secret:
+          defaultMode: 420
```
tthvo, line 92: Some reasons to set this permission bit to `420`?

config/samples/sample-app-agent.yml
```diff
@@ -0,0 +1,132 @@
...
+        - name: KEYSTORE_PASS
+          valueFrom:
+            secretKeyRef:
+              key: KEYSTORE_PASS
+              name: cryostat-sample-keystore
+        - name: JAVA_OPTS_APPEND
+          value: |-
+            -Djavax.net.ssl.trustStore=/var/run/secrets/myapp/truststore.p12
+            -Djavax.net.ssl.trustStorePassword=$(KEYSTORE_PASS)
+        image: quay.io/andrewazores/quarkus-test@sha256:96f0f8bf0878a0843df08f9fec8ee9cf9c0ea09ee8c462b2c28ae504085a8b1b
```
tthvo, line 59: Should we use a tag here instead of the image hash?

config/samples/sample-app-agent.yml
andrewazores: I think I got this from an example that @ebaron had put together elsewhere, but I don't know where that came from originally. I can set up a testing environment again and experiment with leaving this configuration out and see if that works.

tthvo: I think `420` should already be the default value in decimal (octal `0644`).

```bash
oc explain pods.spec.volumes.secret.defaultMode
KIND:     Pod
VERSION:  v1

FIELD:    defaultMode <integer>

DESCRIPTION:
     defaultMode is Optional: mode bits used to set permissions on created files
     by default. Must be an octal value between 0000 and 0777 or a decimal value
     between 0 and 511. YAML accepts both octal and decimal values, JSON
     requires decimal values for mode bits. Defaults to 0644. Directories within
     the path are not affected by this setting. This might be in conflict with
     other options that affect the file mode, like fsGroup, and the result can
     be other mode bits set.
```

config/samples/sample-app-agent.yml
```diff
@@ -0,0 +1,106 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: quarkus-test-agent
+  template:
+    metadata:
+      labels:
+        app: quarkus-test-agent
+    spec:
+      containers:
```
tthvo, line 18: We probably should define some resource requests (cpu, memory) here for this container.

andrewazores: The agent itself should have almost negligible overhead, so this really just depends on whatever the application is that the agent is attached to. In this case it's actually the same as the `sample_app_quarkus`, ie it's just a Quarkus "Hello World" kind of application that just does nothing but serve a static webpage, so the resource requirements are super low.

tthvo (commented):
Looks good to me! Just seeing if @ebaron has any comments.

ebaron (approved):

tthvo (approved):

Issue #584, opened 2023-05-11
[Bug] Namespace deletion can hang


ebaron:
### Current Behavior

When attempting to delete a namespace containing a Cryostat installation, the deletion may hang. This happens if the lock config map is deleted before the finalizer logic completes for the Cryostat or ClusterCryostat.

```
2023-05-11T16:52:37Z	ERROR	Reconciler error	{"controller": "cryostat", "controllerGroup": "operator.cryostat.io", "controllerKind": "Cryostat", "Cryostat": {"name":"cryostat-sample","namespace":"thvo-topology"}, "namespace": "thvo-topology", "name": "cryostat-sample", "reconcileID": "a89c44d0-cce0-402b-bead-571f00d52288", "error": "configmaps \"cryostat-sample-lock\" is forbidden: unable to create new content in namespace thvo-topology because it is being terminated"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/remote-source/deps/gomod/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.5/pkg/internal/controller/controller.go:329
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/remote-source/deps/gomod/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.5/pkg/internal/controller/controller.go:274
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/remote-source/deps/gomod/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.5/pkg/internal/controller/controller.go:235
```

### Expected Behavior

Deletion should not hang, and any resources created for that Cryostat installation are cleaned up.

### Steps To Reproduce

1. Install the Cryostat Operator with AllNamespaces mode in namespace A
2. Create a Cryostat CR in a namespace B, or a ClusterCryostat with install namespace of B
3. Wait for Cryostat or ClusterCryostat to be ready
4. Scale down operator deployment
5. Delete namespace B
6. Scale up operator deployment

### Environment

```Markdown
- Environment: All
- Version: 2.3.0
```

PR #593, opened 2023-05-26
fix(cleanup): allow deletion to proceed without config map

ebaron:
# Welcome to Cryostat! 👋

Fixes: #584 

## Description of the change:
The main change is to move the lock config map creation after handling the finalizer logic. The rest of the patch is additional test assurances that this doesn't break the name conflict handling introduced in #545.

## Motivation for the change:
This fixes the namespace deletion bug that's present in 2.3.0.

## How to manually test:
Ensure that deleting a conflicting CR doesn't delete/modify the original Cryostat installation:

1. Deploy operator into cryostat-operator-system
```
$ make deploy OPERATOR_IMG=quay.io/ebaron/cryostat-operator:namespace-del-fix-01
```
2. Create namespace c
```
$ kubectl create namespace c
```
3. Create Cryostat
```yaml
apiVersion: operator.cryostat.io/v1beta1
kind: Cryostat
metadata:
  name: cryostat-sample
  namespace: c
spec:
  minimal: false
  enableCertManager: true
  reportOptions:
    replicas: 0
```
4. Create ClusterCryostat
```yaml
apiVersion: operator.cryostat.io/v1beta1
kind: ClusterCryostat
metadata:
  name: cryostat-sample
spec:
  minimal: false
  enableCertManager: true
  installNamespace: c
  targetNamespaces:
  - c
  reportOptions:
    replicas: 0
```
5. Observe finalizer on ClusterCryostat
```
$ kubectl get clustercryostat cryostat-sample -o jsonpath='{.metadata.finalizers}'
["operator.cryostat.io/cryostat.finalizer"]
```
6. Delete ClusterCryostat
```
$ kubectl delete clustercryostat cryostat-sample
```
7. Check that ClusterRoleBinding and ConsoleLink were not deleted
```
$ kubectl get clusterrolebinding cryostat-22ca2798ad950a2953c096d484826501ddbae477d771250523c5523d9bd5a842
NAME                                                                        ROLE                                     AGE
cryostat-22ca2798ad950a2953c096d484826501ddbae477d771250523c5523d9bd5a842   ClusterRole/cryostat-operator-cryostat   46s
$ kubectl get consolelink cryostat-22ca2798ad950a2953c096d484826501ddbae477d771250523c5523d9bd5a842
NAME                                                                        TEXT       URL                                                                 MENU   AGE
cryostat-22ca2798ad950a2953c096d484826501ddbae477d771250523c5523d9bd5a842   Cryostat   https://cryostat-sample-cryostat-operator-system.apps.example.com          80s
```
8. Check that additionalCORSAllowedOrigins entry is still present in APIServer config:
```
$ kubectl get apiserver cluster -o jsonpath='{.spec.additionalCORSAllowedOrigins}'
["https://cryostat-sample-cryostat-operator-system\\.apps\\.example\\.com"]
```

Check that the namespace deletion bug is fixed:

1. Scale down the operator deployment
```
$ kubectl scale --replicas=0 deploy cryostat-operator-controller-manager
deployment.apps/cryostat-operator-controller-manager scaled
```
2. In another terminal, delete the namespace
```
$ kubectl delete namespace c
namespace "c" deleted
# hangs...
```
3. Scale the operator deployment back up
```
$ kubectl scale --replicas=1 deploy cryostat-operator-controller-manager
deployment.apps/cryostat-operator-controller-manager scaled
```
4. After some time, the namespace deletion proceeds

### DIFF
--- a/internal/controllers/clustercryostat_controller_test.go
+++ b/internal/controllers/clustercryostat_controller_test.go
@@ -73,6 +73,7 @@ var _ = Describe("ClusterCryostatController", func() {
 
 		JustBeforeEach(func() {
 			c.commonJustBeforeEach(t)
+			t.reconcileCryostatFully()
 		})
 
 		JustAfterEach(func() {
@@ -80,7 +81,7 @@ var _ = Describe("ClusterCryostatController", func() {
 		})
 
 		It("should create the expected main deployment", func() {
-			t.expectDeployment()
+			t.expectMainDeployment()
 		})
 
 		It("should create RBAC in each namespace", func() {
@@ -103,24 +104,18 @@ var _ = Describe("ClusterCryostatController", func() {
 					t.NewRoleBinding(targetNamespaces[0]),
 					t.NewRoleBinding(targetNamespaces[1]))
 			})
-
 			It("should create the expected main deployment", func() {
-				t.expectDeployment()
+				t.expectMainDeployment()
 			})
-
 			It("leave RBAC for the first namespace", func() {
 				t.expectRBAC()
 			})
-
 			It("should remove RBAC from the second namespace", func() {
-				t.reconcileCryostatFully()
-
 				binding := t.NewRoleBinding(targetNamespaces[1])
 				err := t.Client.Get(context.Background(), types.NamespacedName{Name: binding.Name, Namespace: binding.Namespace}, binding)
 				Expect(err).ToNot(BeNil())
 				Expect(errors.IsNotFound(err)).To(BeTrue())
 			})
-
 			It("should update the target namespaces in Status", func() {
 				t.expectTargetNamespaces()
 			})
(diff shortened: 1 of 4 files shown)


andrewazores:
> The main change is to move the lock config map creation after handling the finalizer logic.

I'm still trying to wrap my head around this because it seems like it should work, but it also seems like it could cause other issues to perform other logic before acquiring the lock. But, the test case worked as expected in crc too. In this case since the only other logic is either adding a finalizer or acting on the finalizer, that doesn't need mutex locking since it shouldn't cause conflicts elsewhere in the object graph. Maybe at worst you end up trying to delete the same object twice?

ebaron:
The objects that require finalizer logic should never conflict because of the following:
1. The ClusterRoleBinding and ConsoleLink have a prefix matching their Kind. For Cryostat the name is of the form `cryostat-namespacedNameHash`, for ClusterCryostat it's `clustercryostat-namespacedNameHash`.
2. The APIServer `additionalCORSAllowedOrigins`. This one uses the value of `status.applicationURL` from the CR to determine what to remove from the list of URLs. Since the lock config map creation occurs much before this field is set, only the first of the conflicting CRs can have this value set.

As for adding the finalizer to the CR, that's also fine since it only affects the CR itself and even CRs that fail to "acquire" the lock CR can still be finalized. The ClusterRoleBinding and ConsoleLink won't exist and there will be no URL to remove from the APIServer config. These would all be no-ops.

PR #599, opened 2023-07-13
chore(oauth): remove modifications to OpenShift api-server

tthvo:
# Welcome to Cryostat! 👋

Related to https://github.com/cryostatio/cryostat/issues/1489
Depends on https://github.com/cryostatio/cryostat-web/pull/1059

## Description of the change:

Remove the reconciler logics that modify the api-server to add cryostat application URL as allowed CORS origins.

## Motivation for the change:

https://github.com/cryostatio/cryostat-web/pull/1059 will submit form on the web client to clear login session. This removes the requirement for making CORS request.

### DIFF
--- a/internal/controllers/reconciler_test.go
+++ b/internal/controllers/reconciler_test.go
@@ -1282,9 +1282,6 @@ func (c *controllerTest) commonTests() {
 					Expect(link.Spec.NamespaceDashboard).To(Equal(expectedLink.Spec.NamespaceDashboard))
 				})
 			})
-			It("should add application url to APIServer AdditionalCORSAllowedOrigins", func() {
-				t.expectAPIServer()
-			})
 			It("should add the finalizer", func() {
 				t.expectCryostatFinalizerPresent()
 			})
@@ -1810,9 +1807,6 @@ func (c *controllerTest) commonTests() {
 					Context("on OpenShift", func() {
 						It("should not delete exisiting ConsoleLink", func() {
 							otherInput.expectConsoleLink()
-						})
-						It("should not remove CORS entry from APIServer", func() {
-							otherInput.expectAPIServer()
 						})
 					})
 				})
--- a/internal/controllers/reconciler_test.go
+++ b/internal/controllers/reconciler_test.go
@@ -1293,7 +1293,7 @@ func (c *controllerTest) commonTests() {
 					apiServer := &configv1.APIServer{}
 					err := t.Client.Get(context.Background(), types.NamespacedName{Name: "cluster"}, apiServer)
 					Expect(err).ToNot(HaveOccurred())
-					Expect(apiServer.Spec.AdditionalCORSAllowedOrigins).ToNot(ContainElement("https://cryostat\\.example\\.com"))
+					Expect(apiServer.Spec.AdditionalCORSAllowedOrigins).ToNot(ContainElement(fmt.Sprintf("https://%s\\.example\\.com", t.Name)))
 					Expect(apiServer.Spec.AdditionalCORSAllowedOrigins).To(ContainElement("https://an-existing-user-specified\\.allowed\\.origin\\.com"))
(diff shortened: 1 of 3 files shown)


tthvo:
I think the main issue is that the application URL is not cleaned up after upgrades to newer operator. Wonder if we could do something about it or just document it for customers?

ebaron (commented):
Hi @tthvo, thanks for handling this. I just have one suggestion: we should keep `deleteCorsAllowedOrigins`. This ensures that our modifications get removed when the user upgrades the operator.

tthvo:
Hi I updated to only remove the addition logic and also updated tests. Any other tests you think we should add here?

ebaron:
> Hi I updated to only remove the addition logic and also updated tests. Any other tests you think we should add here?

Hey Thuan, that's a good point. I think a test that verifies that an existing API server modification actually gets removed would be useful. We'd have to modify the APIServer manually as part of the test setup. Then the test itself could be like: https://github.com/cryostatio/cryostat-operator/blob/1a02ba82282cec37931b9993970dc416c0e1d08a/internal/controllers/reconciler_test.go#L1318-L1324

I also just realized that we should probably call `deleteCorsAllowedOrigins` from `reconcileOpenShift` and not `finalizeOpenShift`. We don't want the user to have to delete the CR for our modifications to be removed.

ebaron:
> I also just realized that we should probably call `deleteCorsAllowedOrigins` from `reconcileOpenShift` and not `finalizeOpenShift`. We don't want the user to have to delete the CR for our modifications to be removed.

I guess it should actually be called from both methods. If for some reason a user upgraded the operator with a deleted Cryostat, we would need to make the modification during the finalizer step.

tthvo:
Right! Thanks for the suggestions. I called `deleteCorsAllowedOrigins` in both functions now and added a new test to check if an existing application URL is removed.

internal/controllers/reconciler_test.go
```diff
@@ -1282,8 +1282,20 @@ func (c *controllerTest) commonTests() {
...
+					t.objs = []ctrlclient.Object{
+						t.NewApiServerWithApplicationURL(),
+						t.NewNamespace(),
+					}
+				})
+				It("should remove the application url", func() {
+					apiServer := &configv1.APIServer{}
+					err := t.Client.Get(context.Background(), types.NamespacedName{Name: "cluster"}, apiServer)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(apiServer.Spec.AdditionalCORSAllowedOrigins).ToNot(ContainElement("https://cryostat\\.example\\.com"))
```
ebaron, line 1296: I would use `t.Name` here instead of `cryostat`.

ebaron (commented):
Hey @tthvo, looks good! Just going to give it a try in OpenShift

internal/controllers/reconciler_test.go
tthvo: Thanks! Updated now!

ebaron (commented):
While testing, I noticed we get a lot of `Removed from APIServer CORS allowed origins` messages in the log. In the majority of cases there will be nothing to remove and I'd rather not have us interact with this config at all, even for a no-op.

Could I trouble you to just wrap the `client.Update` and log message in a conditional that only runs when a match is found?

https://github.com/cryostatio/cryostat-operator/blob/1a02ba82282cec37931b9993970dc416c0e1d08a/internal/controllers/openshift.go#L189-L204

tthvo:
> While testing, I noticed we get a lot of Removed from APIServer CORS allowed origins messages in the log. In the majority of cases there will be nothing to remove and I'd rather not have us interact with this config at all, even for a no-op

Oh yess! I wonder so. Make sense to me! Updated now!

ebaron (approved):
Thanks @tthvo!

Issue #618, opened 2023-08-30
[Task] PR command to run build + tests


mwangggg:
### Current Behavior

Instead of running the CI build + tests when the PR is changed (label applied, new commit, etc.), they should only run when a PR command (/build_test for example) is commented.

PR #619, opened 2023-08-30
chore(ci): PR command build_test

mwangggg:
# Welcome to Cryostat! 👋

Fixes: #618


mwangggg:
demo: https://github.com/mwangggg/cryostat-operator/pull/1

.github/workflows/ci.yaml
```diff
@@ -148,20 +190,21 @@ jobs:
...
+        SCORECARD_REGISTRY_PASSWORD="${{ secrets.GITHUB_TOKEN }}" \
         BUNDLE_IMG="${{ steps.push-bundle-to-ghcr.outputs.registry-path }}" \
         make test-scorecard
     - name: Clean up Kind cluster 
       run: kind delete cluster -n ci-${{ github.run_id }}
+
   build-operator:
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'push' && github.repository_owner == 'cryostatio' }}
+    needs: [checkout-branch]
```
tthvo, line 207: Jobs that starts with `build_` should not be touched in these cases because its triggered on the `main` push, not PR.

mwangggg: oh ok so I'll separate into two different yml files

tthvo: Sounds good!

andrewazores (dismissed):
LGTM, any other feedback @tthvo ?

.github/workflows/ci.yaml
andrewazores: Ah wait, is this feedback still outstanding? I don't see a more recent commit.

mwangggg: oh I think I forgot to push my changes

andrewazores (dismissed):

andrewazores:
LGTM but since @ebaron does more of the maintainer work here, I think he should at least take a look over the changes.

.github/workflows/test-ci.yaml
```diff
@@ -0,0 +1,228 @@
...
+      pull-requests: write
+    if: github.repository_owner == 'cryostatio' && github.event.issue.pull_request && startsWith(github.event.comment.body, '/build_test')
+    steps:
+    - name: Fail if needs-triage label applied 
+      if: ${{ contains(github.event.issue.labels.*.name, 'needs-triage') }}
+      run: exit 1
+    - name: Show warning if permission is denied
+      if: |
+        !(github.event.comment.author_association == 'MEMBER' || github.event.comment.author_association == 'OWNER')
+        && (!contains(github.event.issue.labels.*.name, 'safe-to-test') || github.event.issue.user.name != github.event.comment.user.name)
```
ebaron, line 35: I'm having a bit of trouble parsing this conditional. What is this comparison used for: `github.event.issue.user.name != github.event.comment.user.name`?

ebaron (commented):
Hi @mwangggg, very sorry for taking so long to get to this. Looks great overall, I just have a couple questions below. I really like splitting the workflows into separate PR and push files. This should improve security as well since the Quay.io secret isn't available with `pull_request_target` events anymore.

.github/workflows/build-ci.yml
```diff
@@ -0,0 +1,148 @@
...
+  OPENSUSE_UNOFFICIAL_LIBCONTAINERS_SOURCE_URL: "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/unstable/xUbuntu_22.04"
+
+jobs:
+  check-before-build:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'cryostatio' }}
+    steps:
+    - name: Fail if safe-to-test label NOT applied
+      if: ${{ github.event_name == 'pull_request_target' && !contains(github.event.pull_request.labels.*.name, 'safe-to-test') }}
+      run: exit 1
```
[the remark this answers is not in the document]
andrewazores: Ah yes, that sounds right.

.github/workflows/test-ci.yaml
```diff
@@ -0,0 +1,228 @@
+name: test-CI
+
+concurrency:
+  group: ci-${{ github.run_id }}
+  cancel-in-progress: true
+
+on:
+  issue_comment:
```
tthvo, line 8: We also wants to run tests on push too. One way is to extract the test jobs into a separate reusable workflow: https://docs.github.com/en/actions/using-workflows/reusing-workflows

.github/workflows/test-ci.yaml
```diff
@@ -0,0 +1,228 @@
...
+        kubectl rollout status -w deployment/ingress-nginx-controller -n ingress-nginx --timeout 5m
+    - name: Install Operator Lifecycle Manager
+      run: curl -sL https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.24.0/install.sh | bash -s v0.24.0
+    - name: Install Cert Manager
+      run: make cert_manager
+    - uses: redhat-actions/podman-login@v1
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.GITHUB_TOKEN }}
```
tthvo, line 218: We should set the default token to "restrictive" (read-only) by defaults and explicity request write permissions for jobs.

See https://github.com/cryostatio/cryostat/pull/1635

.github/workflows/test-ci.yaml
tthvo: I believe in the case `safe-to-test` label is applied, only the author of the PR can use the command. So, not anyone (i.e. some spam bots) to use it freely.

.github/workflows/test-ci.yaml
mwangggg: I thought this was what I did with setting the packages permission to write?

.github/workflows/test-ci.yaml
mwangggg: I thought since the push and issue_comment events are so different, it would be simpler to just make a new workflow for push - but if necessary, I can make the test jobs into a reusable workflow I think?

.github/workflows/test-ci.yaml
andrewazores: I think it's a two-parter: set the workflow to request `packages: write` permissions, and then also remove the `GITHUB_TOKEN` secret-as-password reference so that the job uses the default token that only has the `packages: write`, rather than the PAT in the `GITHUB_TOKEN` secret that probably still has broader permissions.

.github/workflows/test-ci.yaml
tthvo: Yehh right! We should split the test jobs into resusable workflow!

tthvo: And have 2 workflows: push + issue_comment that calls it.

.github/workflows/test-ci.yaml
tthvo: Yehh, the restrictive mode can be set for repo or org scope in settings, not workflow tho. So, this is good. 

But I think we should revert it back to the old PAT? Ali is working on the replacement here: https://github.com/cryostatio/cryostat-operator/pull/628

Just to keep the history clear :))

mwangggg: ah I see ok I'll keep the old PAT :+1:

mwangggg:
demo: https://github.com/mwangggg/cryostat-operator/pull/2

.github/workflows/test-ci-push.yml
```diff
@@ -0,0 +1,36 @@
...
+          prefix="pr-${{ github.event.number }}"
+        fi
+        echo "tag=${prefix}-$GITHUB_SHA" >> $GITHUB_OUTPUT
+  
+  run-test-jobs:
+    uses: ./.github/workflows/test-ci-reusable.yml
+    needs: [get-test-image-tag]
+    with:
+      repository: ${{ github.event.pull_request.head.repo.full_name }}
+      ref: ${{ github.event.pull_request.head.ref }}
```
ebaron, line 35: I think these `github.event.pull_request` properties are undefined on `push` event, in which case we're getting the Checkout action to use its defaults. These defaults should be the originating repository and ref, which for push is what we want anyways. Maybe for clarity we should make these parameters optional and omit them for the push workflow?

ebaron (approved):
This looks good to me! @tthvo any other comments?

tthvo (approved):
Looks good to me too! Nice job @mwangggg! Thankss!

.github/workflows/test-ci.yaml
tthvo: I can't resolve comments ahha but this should be already resolved :))

ebaron:
@mwangggg I noticed the CI is failing with the following warning when trying to push the test images to GHCR:
`Warning: Username is provided, but password is missing`

I think we need to add something like this when calling the reusable workflow:
```
    secrets:
       GHCR_PR_TOKEN: ${{ secrets.GHCR_PR_TOKEN }}
```

https://docs.github.com/en/actions/using-workflows/reusing-workflows#passing-secrets-to-nested-workflows

tthvo:
Or after #628, I think the `GITHUB_TOKEN` is automatically inherited. Mayybe we wait for those changes?

PR #617, opened 2023-08-29
chore(deployment): adjust resource requests

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits): `git commit -S -m "YOUR_COMMIT_MESSAGE"`
_______________________________________________

Related to https://github.com/cryostatio/cryostat/issues/1629

## Description of the change:
Tweaks default resource requests for operand containers.

## Motivation for the change:
This aligns the resource requests with the [rough minimums applied in the Cryostat smoketest](https://github.com/cryostatio/cryostat/pull/1636). It is no hard rule or exact science, but these numbers should generally result in a working Cryostat deployment that is able to reliably handle modestly sized JFR files with reasonable speed.

PR #626, opened 2023-09-12
feat(deployment): core resource requests split with sidecar

andrewazores:
# Welcome to Cryostat! 👋

Based on #617

## Description of the change:
If there is no configuration for `-reports` sidecar(s), increase the resources requested for the main `cryostat` container instead.

## Motivation for the change:
If the CR does not have any report generation configuration then the Cryostat server will fall back to generating reports itself by forking a subprocess within its container. This at least needs some memory allocation of its own and preferably some CPU time as well to complete its work in a reasonable time. So, in deployments like this, assign the maximum of the CPU and Memory requests that would go to either the core container or the reports container(s) to the core container so that the additional headroom may be used for subprocess report generation.


andrewazores:
Dropping this. It's long out of date, and since 3.0 where Cryostat itself was rewritten on top of Quarkus, it no longer forks a separate subprocess and the OOM handling is much more robust (and just results in a request failure, not an OOM crash).

Issue #630, opened 2023-09-20
[Request] Support custom labels in deployments and pods


OpenGuidou:
### Describe the feature

Hello,

It's possible today to provide labels for the services and ingresses, but not for the deployments and pods (kind of static definition here: https://github.com/cryostatio/cryostat-operator/blob/main/internal/controllers/common/resource_definitions/resource_definitions.go#L105).

It would be a nice improvement to support custom labels there.

PR #632, opened 2023-09-21
feat(openshift): custom extra labels/annotations on deployments/pods

OpenGuidou:
# Welcome to Cryostat! 👋

Fixes: #630 

## Description of the change:
This change adds allows the users to provide additional labels and annotations to be used in the cryostat deployment and pod.

## Motivation for the change:
This change is helpful because some environment have a strong labeling policy for all resources.


andrewazores:
https://github.com/cryostatio/cryostat-operator/actions/runs/6263310857

andrewazores:
Ah right, that CI command is currently not working: https://github.com/cryostatio/cryostat-operator/pull/619#issuecomment-1728475354

api/v1beta1/cryostat_types.go
```diff
@@ -88,6 +88,21 @@ type CryostatSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Credentials Database Options"
 	JmxCredentialsDatabaseOptions *JmxCredentialsDatabaseOptions `json:"jmxCredentialsDatabaseOptions,omitempty"`
+	// Options to configure the Cryostat resources metadata
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Resources metadata"
+	Metadata *ResourceMetadata `json:"metadata,omitempty"`
```
ebaron, line 94: I'd like to differentiate this from other metadata defined elsewhere in this CRD (e.g. services, ingress). Maybe `PodMetadata` or `WorkloadMetadata`, any preferences @andrewazores?

internal/test/resources.go
```diff
@@ -654,6 +655,19 @@ func (r *TestResources) NewCryostatWithDatabaseSecretProvided() *model.CryostatI
...
+func (r *TestResources) NewCryostatWithAdditionalMetadata() *model.CryostatInstance {
+	cr := r.NewCryostat()
+	cr.Spec.Metadata = &v1beta1.ResourceMetadata{
+		Labels: map[string]string{
+			"myExtraLabel": "myLabel",
+		},
+		Annotations: map[string]string{
+			"myExtraAnnotation": "myAnnotation",
+		},
+	}
```
ebaron, line 667: Could I ask you to add some of the default labels and annotations here (e.g. `app`, `component`, `app.openshift.io/connects-to`) with arbitrary values. Then the test case can verify that the operator replaces the values for those labels/annotations with its own values.

internal/test/resources.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	certv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	certMeta "github.com/cert-manager/cert-manager/pkg/apis/meta/v1"
+	"github.com/cryostatio/cryostat-operator/api/v1beta1"
```
ebaron, line 24: This package is already imported under `operatorv1beta1`.

ebaron (commented):
Hi @OpenGuidou, thank you for your contribution! Excellent work as well. I have some comments below, but more generally, do you think it makes sense to have a separate set of labels and annotations for the reports deployment (if enabled)? Typically, we'd put any properties that affect the reports deployment under the ReportOptions property.

api/v1beta1/cryostat_types.go
andrewazores: `PodMetadata` or even `OperandMetadata` maybe, since that name would also suggest that the metadata could be applied to ex. the `Deployment`?

OpenGuidou: My initial idea was to provide there labels and annotations that would apply to all resources created, but you're right I missed the ingress/services :smile: 

Should I split that in 2 to differentiate the deployment and pod metadata, or apply it on top for the ingress/services as well ?

internal/test/resources.go
OpenGuidou: Added

internal/test/resources.go
OpenGuidou: Removed

api/v1beta1/cryostat_types.go
andrewazores: I suppose if it's possible to configure custom labels/annotations like this then it may as well be fully customizable for each kind of resource. On the other hand, I imagine that most of the time users will just want the same labels/annotations across all of the resources, so duplicating the configuration for each kind is a bit of a waste. Perhaps in that case the user can put their common labels/annotations into a ConfigMap and reference that from each resource kind's custom metadata config...

@OpenGuidou which way would make more sense to you as a user? It's easy for me to look at this and think that if it's done at all then it should be done with maximum flexibility, but that also entails maximum complexity :-)

OpenGuidou: For my I would look at it the way we are using helm templates usually: have common labels/annotations that apply everywhere, then let the possibility to override it at each level (resource type).
Then it'll complexify a bit the code here :smile:

OpenGuidou: Well after a bit of thinking, I guess the lesser of two evils would indeed to differentiate pod and deployment metadata in the configuration. Not a lot of overload in the configuration, quite simple in the code and we keep the flexibility.

Are you ok with that ?

andrewazores: Sounds fine to me. Would that also have separate metadata configuration for ingress/services too? And each of the pod/deployments for the main Cryostat container, the reports sidecar(s), the jfr-datasource, and the Grafana dashboard...

OpenGuidou: Updated

internal/controllers/reconciler.go
```diff
@@ -442,7 +442,7 @@ func (r *Reconciler) createOrUpdateDeployment(ctx context.Context, deploy *appsv
 	op, err := controllerutil.CreateOrUpdate(ctx, r.Client, deploy, func() error {
 		// TODO consider managing labels and annotations using CRD
```
andrewazores, line 443: @ebaron would this TODO be addressed by these changes?

andrewazores (approved):
I think this looks good. @ebaron would you want the ingresses/services/routes to also have customizable labels/annotations like this or are you OK with it just going as far as pod/deployment for now?

internal/controllers/reconciler.go
ebaron: Yes, I think that TODO is now covered by this PR and can be deleted if you don't mind @OpenGuidou

api/v1beta1/cryostat_types.go
ebaron: Sounds good to me.

internal/controllers/reconciler.go
OpenGuidou: Done

ebaron (commented):
@OpenGuidou looks good! Just noticed one more thing. Could you run `make bundle` to regenerate the CSV to include these new properties? We recently upgraded to Operator SDK 1.31.0, so if you have another version installed locally you'll need to upgrade.

OpenGuidou:
> @OpenGuidou looks good! Just noticed one more thing. Could you run `make bundle` to regenerate the CSV to include these new properties? We recently upgraded to Operator SDK 1.31.0, so if you have another version installed locally you'll need to upgrade.

Ha I missed that, it's now done

ebaron:
@OpenGuidou Sorry one more thing. We require that all commits be signed in order for me to merge this PR:
https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

OpenGuidou:
> @OpenGuidou Sorry one more thing. We require that all commits be signed in order for me to merge this PR: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

Done !

andrewazores (approved):

Issue #537, opened 2023-03-16
Enable Dependabot to upgrade dependencies


ebaron:
We should be somewhat careful about dependency versions needed by the Operator SDK (see: #536). I think it would still be safe to tell Dependabot to upgrade patch versions of these dependencies at least (e.g. X.Y.Z -> X.Y.Z+1).

PR #648, opened 2023-10-18
ci(dependabot): configure dependabot updates

tthvo:
# Welcome to Cryostat! 👋

Fixes: #537 

## Description of the change:

Configure dependabot updates to ignore k8s related deps. These deps will be upgrade with operator-sdk upgrades.

## Motivation for the change:

See #537

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+go.mod @cryostatio/maintainers 
+go.sum @cryostatio/maintainers 
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+version: 2
+
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: daily
+    reviewers:
+      - "cryostatio/reviewers"
+    labels:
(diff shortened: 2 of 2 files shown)


tthvo:
Reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

andrewazores:
I think this makes sense but I defer to @ebaron .

ebaron (commented):
Thanks for doing this @tthvo! For the k8s dependencies, this will have the effect of only updating patch releases right?

ebaron:
> Hi yes, I believe so.
> 
> > By default all dependencies that are explicitly defined in a manifest are kept up to date by Dependabot version updates.
> 
> > You can use allow and ignore to customize which dependencies to maintain. Dependabot checks for all allowed dependencies and then filters out any ignored dependencies or versions. So a dependency that is matched by both an allow and an ignore will be ignored.
> 
> So, unless specified with `version-update:semver-patch` in `ignore`, the deps should receive patch update.

Excellent, thank you!

ebaron (approved):

Issue #595, opened 2023-06-06
[Story] Mirror CA secret in target namespaces


ebaron:
### Describe the feature

When communicating with Cryostat using TLS, target applications may need to trust Cryostat's CA certificate. Since Kubernetes only allows workloads to mount secrets within the same namespace, users would have to manually copy the Cryostat CA certificate secret. It would be useful if the operator could automate this by placing a copy of the CA certificate in each target namespace.

See: https://cert-manager.io/docs/tutorials/syncing-secrets-across-namespaces/

PR #661, opened 2023-10-24
feat(cert): add cryostat CA cert to target namespaces

mwangggg:
# Welcome to Cryostat! 👋

Fixes: #595

### DIFF
--- a/internal/controllers/certmanager.go
+++ b/internal/controllers/certmanager.go
@@ -137,7 +137,10 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 				Type: secret.Type,
 			}
 			err = r.createOrUpdateSecret(ctx, namespaceSecret, cr.Object, func() error {
-				namespaceSecret.Data = secret.Data
+				if namespaceSecret.Data == nil {
+					namespaceSecret.Data = map[string][]byte{}
+				}
+				namespaceSecret.Data[corev1.TLSCertKey] = secret.Data[corev1.TLSCertKey]
 				return nil
 			})
 			if err != nil {
--- a/internal/controllers/certmanager.go
+++ b/internal/controllers/certmanager.go
@@ -130,6 +130,12 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 		}
 		namespaceSecret := secret.DeepCopy()
 		namespaceSecret.Namespace = ns
+		err = r.createOrUpdateSecret(ctx, namespaceSecret, cr.Object, func() error {
+			if secret.StringData == nil {
+				secret.StringData = map[string]string{}
+			}
+			return nil
+		})
 	}
 	// Delete any Cryostat CA secrets in target namespaces that are no longer requested
 	for _, ns := range toDelete(cr) {
@@ -156,6 +162,37 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 	}
 	tlsConfig.CACert = caBytes
 	return tlsConfig, nil
+}
+
+func (r *Reconciler) finalizeTLS(ctx context.Context, cr *model.CryostatInstance) error {
+	caCert := resources.NewCryostatCACert(cr)
+	err := r.createOrUpdateCertificate(ctx, caCert, cr.Object)
+	if err != nil {
+		return err
+	}
+	for _, ns := range cr.TargetNamespaces {
+		secret, err := r.GetCertificateSecret(ctx, caCert)
+		if err != nil {
+			return err
+		}
+		namespaceSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secret.Name,
+				Namespace: ns,
+			},
+		}
+		err = r.deleteSecret(ctx, namespaceSecret)
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				r.Log.Info("secret not found, proceeding with deletion", "name", namespaceSecret.Name, "namespace", namespaceSecret.Namespace)
+				return nil
+			}
+			r.Log.Error(err, "failed to delete secret", "name", namespaceSecret.Name, "namespace", namespaceSecret.Namespace)
+			return err
+		}
+		r.Log.Info("deleted secret", "name", namespaceSecret.Name, "namespace", namespaceSecret.Namespace)
+	}
+	return nil
 }
 
 func (r *Reconciler) setCertSecretOwner(ctx context.Context, owner metav1.Object, certs ...*certv1.Certificate) error {
--- a/internal/controllers/certmanager.go
+++ b/internal/controllers/certmanager.go
@@ -56,7 +56,7 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 	}
 
 	// Create CA certificate for Cryostat using the self-signed issuer
-	caCert := resources.NewCryostatCACert(cr)
+	caCert := resources.NewCryostatCACert(cr, cr.InstallNamespace)
 	err = r.createOrUpdateCertificate(ctx, caCert, cr.Object)
 	if err != nil {
 		return nil, err
@@ -120,6 +120,33 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 	err = r.setCertSecretOwner(ctx, cr.Object, certificates...)
 	if err != nil {
 		return nil, err
+	}
+
+	// Copy Cryostat CA secret in each target namespace
+	for _, ns := range cr.TargetNamespaces {
+		secret, err := r.GetCertificateSecret(ctx, caCert)
+		if err != nil {
+			return nil, err
+		}
+		namespaceSecret := secret.DeepCopy()
+		namespaceSecret.Namespace = ns
+	}
+	// Delete any Cryostat CA secrets in target namespaces that are no longer requested
+	for _, ns := range toDelete(cr) {
+		secret, err := r.GetCertificateSecret(ctx, caCert)
+		if err != nil {
+			return nil, err
+		}
+		namespaceSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secret.Name,
+				Namespace: ns,
+			},
+		}
+		err = r.deleteSecret(ctx, namespaceSecret)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	// Get the Cryostat CA certificate bytes from certificate secret
--- a/internal/controllers/certmanager.go
+++ b/internal/controllers/certmanager.go
@@ -134,7 +134,7 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 					Name:      caCert.Spec.SecretName,
 					Namespace: ns,
 				},
-				Type: secret.Type,
+				Type: corev1.SecretTypeOpaque,
 			}
 			err = r.createOrUpdateSecret(ctx, namespaceSecret, cr.Object, func() error {
 				if namespaceSecret.Data == nil {
--- a/internal/controllers/certmanager.go
+++ b/internal/controllers/certmanager.go
@@ -122,32 +122,42 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 		return nil, err
 	}
 
+	secret, err := r.GetCertificateSecret(ctx, caCert)
+	if err != nil {
+		return nil, err
+	}
 	// Copy Cryostat CA secret in each target namespace
 	for _, ns := range cr.TargetNamespaces {
-		secret, err := r.GetCertificateSecret(ctx, caCert)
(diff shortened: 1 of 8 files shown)


mwangggg:
https://github.com/cryostatio/cryostat-operator/actions/runs/6630750805

internal/controllers/certmanager.go
```diff
@@ -68,6 +68,23 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 		return nil, err
 	}
 
+	// Create a Cryostat CA certificate in each target namespace
+	for _, ns := range cr.TargetNamespaces {
+		cryostatCert := resources.NewCryostatCACert(cr, ns)
```
andrewazores, line 73: Does this create a brand new certificate each time it's invoked or will it return a reference to an existing one? I think the idea here is that Cryostat has one cert of its own, and a copy of that same cert should be made available to other applications that may need it. If this creates a brand new (different) cert and places that into the other namespaces then I'm not sure the applications would be able to use that cert for what it's intended for.

ebaron: Right, we want to have one CA certificate per Cryostat. The workflow is like this:
1. Operator creates a cert-manager Certificate for the CA
2. cert-manager responds by creating a Secret containing the CA cert
3. Operator waits for this Secret and sets itself as the owner of it
4. New: Operator copies this Secret into each target namespace, and deletes the Secret from any removed target namespaces

internal/controllers/certmanager.go
```diff
@@ -122,6 +122,33 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
...
 	}
 
+	// Copy Cryostat CA secret in each target namespace
+	for _, ns := range cr.TargetNamespaces {
+		secret, err := r.GetCertificateSecret(ctx, caCert)
+		if err != nil {
+			return nil, err
+		}
+		namespaceSecret := secret.DeepCopy()
+		namespaceSecret.Namespace = ns
```
ebaron, line 132: Hi @mwangggg, this looks good, but you need to tell the Kubernetes client to create this new secret. This part actually sends the request to the Kubernetes API server to [create the secret](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/secret-v1/#create-create-a-secret).

While you can use the [controller-runtime client](https://sdk.operatorframework.io/docs/building-operators/golang/references/client/) directly, I like to use the [`controllerutil.CreateOrUpdate`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate) function instead for things like this. That will create the object if it doesn't exist, or allow you to update it if it does exist. `CreateOrUpdate` takes a callback argument where you set properties that you want to be reflected in the resulting object. In this case, it would just be the secret's `Data` field.

ebaron:
Sorry to complicate things further, but since these secrets will be in different namespaces from the CR, they cannot be owned by the CR. This means we can't leverage Kubernetes [garbage collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/) to clean them up when the CR is deleted. The get around this we can use the existing [finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) to manually delete the secrets when the user attempts to delete the CR.

Here's where we use finalizers in the operator currently: https://github.com/cryostatio/cryostat-operator/blob/3f928cabf2ce5345c170ae08e60522cf36916228/internal/controllers/reconciler.go#L126-L148

We'll want want to add something here like `finalizeTLS` that runs the deletion logic you've added already, but in this case it would be for all target namespaces.

internal/controllers/certmanager.go
```diff
@@ -131,6 +164,37 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
 	return tlsConfig, nil
 }
 
+func (r *Reconciler) finalizeTLS(ctx context.Context, cr *model.CryostatInstance) error {
+	caCert := resources.NewCryostatCACert(cr)
+	err := r.createOrUpdateCertificate(ctx, caCert, cr.Object)
+	if err != nil {
+		return err
+	}
```
ebaron, line 172: This part can be removed, we don't want to create/update the certificate here since it's in the process of being deleted.

internal/controllers/certmanager.go
```diff
@@ -131,6 +164,37 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
...
+		secret, err := r.GetCertificateSecret(ctx, caCert)
+		if err != nil {
+			return err
+		}
+		namespaceSecret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secret.Name,
+				Namespace: ns,
+			},
+		}
```
ebaron, line 183: Same as above, re: `caCert.Spec.SecretName`.

internal/controllers/certmanager.go
```diff
@@ -122,6 +122,39 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
...
+			}
+			return nil
+		})
+	}
+	// Delete any Cryostat CA secrets in target namespaces that are no longer requested
+	for _, ns := range toDelete(cr) {
+		secret, err := r.GetCertificateSecret(ctx, caCert)
+		if err != nil {
+			return nil, err
+		}
```
ebaron, line 145: You can avoid calling GetCertificateSecret here since we don't actually care about the secret's contents. In order to get the Secret's name, we can use `caCert.Spec.SecretName`.

ebaron (commented):
Hi Ming! Just a few more comments. Are you planning to add some tests for these changes? I can help you get started with them

mwangggg:
https://github.com/cryostatio/cryostat-operator/actions/runs/6817721645

internal/controllers/certmanager.go
```diff
@@ -122,6 +122,35 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
...
+		namespaceSecret := secret.DeepCopy()
+		namespaceSecret.Namespace = ns
+		err = r.createOrUpdateSecret(ctx, namespaceSecret, cr.Object, func() error {
+			if secret.StringData == nil {
+				secret.StringData = map[string]string{}
+			}
```
ebaron, line 136: I think this should be `namespaceSecret` inside the closure since that's what we're intending to modify.

We should copy the `Data` from the original secret here too. Just `namespaceSecret.Data = secret.Data` should work. Also we should change this initialization to `Data` instead of `StringData` because of this. The latter is a write-only convenience field that the server Base64 encodes and moves to `Data`.

ebaron (commented):
Hi @mwangggg, sorry it took me a while to get back to this PR. I have some more suggestions below. If you have any questions about them, please let me know.

mwangggg:
https://github.com/cryostatio/cryostat-operator/actions/runs/6907075648

internal/controllers/reconciler.go
```diff
@@ -198,6 +204,13 @@ func (r *Reconciler) reconcile(ctx context.Context, cr *model.CryostatInstance)
 
 	// Set up TLS using cert-manager, if available
 	var tlsConfig *resources.TLSConfig
+
+	// Reconcile RBAC resources for Cryostat
+	err = r.reconcileRBAC(ctx, cr)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
```
ebaron, line 213: Just a small nit, could I bug you to move this block above the `// Set up TLS using cert-manager, if available` comment? That refers to the code after `reconcileRBAC`.

internal/controllers/certmanager.go
```diff
@@ -122,6 +122,45 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
...
+		if ns != cr.InstallNamespace {
+			namespaceSecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      caCert.Spec.SecretName,
+					Namespace: ns,
+				},
+				Type: secret.Type,
+			}
+			err = r.createOrUpdateSecret(ctx, namespaceSecret, cr.Object, func() error {
+				namespaceSecret.Data = secret.Data
```
ebaron, line 140: Could you change this to copy only the `corev1.TLSCertKey` key? You'll need to check if `namespaceSecret.Data == nil` and initialize `namespaceSecret.Data` to an empty map if so.

ebaron (commented):
Oh whoops, I noticed while testing this in OpenShift that we need to be more selective with what we copy from the Secret. cert-manager includes the private key in addition to the certificate in the CA secret. We definitely don't want to copy that all over the cluster.

I've made some suggestions below to remedy this.

internal/controllers/certmanager.go
```diff
@@ -122,6 +122,48 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
...
+					Name:      caCert.Spec.SecretName,
+					Namespace: ns,
+				},
+				Type: secret.Type,
+			}
+			err = r.createOrUpdateSecret(ctx, namespaceSecret, cr.Object, func() error {
+				if namespaceSecret.Data == nil {
+					namespaceSecret.Data = map[string][]byte{}
+				}
+				namespaceSecret.Data["corev1.TLSCertKey"] = secret.Data["corev1.TLSCertKey"]
```
ebaron, line 143: Sorry for confusion, I meant the constant `corev1.TLSCertKey` and not the string `"corev1.TLSCertKey"`.

ebaron (commented):
Hi @mwangggg, sorry for the delay on this. This one thing jumped out at me.

internal/controllers/certmanager.go
mwangggg: ah sorry that's my bad for misunderstanding

internal/controllers/certmanager.go
```diff
@@ -122,6 +122,48 @@ func (r *Reconciler) setupTLS(ctx context.Context, cr *model.CryostatInstance) (
...
+	}
+	// Copy Cryostat CA secret in each target namespace
+	for _, ns := range cr.TargetNamespaces {
+		if ns != cr.InstallNamespace {
+			namespaceSecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      caCert.Spec.SecretName,
+					Namespace: ns,
+				},
+				Type: secret.Type,
```
ebaron, line 137: I tried this out in OpenShift and hit the following error:
```
2023-11-29T18:15:34Z ERROR controllers.ClusterCryostat Failed to set up TLS for Cryostat {"Request.Namespace": "cryostat-operator-system", "Request.Name": "clustercryostat-sample", "error": "Secret \"clustercryostat-sample-ca\" is invalid: data[tls.key]: Required value"}
```

I think this is because the original secret uses the type [`kubernetes.io/tls`](https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets). Apparently with that type of secret, both the certificate and key are mandatory. We can fix this by changing this to:
```suggestion
				Type: corev1.SecretTypeOpaque,
```

Opaque secrets can contain any arbitrary data.

internal/controllers/reconciler_test.go
```diff
@@ -2336,7 +2336,18 @@ func (t *cryostatTestInput) expectCertificates() {
...
+	Expect(secret.StringData).To(Equal(expectedSecret.StringData))
+
+	// Check CA Cert secrets in each target namespace
+	Expect(t.TargetNamespaces).ToNot(BeEmpty())
+	for _, ns := range t.TargetNamespaces {
+		namespaceSecret := t.NewCACertSecret(ns)
+		secret := &corev1.Secret{}
+		err := t.Client.Get(context.Background(), types.NamespacedName{Name: namespaceSecret.Name, Namespace: ns}, secret)
+		Expect(err).ToNot(HaveOccurred())
+		t.checkMetadata(secret, namespaceSecret)
```
ebaron, line 2348: Let's add a check here for the Secret `type`. You could add the Opaque type to `NewCACertSecret` and check for equality here.

internal/controllers/reconciler_test.go
```diff
@@ -2336,7 +2336,18 @@ func (t *cryostatTestInput) expectCertificates() {
...
+		Expect(secret.Data).To(Equal(namespaceSecret.Data))
```
ebaron, line 2349: It would be good to add a regression test here to ensure that only the certificate gets copied but not the private key. To do this, you could add a dummy private key here: https://github.com/cryostatio/cryostat-operator/blob/e8e0bef0bd1c00bcc16d6a238a27c05263085a3b/internal/test/clients.go#L95

Something like: `corev1.TLSPrivateKeyKey: []byte(cert.Name + "-key"),` would work.

Back inside `expectCertificates`, we just need to exclude the install namespace from this check, since the original copy of the Secret will contain the private key. You can do that by adding a `if ns != t.Namespace {` block inside of the `for _, ns := range t.TargetNamespaces {` loop body.

I intended to mention this on my last review, but I must not have submitted the comment. Sorry about that.

ebaron (commented):
Hi Ming, a few more comments below.

ebaron (approved):
Looks good! Thanks for the great work!

PR #677, opened 2023-11-13
fix(insights): set User-Agent header for UHC Auth Proxy

ebaron:
# Welcome to Cryostat! 👋

Related to: https://github.com/cryostatio/cryostat/issues/1763

## Description of the change:
Add a specific User-Agent header to route traffic through the UHC Auth Proxy: https://github.com/RedHatInsights/uhc-auth-proxy/commit/663067c681e0557cedc4358e62f130ab1dff6085

## Motivation for the change:
Traffic not routed through the UHC Auth Proxy requires a JWT Bearer token, which the cloud.openshift.com token is not.

### DIFF
--- a/internal/controllers/insights/test/resources.go
+++ b/internal/controllers/insights/test/resources.go
@@ -17,7 +17,6 @@ package test
 import (
 	"fmt"
 
-	"github.com/cryostatio/cryostat-operator/internal/controllers"
 	"github.com/cryostatio/cryostat-operator/internal/test"
 	configv1 "github.com/openshift/api/config/v1"
 	appsv1 "k8s.io/api/apps/v1"
@@ -31,6 +30,8 @@ type InsightsTestResources struct {
 	*test.TestResources
 	Resources *corev1.ResourceRequirements
 }
+
+const expectedOperatorVersion = "2.5.0-dev"
 
 func (r *InsightsTestResources) NewGlobalPullSecret() *corev1.Secret {
 	config := `{"auths":{"example.com":{"auth":"hello"},"cloud.openshift.com":{"auth":"world"}}}`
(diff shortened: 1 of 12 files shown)


internal/controllers/insights/test/resources.go
```diff
@@ -118,6 +120,12 @@ func (r *InsightsTestResources) NewInsightsProxySecret() *corev1.Secret {
 								"header": "Authorization",
 								"value_type": "plain",
 								"value": "Bearer world"
+							  },
+							  {
+								"op": "set",
+								"header": "User-Agent",
+								"value_type": "plain",
+								"value": "cryostat-operator/%s cluster/abcde"
```
andrewazores, line 128: Why doesn't the `%s` format specifier for the operator version get replaced here?

ebaron: I wasn't sure about this. Ideally we would hardcode the version here, but that would add an additional step to update every time we alter the version. Also if any developers override `IMAGE_VERSION` in the Makefile for their own uses, it would cause the test to fail. What do you think?

andrewazores: Hmm, I see. If there is no straightforward way to inject the actual build version number into the test harness so that the test compares the real expected value to what the controller generates, then I think leaving it like this is okay.

andrewazores (dismissed):

internal/controllers/insights/test/resources.go
ebaron: We could decouple the version used to generate this constant from the version used as an image tag. and then hardcode the version in the test. This would prevent the latter issue where developers might want to . It will still need manual updating when we bump versions, but maybe that's worth it. How does that sound?

andrewazores: If that isn't too much work then it sounds like a reasonable middle ground. At least that way there is a test that validates that this resources normally gets generated properly, and it isn't too annoying in normal development workflows. If it's just a string to bump at release time that's pretty good.

andrewazores (approved):

Issue #583, opened 2023-05-09
[Story] Look into admission webhook as alternative to ClusterCryostat


ebaron:
It would be nice for Cryostat 3.0 if we could combine the Cryostat and ClusterCryostat APIs. One possibility is keeping the namespaced Cryostat CRD and using a validating admission webhook:
https://sdk.operatorframework.io/docs/building-operators/golang/webhook/#1-validating-admission-webhook

We could add a targetNamespaces field to the Cryostat instance. If possible, the webhook would use the user info the [AdmissionRequest](https://pkg.go.dev/k8s.io/api/admission/v1#AdmissionRequest) to do an authorization check against the target namespaces (e.g. can user X create a Cryostat in namespaces Y, Z).

There would still be potential for role binding name conflicts to account for as we already deal with between the ClusterCryostat and Cryostat APIs.

ebaron:
Consider also using the webhook to check for any potential naming conflicts across target namespaces.

ebaron:
Another possibility to avoid name conflicts is to use `GenerateName` on some initial resource, maybe the lock ConfigMap, and store the resulting name in the CR. Then use that generated name consistently for all other objects owned by that CR.

ebaron:
Yet another possibility is to not rely on the names of objects but to use labels or annotations to indicate which CR owns them, and then use a List instead of Get operation.

PR #721, opened 2024-02-02
feat(api): combine Cryostat and ClusterCryostat CRDs

ebaron:
# Welcome to Cryostat! 👋

Fixes: #583 

## Description of the change:

## Motivation for the change:

## How to manually test:
1. *Insert steps here...*
2. *...*


api/v1beta1/cryostat_conversion.go
```diff
@@ -0,0 +1,590 @@
...
+func convertSpecTo(src *CryostatSpec, dst *operatorv1beta2.CryostatSpec) {
+	dst.Minimal = src.Minimal
+	dst.EnableCertManager = src.EnableCertManager
+	dst.TrustedCertSecrets = convertCertSecretsTo(src.TrustedCertSecrets)
+	dst.EventTemplates = convertEventTemplatesTo(src.EventTemplates)
+	dst.StorageOptions = convertStorageOptionsTo(src.StorageOptions)
+	dst.ServiceOptions = convertServiceOptionsTo(src.ServiceOptions)
+	dst.NetworkOptions = convertNetworkOptionsTo(src.NetworkOptions)
+	dst.ReportOptions = convertReportOptionsTo(src.ReportOptions)
+	dst.MaxWsConnections = src.MaxWsConnections
```
andrewazores, line 53: I don't think we have an equivalent `MaxWsConnections` config in 3.0 and I'm not sure if we should bother reimplementing this. I think this is an old holdover from when the UI didn't work too well with concurrent users in the 1.0 days.

api/v1beta1/cryostat_conversion.go
```diff
@@ -0,0 +1,590 @@
...
+	dst.Status.TargetNamespaces = []string{src.Namespace}
+
+	return nil
+}
+
+func convertSpecTo(src *CryostatSpec, dst *operatorv1beta2.CryostatSpec) {
+	dst.Minimal = src.Minimal
+	dst.EnableCertManager = src.EnableCertManager
+	dst.TrustedCertSecrets = convertCertSecretsTo(src.TrustedCertSecrets)
+	dst.EventTemplates = convertEventTemplatesTo(src.EventTemplates)
```
andrewazores, line 48: 3.0 hasn't yet reimplemented custom event templates, but when it does I'm expecting that to be done by using S3 storage rather than direct PVC access. That'll make it more complicated to support the feature from the CR level. Maybe we should drop this like we dropped the FlightRecording CRD?

api/v1beta1/cryostat_conversion.go
```diff
@@ -0,0 +1,590 @@
...
+	convertSpecFrom(&src.Spec, &dst.Spec)
+
+	// Convert existing Status fields
+	convertStatusFrom(&src.Status, &dst.Status)
+
+	return nil
+}
+
+func convertSpecFrom(src *operatorv1beta2.CryostatSpec, dst *CryostatSpec) {
+	dst.Minimal = src.Minimal
```
andrewazores, line 327: I'm in favour of removing the `minimal` config option: https://github.com/cryostatio/cryostat-helm/pull/122

ebaron:
This should be ready for review now. Apologies for the large PR, there's a lot of interconnected pieces when bumping the API version.

andrewazores (commented):
Just a few thoughts that came up. These are all things to address later.

api/v1beta1/cryostat_conversion.go
ebaron: Would there still be a way to do this declaratively? I haven't really used this feature much, but I could see this being useful for GitOps-managed clusters.

andrewazores: I guess there are two apparent routes:
1. the Operator uses an HTTP client and creates the custom event templates on the Cryostat application instances through the API like any other client
2. when we do reimplement custom event templates in 3.0, we do it by direct filesystem access (PVC) like we had before rather than S3. Or, use S3 as a read-write store in addition to a read-only direct filesystem PVC source where additional templates can be mounted.

andrewazores: I think there are other cases of things that could be useful for things like GitOps clusters, like Automated Rules as a CRD. In 3.0 those are defined in a Postgres table so IMO the most sensible way would be for the Operator to talk to Cryostat via the HTTP API there, too.

ebaron: That seems reasonable. These sorts of features would likely never be possible for the Helm Chart, but that could be an advantage of using the operator instead.

andrewazores: Then again, we do have other things that seem to make the most sense as files directly injected into the PVC still, like TLS certs...

ebaron:
https://github.com/cryostatio/cryostat-operator/actions/runs/7874055819

ebaron:
> So the equivalent to the old ClusterCryostat InstallNamespace is now just that the Operator will deploy Cryostat into the Namespace in which the Cryostat CR is created, right?

That's right.

> 
> When I try to create a new Cryostat CR through the OpenShift UI:
> 
> image
> 
> The TargetNamespaces list is just a text list, no dropdown selector of namespaces to choose from. Is that expected now?

Thanks for noticing this. We had a Kustomize patch for the ClusterCryostat CRD to add the descriptors. It didn't seem to be possible to define the x-descriptor using markers for the array elements, as opposed to the array itself.

I updated the patch and also the initialization resource (the UI prompt to create a Cryostat) from v1beta1 to v1beta2.

andrewazores (approved):

Issue #730, opened 2024-02-13
[Story] Remove CommandConfig from NetworkOptions


ebaron:
This has long been deprecated and no longer has any effect. (https://github.com/cryostatio/cryostat-operator/issues/201)

PR #743, opened 2024-02-28
chore(api): remove CommandConfig from NetworkOptions

mwangggg:
# Welcome to Cryostat! 👋

Fixes: #730


mwangggg:
https://github.com/cryostatio/cryostat-operator/actions/runs/8073776021

andrewazores (approved):

api/v1beta1/cryostat_types.go
```diff
@@ -342,13 +342,6 @@ type NetworkConfigurationList struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	CoreConfig *NetworkConfiguration `json:"coreConfig,omitempty"`
-	// Specifications for how to expose the Cryostat command service,
-	// which serves the WebSocket command channel.
-	// +optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:hidden"}
-	//
-	// Deprecated: CommandConfig is no longer used.
-	CommandConfig *NetworkConfiguration `json:"commandConfig,omitempty"`
```
ebaron, line 351: I would leave this in v1beta1 for API compatibility. We'll just drop the value when converting to v1beta2 as you've already done.

ebaron (commented):
Thanks for doing this @mwangggg! Could I get you do add a test case for this like this one? https://github.com/cryostatio/cryostat-operator/blob/46c70ce9254c99fa9a8845534637b422470dd67c/api/v1beta1/cryostat_conversion_test.go#L80-L81

I don't think we have any existing test function to create an CR with CommandConfig set, but the existing helper functions in the call graph here might make it simple:
https://github.com/cryostatio/cryostat-operator/blob/46c70ce9254c99fa9a8845534637b422470dd67c/internal/test/conversion.go#L82-L84

api/v1beta1/cryostat_conversion_test.go
```diff
@@ -77,8 +77,11 @@ var _ = Describe("Cryostat", func() {
 })
 
 func tableEntriesTo() []TableEntry {
-	return append(tableEntries(), Entry("WS connections", (*test.TestResources).NewCryostatWithWsConnectionsSpecV1Beta1,
-		(*test.TestResources).NewCryostat))
+	return append(tableEntries(),
+		Entry("WS connections", (*test.TestResources).NewCryostatWithWsConnectionsSpecV1Beta1,
+			(*test.TestResources).NewCryostat),
+		Entry("command config", (*test.TestResources).NewCryostatWithCommandConfigV1Beta1,
+			(*test.TestResources).NewCryostat))
```
ebaron, line 84: This test is failing because the conversion leaves an empty NetworkConfigurationList, which is expected. Once serialized to JSON/YAML it will be removed. Of the possibilities to fix this test, the easiest is probably to include the Core and Grafana configs as well. Basically, test that it becomes `NewCryostatWithIngress`.

mwangggg: sorry I didn't get the change to test it yet!

internal/test/conversion.go
```diff
@@ -479,13 +479,25 @@ func (r *TestResources) NewCryostatWithWsConnectionsSpecV1Beta1() *operatorv1bet
...
+	coreSVC := r.NewCryostatService()
+	coreIng := r.newNetworkConfigurationV1Beta1(coreSVC.Name, coreSVC.Spec.Ports[0].Port)
+	coreIng.Annotations["custom"] = "annotation"
+	coreIng.Labels["custom"] = "label"
+
+	grafanaSVC := r.NewGrafanaService()
+	grafanaIng := r.newNetworkConfigurationV1Beta1(grafanaSVC.Name, grafanaSVC.Spec.Ports[0].Port)
+	grafanaIng.Annotations["grafana"] = "annotation"
+	grafanaIng.Labels["grafana"] = "label"
+
```
ebaron, line 491: Could I suggest calling `NewCryostatWithIngressV1Beta1` and then augmenting it with the command config before returning? It would be a bit less duplication.

ebaron (approved):
Looks good, thanks Ming!

Issue #749, opened 2024-03-05
[Task] Log workload events for more Scorecard test failures


ebaron:
I've encountered some occasional 503 - Service Unavailable errors when running the `cryostat-recording` Scorecard test. This appears to be after the pod is passing it's readiness checks. This could be caused by a container restarting. An easy way to tell would be to use the existing `logWorkloadEvents` method on any failure after a Cryostat deployment has been created:
https://github.com/cryostatio/cryostat-operator/blob/cfcbfc74fab65b78bf5fb63a8c5abe658d65fe6a/internal/test/scorecard/common_utils.go#L95

PR #759, opened 2024-03-12
test(scorecard): logWorkLoadEvent for cryostat-recording errors

mwangggg:
# Welcome to Cryostat! 👋

Fixes: #749


internal/test/scorecard/tests.go
```diff
@@ -177,6 +178,10 @@ func CryostatRecordingTest(bundle *apimanifests.Bundle, namespace string, openSh
 	}
 	target, err := apiClient.Targets().Create(context.Background(), targetOptions)
 	if err != nil {
+		logErr := logWorkloadEvents(r, client, namespace, CryostatRecordingTestName)
+		if logErr != nil {
+			return fail(*r, fmt.Sprintf("failed to create a target: %s", logErr))
+		}
```
tthvo, line 184: Would it make sense to `defer` the `logWorkloadEvents`, but before deleting Cryostat CR?

tthvo: because currently, the error might appear as:

```
failed to create a target: <some-error-during-log>
```

so, its might not be clear which error causes the failed API request...

tthvo (commented):
Looks nice :)) Did a mock error and got:

```
Image:      quay.io/thvo/cryostat-operator-scorecard:2.5.0-20240311082913
Labels:
	"test":"cryostat-recording"
	"suite":"cryostat"
Results:
	Name: cryostat-recording
	State: pass

	Log:
		deployment cryostat-recording is not yet found
		deployment cryostat-recording is not yet found
		deployment cryostat-recording is not yet found
		deployment cryostat-recording is not yet found
		deployment cryostat-recording is not yet found
		deployment cryostat-recording is not yet found
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is not yet available
		deployment cryostat-recording is available
		application is available at https://testing.cryostat
		application is not yet reachable at https://testing.cryostat
		application is not yet reachable at https://testing.cryostat
		application is ready at https://testing.cryostat
		created a custom target: &{ConnectUrl:service:jmx:rmi:///jndi/rmi://localhost:0/jmxrmi Alias:customTarget}
		created stored credential with match expression: target.alias=="customTarget"
		created a recording: &{DownloadURL:https://testing.cryostat:443/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/recordings/scorecard_test_rec ReportURL:https://testing.cryostat:443/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/reports/scorecard_test_rec Id:1 Name:scorecard_test_rec StartTime:1710145895440 State:RUNNING Duration:0 Continuous:true ToDisk:true MaxSize:0 MaxAge:0}
		current list of recordings: [{DownloadURL:https://testing.cryostat:443/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/recordings/scorecard_test_rec ReportURL:https://testing.cryostat:443/api/v1/targets/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/reports/scorecard_test_rec Id:1 Name:scorecard_test_rec StartTime:1710145895440 State:RUNNING Duration:0 Continuous:true ToDisk:true MaxSize:0 MaxAge:0}]
		archived the recording scorecard_test_rec at: customTarget_scorecard_test_rec_20240311T083205Z.jfr
		current list of archives: [{Name:customTarget_scorecard_test_rec_20240311T083205Z.jfr DownloadUrl:https://testing.cryostat:443/api/beta/recordings/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/customTarget_scorecard_test_rec_20240311T083205Z.jfr ReportUrl:https://testing.cryostat:443/api/beta/reports/service:jmx:rmi:%2F%2F%2Fjndi%2Frmi:%2F%2Flocalhost:0%2Fjmxrmi/customTarget_scorecard_test_rec_20240311T083205Z.jfr Metadata:{Labels:map[template.name:ALL template.type:TARGET]} Size:2777124}]
		generated report for the recording scorecard_test_rec: map[Allocations.class:map[evaluation:map[explanation:Frequently allocated types are good places to start when trying to reduce garbage collections. Look at where the most common types are being allocated to see if many instances are created along the same call path. Try to reduce the number of instances created by invoking the most commonly taken paths less. suggestions:[] summary:The most allocated type is likely ''byte[]'', most commonly allocated by: org.openjdk.jmc.flightrecorder.internal.parser.v1.StructTypes$JfrMethod@b73f018f] name:Allocated Classes score:6.170517666010364 topic:heap] Allocations.thread:map[evaluation:map[explanation:Many allocations performed by the same thread might indicate a problem in a multi-threaded program. Look at the stack traces for the thread with the highest allocation rate. See if the allocation rate can be brought down, or balanced among the active threads. suggestions:[] summary:The most allocations were likely done by thread ''vert.x-worker-thread-12'' at: org.openjdk.jmc.flightrecorder.internal.parser.v1.StructTypes$JfrMethod@b73f018f] name:Threads Allocating score:3.498549753837512 topic:java_application] ApplicationHalts:map[evaluation:map[explanation:The highest ratio of application halts to execution time was 0.279 % during 3/11/2024, 8:31:35.000 AM – 8:32:35 AM. 19.3 % of the halts were for reasons other than GC. The halts ratio for the entire recording was 0.547 %. 19.3 % of the total halts were for reasons other than GC. suggestions:[] summary:Application efficiency was not highly affected by halts.] name:Application Halts score:1.3930892333333333 topic:java_application] BufferLost:map[evaluation:map[suggestions:[] summary:No Flight Recorder buffers were lost during the recording.] name:Lost Flight Recorder Buffers score:0 topic:recording] BytecodeVerification:map[evaluation:map[suggestions:[] summary:The application ran with bytecode verification enabled.] name:Bytecode Verification score:0 topic:jvm_information] ClassLeak:map[evaluation:map[suggestions:[]] name:Class Leak score:-1 topic:classloading] ClassLoading:map[evaluation:map[suggestions:[] summary:No significant time was spent loading new classes during this recording.] name:Class Loading Pressure score:0 topic:classloading] CodeCache:map[evaluation:map[suggestions:[]] name:Code Cache score:-1 topic:code_cache] CompareCpu:map[evaluation:map[explanation:The application performance can be affected when the machine is under heavy load and there are other processes that use CPU or other resources on the same computer. To profile representatively or get higher throughput, shut down other resource intensive processes running on the machine. suggestions:[] summary:An average CPU load of 12 % was caused by other processes for during 3/11/2024, 8:31:35.000 AM – 8:32:05 AM.] name:Competing CPU Ratio Usage score:2.5978083024958645 topic:processes] CompressedOops:map[evaluation:map[suggestions:[] summary:The settings for Compressed Oops were OK.] name:Compressed Oops score:0 topic:gc_configuration] ContextSwitch:map[evaluation:map[suggestions:[] summary:The program did not context switch excessively during the recording.] name:Context Switches score:1.5 topic:lock_instances] DMSIncident:map[evaluation:map[suggestions:[]] name:DMS Incidents score:-1 topic:DMS] DebugNonSafepoints:map[evaluation:map[suggestions:[] summary:DebugNonSafepoints was implicitly enabled in the JVM version used to create this recording.] name:DebugNonSafepoints score:0 topic:jvm_information] DiscouragedVmOptions:map[evaluation:map[suggestions:[] summary:No problems were found with the VM options.] name:Discouraged VM Options score:0 topic:jvm_information] DumpReason:map[evaluation:map[suggestions:[]] name:Exceptional Dump Reason score:-1 topic:recording] DuplicateFlags:map[evaluation:map[suggestions:[] summary:There were no duplicate JVM flags on the command line.] name:Duplicated Flags score:0 topic:jvm_information] Errors:map[evaluation:map[explanation:3 errors were thrown in total. The most common error was ''java.lang.NoSuchMethodError'', which was thrown 3 times. Investigate the thrown errors to see if they can be avoided. Errors indicate that something went wrong with the code execution and should never be used for flow control. suggestions:[] summary:The program generated an average of 3 errors per minute during 3/11/2024, 8:31:35.000 AM – 8:32:35 AM.] name:Thrown Errors score:2.5 topic:exceptions] Exceptions:map[evaluation:map[explanation:Throwing exceptions is more expensive than normal code execution, which means that they should only be used for exceptional situations. Investigate the thrown exceptions to see if any of them can be avoided with a non-exceptional control flow. suggestions:[] summary:The program generated 66.8 exceptions per second during 3/11/2024, 8:32:05.720.000 AM – .960.] name:Thrown Exceptions score:0.33386043497023465 topic:exceptions] Fatal Errors:map[evaluation:map[suggestions:[]] name:Fatal Errors score:-1 topic:jvm_information] FewSampledThreads:map[evaluation:map[suggestions:[]] name:Parallel Threads score:-1 topic:java_application] FileRead:map[evaluation:map[suggestions:[] summary:No long file read pauses were found in this recording (the longest was 52.843 μs).] name:File Read Peak Duration score:0.0006605375000000001 topic:file_io] FileWrite:map[evaluation:map[suggestions:[] summary:No long file write pauses were found in this recording (the longest was 244.260 μs).] name:File Write Peak Duration score:0 topic:file_io] FlightRecordingSupport:map[evaluation:map[suggestions:[] summary:The JVM version used for this recording has full Flight Recorder support.] name:Flight Recording Support score:0 topic:jvm_information] FullGc:map[evaluation:map[explanation:At least one Full, Stop-The-World Garbage Collection occurred during this recording. For the CMS and G1 collectors, Full GC events are a strong negative performance indicator. Tunable GC parameters can be used to allow the collector to operate in concurrent mode, avoiding Stop-The-World pauses and increasing GC and application performance. suggestions:[] summary:Full GC detected.] name:G1/CMS Full Collection score:75 topic:garbage_collection] GarbageCollectionInfoRule:map[evaluation:map[suggestions:[]] name:Garbage Collection Info score:0 topic:garbage_collection] GcFreedRatio:map[evaluation:map[suggestions:[] summary:Only 8 heap summary events were found, this rule requires at least 10 events to be able to calculate a relevant result. This likely means that only a few garbage collections occurred during the recording. Having few garbage collections is generally a good sign.] name:GC Freed Ratio score:0 topic:heap] GcLocker:map[evaluation:map[suggestions:[] summary:No GCs were affected by the GC Locker.] name:GCs Caused by GC Locker score:0 topic:garbage_collection] GcOptions:map[evaluation:map[suggestions:[] summary:No problems were found with the GC configuration.] name:GC Setup score:0 topic:jvm_information] GcPauseRatio:map[evaluation:map[explanation:The highest ratio between garbage collection pauses and execution time was 0.225 % during 3/11/2024, 8:31:35.000 AM – 8:32:35 AM. The garbage collection pause ratio of the entire recording was 0.442 %. solution:Pause times may be reduced by increasing the heap size or by trying to reduce allocation. suggestions:[] summary:Application efficiency was not highly affected by GC pauses.] name:GC Pauses score:1.1238405 topic:garbage_collection] GcStall:map[evaluation:map[suggestions:[] summary:No indications that the garbage collector could not keep up with the workload were detected.] name:GC Stall score:0 topic:garbage_collection] HeapContent:map[evaluation:map[explanation:If the heap usage needs to be reduced, then this would be a good place to start. suggestions:[] summary:Most of the heap was used by only a few classes.] name:Heap Content score:90.76165173264532 topic:heap] HeapDump:map[evaluation:map[suggestions:[]] name:Heap Dump score:-1 topic:heap] HeapInspectionGc:map[evaluation:map[explanation:Performing heap inspection garbage collections may be a problem since they usually take a lot of time. suggestions:[] summary:The JVM performed 4 heap inspection garbage collections.] name:GCs Caused by Heap Inspection score:59.77379177936154 topic:garbage_collection] HighGc:map[evaluation:map[explanation:The time spent performing garbage collection may be reduced by increasing the heap size or by trying to reduce allocation.
		To improve rule accuracy and/or get more details for further investigation, it is recommended to enable the following event types: . suggestions:[] summary:The JVM was paused for 71.7 % during 3/11/2024, 8:32:05.625.000 AM – .720] name:GC Pressure score:14.479485296736971 topic:heap] HighJvmCpu:map[evaluation:map[explanation:The sampling period for the 'CPU Load' events was set to Every Chunk, which is too high for CPU load related rules to work. suggestions:[] summary:This recording has a high sampling period for 'CPU Load' events.] name:High JVM CPU Load score:25 topic:java_application] IncreasingLiveSet:map[evaluation:map[explanation:Perform a dump with the 'Trace Paths to GC Roots' option enabled to enable a more detailed analysis of the potential memory leak. suggestions:[] summary:The live set on the heap seems to increase with a speed of about 34.1 KiB per second during the recording.There is no particular class that seems to be leaking more than any other.] name:Heap Live Set Trend score:0.6225772765246449 topic:memoryleak] IncreasingMetaSpaceLiveSet:map[evaluation:map[suggestions:[] summary:The class data does not seem to increase during the recording.] name:Metaspace Live Set Trend score:5.975338591383805 topic:garbage_collection] JavaBlocking:map[evaluation:map[explanation:The following regular expression was used to exclude threads from this rule: ''(.*weblogic\.socket\.Muxer.*)'' suggestions:[] summary:No excessive problems with lock contention found.] name:Java Blocking score:0.03556824436957706 topic:lock_instances] JfrPeriodicEventsFix:map[evaluation:map[suggestions:[] summary:The version of Java you are running is not affected by a performance issue related to periodic events.] name:JFR Periodic Events Fix score:0 topic:jvm_information] LongGcPause:map[evaluation:map[explanation: suggestions:[] summary:The longest GC pause was 37.175 ms.] name:GC Pause Peak Duration score:1.1254219796511695 topic:garbage_collection] LowOnPhysicalMemory:map[evaluation:map[explanation:The maximum amount of memory used was 29.4 GiB. This is 94.4 % of the 31.1 GiB of physical memory available. Having little free memory may lead to swapping, which is very expensive. To avoid this, either decrease the memory usage or increase the amount of available memory. suggestions:[] summary:The maximum amount of used memory was 94.4 % of the physical memory available.] name:Free Physical Memory score:73.4139256480875 topic:heap] ManagementAgent:map[evaluation:map[solution:See the [Java Monitoring and Management Guide](https://docs.oracle.com/javase/8/docs/technotes/guides/management/agent.html) for more information about how to configure the management agent. suggestions:[]] name:Discouraged Management Agent Settings score:-1 topic:jvm_information] ManyRunningProcesses:map[evaluation:map[explanation:At 3/11/24, 8:32:05.612 AM, a total of 1 other processes were running on the host machine that this Flight Recording was made on. solution:If this is a server environment, it may be good to only run other critical processes on that machine. suggestions:[] summary:1 processes were running while this Flight Recording was made.] name:Competing Processes score:0.20309488837692125 topic:processes] MetaspaceOom:map[evaluation:map[suggestions:[]] name:Metaspace Out of Memory score:-1 topic:garbage_collection] MethodProfiling:map[evaluation:map[suggestions:[]] name:Method Profiling score:-1 topic:method_profiling] Options:map[evaluation:map[suggestions:[] summary:No undocumented, deprecated or non-recommended option flags were detected.] name:Command Line Options Check score:0 topic:jvm_information] OverAggressiveRecordingSetting:map[evaluation:map[explanation:Event types without threshold can lead to quite a lot of events being generated, possibly translating to higher overhead. If this was not intended, please check the settings in the template for future recordings. suggestions:[] summary:These following event types had no threshold: 'Java Monitor Blocked', 'Java Thread Park'] name:Discouraged Recording Settings score:25 topic:recording] PasswordsInArguments:map[evaluation:map[suggestions:[] summary:The recording does not seem to contain passwords in the application arguments.] name:Passwords in Java Arguments score:0 topic:jvm_information] PasswordsInEnvironment:map[evaluation:map[explanation:The following suspicious environment variables were found in this recording: CRYOSTAT_JDBC_PASSWORD, CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD. The following regular expression was used to exclude strings from this rule: ''(passworld|passwise)''. solution:If you wish to keep having passwords in your environment variables, but want to be able to share recordings without also sharing the passwords, please disable the ''Initial Environment Variable'' event. suggestions:[] summary:The environment variables in the recording may contain passwords.] name:Passwords in Environment Variables score:75 topic:environment_variables] PasswordsInSystemProperties:map[evaluation:map[explanation:The following suspicious system properties were found in this recording: javax.net.ssl.keyStorePassword,javax.net.ssl.trustStorePassword,com.sun.management.jmxremote.password.file. The following regular expression was used to exclude strings from this rule: ''(passworld|passwise)''. solution:If you wish to keep having passwords in your system properties, but want to be able to share recordings without also sharing the passwords, please disable the ''Initial System Property'' event. suggestions:[] summary:The system properties in the recording may contain passwords.] name:Passwords in System Properties score:75 topic:system_properties] PrimitiveToObjectConversion:map[evaluation:map[explanation:
		The most common object type that primitives are converted into is ''java.lang.Long'', which causes 34.8 KiB to be allocated. The most common call site is ''javax.management.openmbean.CompositeData sun.management.MemoryUsageCompositeData.getCompositeData():62''.
		Conversion from primitives to the corresponding object types can either be done explicitly, or be caused by autoboxing. If a considerable amount of the total allocation is caused by such conversions, consider changing the application source code to avoid this behavior. Look at the allocation stack traces to see which parts of the code to change. This rule finds the calls to the valueOf method for any of the eight object types that have primitive counterparts. suggestions:[] summary:0.136 % of the total allocation (25 MiB) is caused by conversion from primitive types to object types. The most common object type that primitives are converted into is ''java.lang.Long''.] name:Primitive To Object Conversion score:0.16960527202231673 topic:heap] ProcessStarted:map[evaluation:map[suggestions:[]] name:Process Started score:-1 topic:processes] SocketRead:map[evaluation:map[explanation:The longest recorded socket read took 24.873 s to read 5 B from the host at 10.244.0.28. Average time of recorded IO: 469.826 ms. Total time of recorded IO: 24.901 s. Total time of recorded IO for the host 10.244.0.28: 24.900 s. Note that there are some socket read patterns with high duration reads that we consider to be normal and are therefore excluded. Such patterns include JMX RMI communication and MQ series. suggestions:[] summary:There are long socket read pauses in this recording (the longest is 24.873 s).] name:Socket Read Peak Duration score:75 topic:socket_io] SocketWrite:map[evaluation:map[explanation:Note that there are some socket write patterns with high duration writes that we consider to be normal and are therefore excluded. Such patterns include JMX RMI communication. suggestions:[] summary:No long socket write pauses were found in this recording (the longest was 227.807 μs).] name:Socket Write Peak Duration score:0.020709727272727272 topic:socket_io] StackdepthSetting:map[evaluation:map[explanation:The Flight Recorder is configured with a maximum captured stack depth of 64. 4.79 % of all traces were larger than this option, and were therefore truncated. If more detailed traces are required, increase the ''-XX:FlightRecorderOptions=stackdepth=<value>'' value.
		Events of the following types have truncated stack traces: org.openjdk.jmc.flightrecorder.rules.jdk.general.StackDepthSettingRule$StackDepthTruncationData@7dbbb7eb,org.openjdk.jmc.flightrecorder.rules.jdk.general.StackDepthSettingRule$StackDepthTruncationData@35026246,org.openjdk.jmc.flightrecorder.rules.jdk.general.StackDepthSettingRule$StackDepthTruncationData@bf69b93,org.openjdk.jmc.flightrecorder.rules.jdk.general.StackDepthSettingRule$StackDepthTruncationData@d35416f,org.openjdk.jmc.flightrecorder.rules.jdk.general.StackDepthSettingRule$StackDepthTruncationData@2d11bbcd,org.openjdk.jmc.flightrecorder.rules.jdk.general.StackDepthSettingRule$StackDepthTruncationData@b27ec7b,org.openjdk.jmc.flightrecorder.rules.jdk.general.StackDepthSettingRule$StackDepthTruncationData@3432d48f suggestions:[] summary:Some stack traces were truncated in this recording.] name:Stackdepth Setting score:25 topic:jvm_information] StringDeduplication:map[evaluation:map[explanation:String deduplication is enabled using the JVM flag '-XX:+UseStringDeduplication'. This flag can be used together with the G1 garbage collector in JDK 8u20 or later, or with the Shenandoah garbage collector.
		To validate if this gives a performance improvement for your application, create flight recordings both with and without string deduplication. For the run with string deduplication enabled, also enable statistics with '-XX:+PrintStringDeduplicationStatistics' for JDK 8 or '-Xlog:stringdedup*=debug' for JDK 9. Check if the heap live set decrease in the recording with string deduplication enabled is larger than the size of the string deduplication metadata table. The size of the metadata table is printed in the statistics output as 'Table/Memory Usage: XX MB'
		You can read more about string deduplication in the java options documentation or in [JEP 192](https://openjdk.java.net/jeps/192). suggestions:[] summary:Approximately 1,418 % of the live set consists of the internal array type of strings (''byte[]'' for this JDK version).
		The heap is around 0.661 % full. There is likely no big benefit from enabling string deduplication.] name:String Deduplication score:8.166845965026555 topic:heap] SystemGc:map[evaluation:map[suggestions:[] summary:No garbage collections were caused by System.gc().] name:GCs Caused by System.gc() score:0 topic:garbage_collection] TlabAllocationRatio:map[evaluation:map[solution:Allocating objects outside of Thread Local Allocation Buffers (TLABs) is more expensive than allocating inside TLABs. This may be acceptable if the individual allocations are intended to be larger than a reasonable TLAB. It may be possible to avoid this by decreasing the size of the individual allocations. There are some TLAB related JVM flags that you can experiment with, but it is usually better to let the JVM manage TLAB sizes automatically. suggestions:[] summary:The program allocated 8.1 % of the memory outside of TLABs.] name:TLAB Allocation Ratio score:11.38145240564841 topic:tlab] VMOperations:map[evaluation:map[suggestions:[] summary:No excessively long VM operations were found in this recording (the longest was 37.199 ms).] name:VMOperation Peak Duration score:0.929973675 topic:vm_operations] biasedLockingRevocation:map[evaluation:map[suggestions:[]] name:Biased Locking Revocation score:-1 topic:biased_locking] biasedLockingRevocationPause:map[evaluation:map[suggestions:[] summary:No revocation of biased locks found.] name:Biased Locking Revocation Pauses score:0 topic:vm_operations]]
		stopped the recording: scorecard_test_rec
		deleted the recording: scorecard_test_rec
		current list of recordings: []
                WORKLOAD EVENTS: 
                deployment cryostat-recording conditions:
                    Available == True, MinimumReplicasAvailable: Deployment has minimum availability.
                    Progressing == True, NewReplicaSetAvailable: ReplicaSet "cryostat-recording-6868989bd8" has successfully progressed.
                deployment cryostat-recording warning events:
                replica set cryostat-recording-6868989bd8 conditions:
                replica set cryostat-recording-6868989bd8 warning events:
                pod cryostat-recording-6868989bd8-m27l7 phase: Running
                pod cryostat-recording-6868989bd8-m27l7 conditions:
                    Initialized == True, : 
                    Ready == True, : 
                    ContainersReady == True, : 
                    PodScheduled == True, : 
                pod cryostat-recording-6868989bd8-m27l7 warning events:
```

internal/test/scorecard/common_utils.go
```diff
@@ -177,6 +177,16 @@ func logEvents(r *scapiv1alpha3.TestResult, client *CryostatClientset, namespace
 	return nil
 }
 
+func deferLogWorkloadEvents(r *scapiv1alpha3.TestResult, client *CryostatClientset, namespace string, name string) {
```
ebaron, line 180: Maybe rename this to something like `logWorkloadEventsOnError` or something similar. It just reads a bit strangely with `defer defer...`

ebaron (commented):
Hey Ming, great implementation of this. Using defer and checking the Results object for errors makes this really easy to add to future tests. I'd like to see this added to all other tests too, including checking the reports deployment where applicable.

What do you think of this idea? 
1. Add this `defer` to the start of each test.
2. Call logWorkloadEvents in a loop for `name` and `name + "-reports"`
3. Modify logWorkloadEvents to simply return nil if the deployment doesn't exist (use kerrors.IsNotFound to easily check): https://github.com/cryostatio/cryostat-operator/blob/6498b0f92236ceb6374cfeeedfc230c73ac0e67f/internal/test/scorecard/common_utils.go#L102-L105

ebaron (commented):
I think the defer should be moved before the first `return fail...` in each test, otherwise we might miss logging opportunities.

ebaron:
> > I think the defer should be moved before the first `return fail...` in each test, otherwise we might miss logging opportunities.
> 
> When I defer LogWorkloadEvents before defering cleanupCryostat, I get an error when trying to get the Cryostat deployment since it's already been deleted...

Ah defers are in LIFO order, so Cryostat will be gone every time... What about if we just call LogWorkloadEvents from cleanupCryostat before deleting?

mwangggg:
> > > I think the defer should be moved before the first `return fail...` in each test, otherwise we might miss logging opportunities.
> > 
> > 
> > When I defer LogWorkloadEvents before defering cleanupCryostat, I get an error when trying to get the Cryostat deployment since it's already been deleted...
> 
> Ah defers are in LIFO order, so Cryostat will be gone every time... What about if we just call LogWorkloadEvents from cleanupCryostat before deleting?

I like this idea, but I still won't be able to defer the logs before the first return fail in each test.

internal/test/scorecard/common_utils.go
```diff
@@ -485,18 +500,25 @@ func updateAndWaitTillCryostatAvailable(cr *operatorv1beta1.Cryostat, resources
 	return err
 }
 
-func cleanupCryostat(r *scapiv1alpha3.TestResult, client *CryostatClientset, name string, namespace string) {
+func cleanupAndLogs(r *scapiv1alpha3.TestResult, client *CryostatClientset, name string, namespace string, ch *chan *ContainerLog) {
```
tthvo, line 503: I think here we can use `chan *ContainerLog`. The channel itself is already a reference so we can pass `nil` if not available.

tthvo: Maybe name the parameter a bit more descriptive? Something like: `logChannel`?

internal/test/scorecard/tests.go
```diff
@@ -146,24 +146,23 @@ func CryostatConfigChangeTest(bundle *apimanifests.Bundle, namespace string, ope
...
 	cr, err := createAndWaitTillCryostatAvailable(newCryostatCR(CryostatRecordingTestName, namespace, !tr.OpenShift), tr)
 	if err != nil {
 		return fail(*r, fmt.Sprintf("failed to determine application URL: %s", err.Error()))
 	}
-	ch, err := StartLogs(tr.Client.Clientset, cr)
+	ch, err = StartLogs(tr.Client.Clientset, cr)
 	if err != nil {
-		return fail(*r, fmt.Sprintf("failed to retrieve logs for the application: %s", err.Error()))
+		r.Log += fmt.Sprintf("failed to retrieve logs for the application: %s", err.Error())
 	}
```
tthvo, line 165: The channel returned from `StartLogs` should be used to collect logs instead? Maybe call `defer cleanupAndLogs` after creating CR?

mwangggg: the point of all this refractoring is to get cleanupAndLogs right after the first return fail. The channel returned from StartLogs is used to collect the logs- that's why I pass in the pointer `*chan *ContainerLog` through `cleanupAndLogs` to answer your comment above^. This way, cleanupAndLogs has access to the updated channel after collecting the logs.

mwangggg: It took a while for me to figure out a solution that allows `cleanupAndLogs` to be called before creating Cryostat CR and allows both workload and container logs to work :cry: please let me know if you have a better alternative

mwangggg: you should still be able to see the container logs if you run :)

tthvo: Ahh I see. No I think it reasonable to me. Maybe, just declare the channel as below so it does not look like its using the initial one.


```go
var ch chan *ContainerLog
```

tthvo: Otherwise, we can attach and initialize the channel to the `TestResources` and pass it in the deferred cleanup func. WDUT?

internal/test/scorecard/tests.go
```diff
@@ -286,14 +285,15 @@ func CryostatRecordingTest(bundle *apimanifests.Bundle, namespace string, openSh
...
-func CryostatReportTest(bundle *apimanifests.Bundle, namespace string, openShiftCertManager bool) scapiv1alpha3.TestResult {
+func CryostatReportTest(bundle *apimanifests.Bundle, namespace string, openShiftCertManager bool) (result scapiv1alpha3.TestResult) {
 	tr := newTestResources(CryostatReportTestName)
 	r := tr.TestResult
 
 	err := setupCRTestResources(tr, openShiftCertManager)
 	if err != nil {
 		return fail(*r, fmt.Sprintf("failed to set up %s test: %s", CryostatReportTestName, err.Error()))
 	}
+	defer cleanupAndLogs(&result, tr.Client, CryostatReportTestName, namespace, nil)
```
tthvo, line 296: Call before first fail?

ebaron (dismissed):

ebaron (approved):
Looks good!

ebaron:
@Mergifyio backport cryostat3

Issue #710, opened 2024-01-17
[Story] Deploy Cryostat 3.0


Josh-Matsuoka:
### Describe the feature

In line with Helm (https://github.com/cryostatio/cryostat-helm/issues/110) the Operator should be updated to deploy development builds of Cryostat 3.0

This can be roughly broken down into the following sub tasks:

- Deploy Cryostat 3.0 with hardcoded database and storage containers, no auth
- Add in the auth proxy, put storage behind the auth proxy
- Add in grafana, put it behind the auth proxy
- Re-enable basic auth

andrewazores:
Initial round of this is done now. In the future the database and storage containers should be pulled out into their own Pods as well.

PR #799, opened 2024-04-23
feat(auth): deploy openshift auth proxy

andrewazores:
# Welcome to Cryostat! 👋

Related to #710
Based on (merged in) #800

## Description of the change:
Deploys `openshift-oauth-proxy` in the Pod and points the existing Service (and Route or Ingress) at it, rather than at Cryostat directly. The auth proxy is always configured to require one particular set of Roles/permissions for both interactive users as well as programmatic clients, and always delegates to OpenShift SSO. In this PR it cannot be disabled and it cannot be augmented with htpasswd Basic auth. It also does not handle TLS.

## Motivation for the change:
In Cryostat 3.0 there should always be an auth proxy placed in front of the operand containers, since in the new architecture the authz responsibility is always delegated to such an auth proxy and never performed by the Cryostat container. The OpenShift OAuth Proxy with SubjectAccessReview and TokenReview configuration in this manner is very close to what Cryostat <= 2.4 did and generally allows users and clients to authenticate and access the application and API with the same permissions as before.

## How to manually test:
1. `OPERATOR_IMG=quay.io/andrewazores/cryostat-operator:3.0.0-openshift-oauth-proxy-3 DISABLE_SERVICE_TLS=true DEPLOY_NAMESPACE=$(oc project -q) make install deploy`
2. Create a Cryostat CR with `enableCertManager: false`
3. Wait for deployment to become ready, then open the route URL in your browser
4. Try to log in with an unprivileged user, ex. `developer`. You should see a "permission denied" page. Click "Log in" again and this time log in with a privileged user, ex. `kubeadmin`. You should get in to the application and be able to perform some basic functionality ex view Security certs/credentials lists (many still broken due to https://github.com/openshift/oauth-proxy/issues/272)
5. Log out and verify that repeating step 4 behaves the same way again
6. In a terminal, use http/curl/wget: `http --follow cryostat-sample-cryostat3-helm.apps-crc.testing/api/v3/tls/certs` produces a login page response, while `http --follow --auth-type=bearer --auth=$(oc whoami -t) cryostat-sample-cryostat3-helm.apps-crc.testing/api/v3/tls/certs` returns the list response (assuming you are logged in with `oc` as a privileged user). `http --follow --auth-type=bearer --auth=FAKE$(oc whoami -t) cryostat-sample-cryostat3-helm.apps-crc.testing/api/v3/tls/certs` (a bad/corrupted token) should result in the login page response.


andrewazores:
Screenshot_2024-04-23_21-51-44

Screenshot_2024-04-23_21-51-53

Works now in the most basic case: openshift-oauth-proxy is always deployed, never oauth2_proxy. Basic auth via htpasswd is not yet implemented. SubjectAccessReview and TokenReview (therefore Bearer auth) are not yet implemented. This is deploying the upstream proxy image which has the URL decoding/redirecting bug, so many features don't work and it's tough to test much but I suspect that report generation and "View in Grafana" probably aren't working yet, either.

Tests are failing so must currently be skipped during build.

andrewazores:
```
2024/04/24 17:37:38 provider.go:129: Defaulting client-id to system:serviceaccount:cryostat3:cryostat-sample
2024/04/24 17:37:38 provider.go:134: Defaulting client-secret to service account token /var/run/secrets/kubernetes.io/serviceaccount/token
2024/04/24 17:37:38 provider.go:358: Delegation of authentication and authorization to OpenShift is enabled for bearer tokens and client certificates.
2024/04/24 17:37:38 provider.go:400: unable to retrieve authorization information for users: subjectaccessreviews.authorization.k8s.io is forbidden: User "system:serviceaccount:cryostat3:cryostat-sample" cannot create resource "subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
2024/04/24 17:37:40 provider.go:400: unable to retrieve authorization information for users: subjectaccessreviews.authorization.k8s.io is forbidden: User "system:serviceaccount:cryostat3:cryostat-sample" cannot create resource "subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
2024/04/24 17:37:42 provider.go:400: unable to retrieve authorization information for users: subjectaccessreviews.authorization.k8s.io is forbidden: User "system:serviceaccount:cryostat3:cryostat-sample" cannot create resource "subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
2024/04/24 17:37:44 provider.go:400: unable to retrieve authorization information for users: subjectaccessreviews.authorization.k8s.io is forbidden: User "system:serviceaccount:cryostat3:cryostat-sample" cannot create resource "subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
2024/04/24 17:37:46 provider.go:400: unable to retrieve authorization information for users: subjectaccessreviews.authorization.k8s.io is forbidden: User "system:serviceaccount:cryostat3:cryostat-sample" cannot create resource "subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
2024/04/24 17:37:48 provider.go:400: unable to retrieve authorization information for users: subjectaccessreviews.authorization.k8s.io is forbidden: User "system:serviceaccount:cryostat3:cryostat-sample" cannot create resource "subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
2024/04/24 17:37:48 provider.go:400: unable to retrieve authorization information for users: subjectaccessreviews.authorization.k8s.io is forbidden: User "system:serviceaccount:cryostat3:cryostat-sample" cannot create resource "subjectaccessreviews" in API group "authorization.k8s.io" at the cluster scope
2024/04/24 17:37:48 oauthproxy.go:210: mapping path "/" => upstream "http://localhost:8181/"
```

I guess the (Cluster)Role applied to the Pod, or the authproxy container, needs to gain the permission to create SubjectAccessReviews.

andrewazores:
It looks to me like it should already have such permissions: https://github.com/cryostatio/cryostat-operator/blob/53d2b73f5d141e38682f965ab1cfafe63855edda/config/rbac/role.yaml#L82

@ebaron what am I missing here?

andrewazores:
@ebaron this is still quite rough and incomplete, but I think it's a reasonable starting point for the OpenShift OAuth Proxy. The TLS configuration stuff is a little beyond me right now, but after this I'll try configuring the Basic htpasswd auth on this proxy container as a follow-up PR. Once that's done I can try switching out between the OpenShift OAuth Proxy and OAuth2 Proxy depending on where the Operator is running.

ebaron:
> It looks to me like it should already have such permissions:
> 
> https://github.com/cryostatio/cryostat-operator/blob/53d2b73f5d141e38682f965ab1cfafe63855edda/config/rbac/role.yaml#L82
> 
> @ebaron what am I missing here?

Is it working after adding it to the cryostat_role.yaml as well?

andrewazores:
After [this commit](https://github.com/cryostatio/cryostat-operator/pull/799/commits/7026fe88266a1a4a2a477b3fdd7041d41bce8165) the auth proxy stopped complaining like that in its logs and the behaviour started doing what I expected, ie same as in https://github.com/cryostatio/cryostat-helm/pull/132

ebaron (commented):
I noticed this line in the OAuth Proxy [README](https://github.com/openshift/oauth-proxy/blob/30f8012482023689655252dc2af2f17fe6a09253/README.md?plain=1#L49-L58):
> * Requires no additional permissions to be granted for the proxy service account

It seems like if don't use the delegate URLs option, then we somehow don't need SAR permissions on the service account. This would be great since we could likely remove the entire Cluster Role and Cluster Role Binding for Cryostat's service account. Perhaps we could accomplish the same thing by using the `--bypass-auth-for` option instead like this:

```diff
diff --git a/internal/controllers/common/resource_definitions/resource_definitions.go b/internal/controllers/common/resource_definitions/resource_definitions.go
index e05c691d..55e2f899 100644
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -632,12 +632,8 @@ func NewOpenShiftAuthProxyContainer(cr *model.CryostatInstance, specs *ServiceSp
                        `--openshift-sar=[{"group":"","name":"","namespace":"%s","resource":"pods","subresource":"exec","verb":"create","version":""}]`,
                        cr.InstallNamespace,
                ),
-               fmt.Sprintf(
-                       `--openshift-delegate-urls={"/health":{},"/api":{"group":"","name":"","namespace":"%s","resource":"pods","subresource":"exec","verb":"create","version":""}, "/storage":{"group":"","name":"","namespace":"%s","resource":"pods","subresource":"exec","verb":"create","version":""}, "/grafana":{"group":"","name":"","namespace":"%s","resource":"pods","subresource":"exec","verb":"create","version":""} }`,
-                       cr.InstallNamespace,
-                       cr.InstallNamespace,
-                       cr.InstallNamespace,
-               ),
+
+               `--bypass-auth-for=^/health$`,
                "--proxy-prefix=/oauth2",
        }
        // if tls != nil {
```

I tried it in OpenShift and it seemed to work. Does it work for you too?

andrewazores:
I can test it, but from what I understand from the proxy's docs, that config flag is required for the proxy to handle Bearer authorization headers, which we need for our Agent.

andrewazores:
Tested it and it does look like this flag is required for the `Authorization: Bearer` handling to work. When I remove that flag I am still able to log in interactively in my browser, due to the `--openshift-sar` flag, but using `http` as described in the PR body always fails and just returns the login page HTML response.

ebaron (commented):
Understood on the agent use-case. It wasn't immediately clear to me what the delegate URLs option was for. I'm still not totally clear how they get away without the SAR permission in the normal case, but this seems like our only option.

We could use an entry here for `RELATED_IMAGE_OPENSHIFT_OAUTH_PROXY`: https://github.com/cryostatio/cryostat-operator/blob/main/hack/image_tag_patch.yaml.in
Then we'd need another `make bundle`.

andrewazores:
I think in the normal interactive case the proxy doesn't need the SAR permission because it is only acting as a redirector to send the client to the OpenShift cluster's internal SSO provider. That's what performs the SAR, and if it passes, issues a new token which the client stores as a cookie. Then it redirects the user back to the authproxy, which now receives the cookie, passes it back to the internal SSO provider which approves it, and then the authproxy begins forwarding the user's request to the upstream service (Cryostat container).

andrewazores:
As for why the TokenReview for `Authorization: Bearer` headers does require an elevated TokenReview role... I guess because in this case the proxy itself is handling the raw token passed by the client, rather than simply redirecting the client to the cluster's infrastructure SSO provider. I suppose the proxy must therefore perform the SAR/TR against the internal infrastructure. Even if it could simply pass it on to the cluster SSO, it still handled the client's raw token directly, so it should require elevated permissions. Much like how Cryostat 2.4 directly handles the user's auth tokens.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/cryostat/cryostat-operator
+  newName: quay.io/andrewazores/cryostat-operator
```
ebaron, line 15: Need to unset an environment variable I think

ebaron (approved):

PR #795, opened 2024-04-12
chore(cryostat3): rename/remove env vars, remove unused volume mounts, remove unused CR property

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits): `git commit -S -m "YOUR_COMMIT_MESSAGE"`
_______________________________________________

Based on #727
Depends on #727

## Description of the change:
*This change adds allows the users to provide...*

## Motivation for the change:
*This change is helpful because users may want to...*

## How to manually test:
1. *Insert steps here...*
2. *...*

PR #802, opened 2024-04-26
feat(auth): add configuration for htpasswd basic auth

andrewazores:
# Welcome to Cryostat! 👋

Based on #799

## Description of the change:
Adds CRD fields like:
```yaml
spec:
  authorizationOptions:
    basicAuth:
      secretName: basicauth
      fileName: htpasswd.conf
```

and associated implementation.

## Motivation for the change:
This allows the user to create a secret, like so:

```bash
$ oc create secret generic basicauth --from-file=htpasswd.conf
```

where `htpasswd.conf` has this content, corresponding to `user:pass` credentials:

```htpasswd
user:$2y$05$Ju.w7ibNuQ/xaAfKH74lJOJ5xqZbONaNWf7HnrjQuZP.4nhLrr0uS
```

and then create a Cryostat CR with the `authorizationOptions` as specified above. The secret name and the filename within the secret can be configured to match the actual secret that was created.

With such a configuration, the OpenShift OAuth Proxy will allow interactive users to choose whether to log in via OpenShift SSO or via Basic credentials. Clicking the OpenShift SSO button will redirect the user to the usual cluster login page. Using Basic credentials will simply log the user in. In either case, the user should now have full access to the Cryostat application. Programmatic clients have the choice of sending an `Authorization: Bearer $sometoken` header or an `Authorization: Basic base64("user:pass")` header.

If `authorizationOptions` is not configured then the auth proxy should behave the same as before - interactive users are immediately redirected to the cluster SSO and programmatic clients must use `Bearer` authentication.

When using the OpenShift OAuth Proxy it's not so clear why the user would want to additionally enable htpasswd Basic authentication, but it's possible. It will make more sense later when the implementation allows for deploying OAuth2 Proxy instead of OpenShift OAuth Proxy, for users deploying on non-OpenShift Kubernetes, where there may not be any existing OAuth SSO provider to hook into.


api/v1beta2/cryostat_types.go
```diff
@@ -69,6 +69,10 @@ type CryostatSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	Resources *ResourceConfigList `json:"resources,omitempty"`
+	// Additional configuration options for the authorization proxy.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Authorization Options",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:advanced"}
+	AuthorizationOptions *AuthorizationOptions `json:"authorizationOptions,omitempty"`
```
ebaron, line 75: Can the existing AuthProperties be used for this? Or just remove the AuthProperties and have this instead?

andrewazores: I think it makes sense to remove `AuthProperties` - despite the similarity in naming, these two configuration objects have very different contents and meanings. This new one is (currently) just for providing the secret for htpasswd-based Basic, whereas the old one was for configuring the fine-grained role mapping with the old baked-in OAuth2 client (https://github.com/cryostatio/cryostat/blob/main/src/main/resources/io/cryostat/net/openshift/OpenShiftAuthManager.properties).

I was planning to do all the cleanup and removal of unused configuration items like this in a later PR, similar to #795, but if you'd rather I can clean that piece up in this PR since it is somewhat related.

ebaron: Separate PR is fine with me

ebaron (approved):

PR #803, opened 2024-04-29
feat(auth): add config option to deploy oauth2_proxy instead of openshift-oauth-proxy

### LINKED ISSUE #206: [Story] Use Password Authentication for Kubernetes
Currently, Cryostat only authenticates users by default when run in OpenShift. We should configure password authentication when running on Kubernetes, perhaps generating credentials prior to deployment.

andrewazores:
# Welcome to Cryostat! 👋

Related to #710
Fixes #206

## Description of the change:
This enables the Operator to deploy `oauth2_proxy` in place of `openshift-oauth-proxy`.

`openshift-oauth-proxy` is deployed by default when the Operator detects that it is installed on OpenShift. There is a new CRD option to disable this integration and force the use of `oauth2_proxy` instead. If installed on a non-OpenShift Kubernetes then `oauth2_proxy` will always be deployed.

If `oauth2_proxy` is deployed it will also respect the same htpasswd Basic auth configuration that I added in #802. If no htpasswd Basic auth is configured then `oauth2_proxy` will allow any requests to pass without authz.

## Motivation for the change:
1. enable auth proxy functionality for installs on non-OpenShift clusters
2. enable end users to deploy Cryostat with htpasswd Basic, with or without OpenShift SSO login - they may not want to use cluster RBAC to control access to the Cryostat API and application, but rather to define separate htpasswd user accounts. Or they may wish to provide access to users who have cluster accounts as well as some users who don't
3. enable end users to deploy Cryostat without any authentication required at all. The configuration is locked down so integrating with other OAuth providers is not currently possible (not via the Operator anyway), but for quick one-off testing or small/private testing clusters, configuring auth may be an unwanted additional step

## How to manually test:
1. Build and install Operator as usual
2. Create a CR:

```yaml
kind: Cryostat
metadata:
  name: cryostat-sample
spec:
  enableCertManager: false
  authorizationOptions:
    disableOpenShiftSso: false
  trustedCertSecrets: []
  eventTemplates: []
  storageOptions:
    pvc:
      labels: {}
      annotations: {}
      spec: {}
  reportOptions:
    replicas: 0
```

3. ensure that the application comes up and is secured by OpenShift SSO as before
4. create a basic auth secret like in #802 (`oc create secret generic basicauth --from-file=htpasswd.conf`) patch the CR to add:

```yaml
authorizationOptions:
  basicAuth:
    secretName: basicauth
    filename: htpasswd.conf
```

5. ensure that the application is redeployed and is now secured by OpenShift SSO, with the option of Basic auth as well
6. edit the CR to set `disableOpenShiftSso: true`
7. ensure that the application is redeployed and is now secured by OAuth2 Proxy using the same Basic auth
8. edit the CR to remove the `basicAuth` object
9. ensure that the application is redeployed and is now accessible without auth
10. bonus: install operator on a non-OpenShift cluster. The `disableOpenShiftSso` flag should now be ineffective and OAuth2 Proxy should be installed regardless. The `basicAuth` config option should still function as outlined in the previous steps.


ebaron (commented):
I think we may just want to handle this by using the `IsOpenShift` method instead of adding a CRD property. Downstream we want this code path to be inaccessible since we don't plan to build/release the OAuth2 proxy.

andrewazores:
Hmm, okay. I can do that, but it does mean that once installed on OpenShift the only way to control who has access to the application will be around who has the right roles to pass the SAR (https://github.com/cryostatio/cryostat-operator/pull/799/files#diff-587612cdc407d5ab4abf28789670950c567419bc4ada045dda2e0c1070e74470R632). But I guess that's already how it works in 2.4 if you allow the OpenShift auth integration, so that's still at parity, right?

ebaron:
> Hmm, okay. I can do that, but it does mean that once installed on OpenShift the only way to control who has access to the application will be around who has the right roles to pass the SAR (https://github.com/cryostatio/cryostat-operator/pull/799/files#diff-587612cdc407d5ab4abf28789670950c567419bc4ada045dda2e0c1070e74470R632). But I guess that's already how it works in 2.4 if you allow the OpenShift auth integration, so that's still at parity, right?

Right

ebaron:
> > Hmm, okay. I can do that, but it does mean that once installed on OpenShift the only way to control who has access to the application will be around who has the right roles to pass the SAR (https://github.com/cryostatio/cryostat-operator/pull/799/files#diff-587612cdc407d5ab4abf28789670950c567419bc4ada045dda2e0c1070e74470R632). But I guess that's already how it works in 2.4 if you allow the OpenShift auth integration, so that's still at parity, right?
> 
> Right

Isn't there also the Basic Auth option now too?

andrewazores:
Yes, but enabling that in OpenShift gives you the option of logging in with OpenShift SSO or with Basic auth - it doesn't disable the OpenShift SSO part.

I don't think there is a way to configure the OpenShift OAuth Proxy to turn off the OpenShift SSO integration, but maybe there's a way to configure it to be just slightly broken on purpose... it would still present the login button to the user, but maybe it can be made to always fail at that stage. Not sure if that would be worthwhile.

ebaron:
I suppose for full 2.4 parity we'd need the ability to customize the permissions used for the SAR.

andrewazores:
Would it make sense to add that directly as structures within the CR's new `AuthorizationOptions`? That's a bit simpler and IMO nicer compared to 2.4's "configmap containing a properties file", I think.

andrewazores:
Ah, I also just realized I forgot to fix this up: https://github.com/cryostatio/cryostat-operator/blob/d455719eee5eac03872eb42d5cfe149335064c05/internal/controllers/common/resource_definitions/resource_definitions.go#L638

I'll do that now.

andrewazores:
I think the CR API looks a little clumsy, but all these cases are covered:

1. deployed on OpenShift, OpenShift SSO is enabled by default and Basic is not configured
2. deployed on OpenShift, Basic can be additionally configured using `spec.authorizationOptions.basicAuth.[secretName|filename]`
3. deployed on OpenShift, all auth can be disabled (SSO, and Basic if configured) using `spec.authorizationOptions.openShiftSsoRoles.disabled: true`
4. deployed on OpenShift, the SSO SubjectAccessReview and TokenReview can be configured using `spec.authorizationOptions.openShiftSsoRoles.[subjectAccessReview|tokenReview]`
5. deployed on non-OpenShift, all auth is disabled by default
6. deployed on non-OpenShift, Basic can be configured the same way as in point 2

What is not possible is deploying on OpenShift, disabling SSO, and enabling Basic only.

It's also a little clumsy that on OpenShift this defaults to enabling authz, but on non-OpenShift it defaults to deploying wide open. As suggested in #206 we could generate a secret and enable Basic auth by default and attach that to the CR status like the Grafana credentials in 2.4.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -695,10 +739,143 @@ func NewOpenShiftAuthProxyContainer(cr *model.CryostatInstance, specs *ServiceSp
...
+}
+
+func getOpenShiftTokenReview(cr *model.CryostatInstance) authzv1.ResourceAttributes {
+	if cr.Spec.AuthorizationOptions != nil && cr.Spec.AuthorizationOptions.OpenShiftSsoRoles != nil && cr.Spec.AuthorizationOptions.OpenShiftSsoRoles.TokenReview != nil {
+		return *cr.Spec.AuthorizationOptions.OpenShiftSsoRoles.TokenReview
+	}
+	return getDefaultOpenShiftAccessRole(cr)
+}
+
+func getDefaultOpenShiftAccessRole(cr *model.CryostatInstance) authzv1.ResourceAttributes {
```
andrewazores, line 765: I guess I can replace this function and some of the logic around it by using `internal/webhooks/defaulter.go`, right?

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -635,27 +660,34 @@ func NewOpenShiftAuthProxyContainer(cr *model.CryostatInstance, specs *ServiceSp
...
-		"--bypass-auth-for=^/health",
 		"--proxy-prefix=/oauth2",
 	}
+	if isOpenShiftAuthProxyDisabled(cr) {
+		args = append(args, "--bypass-auth-for=.*")
+	} else {
+		args = append(args, "--bypass-auth-for=^/health")
+	}
+
+	// FIXME don't ignore this error! We're far removed from the reconciler loop - we should return this
```
andrewazores, line 674: Any advice for what to do here? I didn't want to add an `error` return to this function and its caller etc. all the way back up to the reconcile loop function. Maybe it's safe to ignore because at this point the object is either a default structure we've created or else it's a reference to a CR field, which means it has already been validated and deserialized by Kubernetes to begin with - how can we have an error in serializing it?

ebaron: I would just propagate the error to the reconcile loop. A more user-friendly way could be to add the validation into our validating webhook, but that may be more RFE territory.

internal/controllers/common/resource_definitions/resource_definitions.go
ebaron: You could, but I think for the most part we still don't set defaults in the CR, and just do like you've done already here. It could cause confusion for the user if we set these fields when running on Kubernetes for example.

api/v1beta2/cryostat_types.go
```diff
@@ -487,12 +488,38 @@ type TemplateConfigMap struct {
 
 // Authorization options provide additional configurations for the auth proxy.
 type AuthorizationOptions struct {
-	// Reference to a secret and file name containing the Basic authentication htpasswd file
+	// Configuration for OpenShift RBAC to define which OpenShift user accounts may access the Cryostat application.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	OpenShiftSsoRoles *OpenShiftSsoRoles `json:"openShiftSsoRoles,omitempty"`
```
ebaron, line 494: This should be `OpenShiftSSORoles`: https://go.dev/wiki/CodeReviewComments#initialisms. We'll also need a displayName here for "OpenShift SSO Roles"

api/v1beta2/cryostat_types.go
```diff
@@ -487,12 +488,38 @@ type TemplateConfigMap struct {
...
+	// application's installation namespace.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	SubjectAccessReview *authzv1.ResourceAttributes `json:"subjectAccessReview,omitempty"`
+	// The TokenReview that programmatic clients (CLI utilities and others presenting Bearer auth tokens) must pass in
+	// order to access the application. If not specified, the default role required is "create pods/exec" in the Cryostat
+	// application's installation namespace.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	TokenReview *authzv1.ResourceAttributes `json:"tokenReview,omitempty"`
```
ebaron, line 520: Is TokenReview right here? That API is used to determine who a token belongs to isn't it? Also, do we need to have separate options for interactive vs non-interactive?

api/v1beta2/cryostat_types.go
```diff
@@ -487,12 +488,38 @@ type TemplateConfigMap struct {
...
+	// will also bypass the BasicAuth, if specified.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Disable OpenShift SSO",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	Disabled *bool `json:"disabled,omitempty"`
+	// The SubjectAccessReview that interactive clients (users visiting the application via web browser) must pass in
+	// order to access the application. If not specified, the default role required is "create pods/exec" in the Cryostat
+	// application's installation namespace.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	SubjectAccessReview *authzv1.ResourceAttributes `json:"subjectAccessReview,omitempty"`
```
ebaron, line 514: Should this accept multiple ResourceAttributes? Then the user would need to have all permissions listed. I think this would be equivalent to the current role-based system in 2.4.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -635,27 +668,36 @@ func NewOpenShiftAuthProxyContainer(cr *model.CryostatInstance, specs *ServiceSp
...
-			`--openshift-delegate-urls={"/":{"group":"","name":"","namespace":"%s","resource":"pods","subresource":"exec","verb":"create","version":""}}`,
-			cr.InstallNamespace,
-		),
-		"--bypass-auth-for=^/health",
 		"--proxy-prefix=/oauth2",
 	}
+	if isOpenShiftAuthProxyDisabled(cr) {
+		args = append(args, "--bypass-auth-for=.*")
+	} else {
+		args = append(args, "--bypass-auth-for=^/health")
```
ebaron, line 679: Do we need something like `$` at the end of the regex here? We probably won't get into a situation where there's a valid path like `/healthfoo` that's not supposed to bypassed, but it might be good just in case.

api/v1beta2/cryostat_types.go
andrewazores: I guess it is actually `TokenAccessReview`, not just `TokenReview`. I'll rename it.

> Also, do we need to have separate options for interactive vs non-interactive?

I think conceptually it makes sense to require the same access reviews either way, but unfortunately the way the proxy works is that you configure the interactive access as a list of objects that map a route prefix to a SAR, whereas the Bearer auth access is only a single object without any route prefix.

https://github.com/cryostatio/cryostat-helm/pull/132#discussion_r1571462608

So I think the most understandable model from the end user's point of view is probably to have a single `AccessReview` field, which is used for both TokenAccessReview and as a singleton list for SubjectAccessReview with the `/` prefix. The most flexible model is to have a prefixed list of SARs and a separate value for TAR.

andrewazores: FWIW over in the Helm chart, I have it as a single Bearer TokenAccessReview and a list of SubjectAccessReviews.

ebaron: Ah so the delegate URLs option doesn't support a set of permissions per path? I agree with the simplest option then. One SubjectAccessReview field with one ResourceAttributes struct, and this will get used for both `openshift-sar` and `openshift-delegate-urls`.

andrewazores: That's unfortunately correct, the delegate URLs option which enables Bearer token handling only takes a single permission per path prefix. But it also doesn't allow for inspecting the request at all beyond checking that the request path matches the path prefix, so we can't even differentiate between ex. `GET` and `POST` to apply different role mapping.

api/v1beta2/cryostat_types.go
```diff
@@ -487,12 +488,32 @@ type TemplateConfigMap struct {
...
 	BasicAuth *SecretFile `json:"basicAuth,omitempty"`
 }
 
+type OpenShiftSSOConfig struct {
+	// Disable OpenShift SSO integration and allow all users to access the application without authentication. This
+	// will also bypass the BasicAuth, if specified.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Disable OpenShift SSO",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	Disable *bool `json:"disable,omitempty"`
+	// The AccessReview or TokenAccessReview that all clients (users visiting the application via web browser as well
```
ebaron, line 509:
```suggestion
	// The SubjectAccessReview that all clients (users visiting the application via web browser as well
```

api/v1beta2/cryostat_types.go
```diff
@@ -487,12 +488,32 @@ type TemplateConfigMap struct {
...
 // Authorization options provide additional configurations for the auth proxy.
 type AuthorizationOptions struct {
-	// Reference to a secret and file name containing the Basic authentication htpasswd file
+	// Configuration for OpenShift RBAC to define which OpenShift user accounts may access the Cryostat application.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="OpenShift SSO"
+	OpenShiftSSO *OpenShiftSSOConfig `json:"openShiftSSO,omitempty"`
+	// Reference to a secret and file name containing the Basic authentication htpasswd file. If deploying on OpenShift this
+	// defines additional user accounts that can access the Cryostat application, on top of the OpenShift user accounts which
+	// pass the OpenShiftSsoRoles checks. If not on OpenShift then this defines the only user accounts that have access.
```
ebaron, line 497:
```suggestion
	// pass the OpenShift SSO Roles checks. If not on OpenShift then this defines the only user accounts that have access.
```

ebaron (approved):
Looks good! Just a couple more doc suggestions you can take or leave

PR #808, opened 2024-05-03
chore(cryostat3): various cleanups

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits): `git commit -S -m "YOUR_COMMIT_MESSAGE"`
_______________________________________________

Related to #710

## Description of the change:
- cleans up some unused volume mounts and env vars
- removes Grafana basic auth secret - no longer needed, since Grafana is placed behind the same auth proxy as Cryostat
- configures Grafana for anonymous access (not really anonymous since traffic must go through auth proxy first)
- corrects environment variables so Cryostat UI "View in Grafana" correctly opens Grafana UI
- renames some CR fields to better reflect what they mean in 3.0
- updates docs/config.md
- removes AuthProperties CR field since this is replaced by AuthorizationOptions now

## Motivation for the change:
General clean up, get Grafana working.

## How to manually test:
1. Check out, build, and test PR. Use `OPENSHIFT_OAUTH_PROXY_IMG=quay.io/andrewazores/openshift-oauth-proxy:test-14` (rebuild with golang 1.22) so the UI is usable without the proxy path redirect bug.
2. Create Cryostat CR with cert_manager integration disabled
3. Note Application Route URL, open a new private/incognito browser tab and visit ex. `http://cryostat-cryostat.apps-crc.testing/grafana` to go directly to the Grafana dashboard. This should prompt you to log in with OpenShift, just like visiting the Cryostat UI.
4. In Cryostat UI, log in with OpenShift, then define a `localhost:0` custom target, start a recording, and "View in Grafana" it. This should pass you directly through to Grafana without re-auth.
5. Copy and paste the "View in Grafana" tab URL into a new incognito window to simulate a different, unauthenticated user. Paste and go. You should be prompted to log in.
6. Archive the recording and repeat the same "View in Grafana" tests.

PR #809, opened 2024-05-03
chore(grafana): remove separate Grafana service/route/ingress

andrewazores:
# Welcome to Cryostat! 👋

Depends on #808
Based on #808
Related to #710

## Description of the change:
Grafana should now only be accessible through the auth proxy at the `/grafana` path. It should not have its own separate service, or Route/Ingress, because all traffic must flow through the auth proxy now.

## Motivation for the change:
This ensures that Grafana is only accessible using the same authn/authz as the rest of the application.

## How to manually test:
1. Check out PR, build with `OPENSHIFT_OAUTH_PROXY_IMG=quay.io/andrewazores/openshift-oauth-proxy:test-14`. Deploy Operator and create a Cryostat CR with `enableCertManager: false`
2. Open Cryostat web UI and log in via OpenShift SSO as cluster admin
3. Create a `localhost:0` custom target
4. Start a recording, give it a few moments, then "View in Grafana" it. This should open a new browser tab. The URL should have the same `https` scheme and the same  domain, but should differ in the path - the Grafana dashboard should be at `https://cryostat-example.apps.crc-testing/grafana`. You should not need to re-login to view this.
5. Copy the Grafana dashboard URL, then paste into a new private/incongnito window and go. You should be prompted to log in via OpenShift SSO. Log in as cluster admin and you should see the same data as in the previous step. Remove `/grafana` from the URL and you should go back to the Cryostat UI, without needing to re-login.
6. Archive the recording, then repeat steps 4 and 5 on that.
7. `oc get routes` and `oc get services` and verify that there is only one Service/Route associated with the Cryostat instance, which points to port `4180` - the authproxy HTTP port.


ebaron (approved):
Looks good!

PR #815, opened 2024-05-09
feat(cryostat3): deploy Cryostat 3.0

andrewazores:
# Welcome to Cryostat! 👋

Merge feature branch `cryostat3` into main, rolling up all 3.0-related PRs.


andrewazores:
One "bug" I realized affects 3.0 here is that since Cryostat itself is behind the proxy, the `9091` port that it exposes on the Pod is actually now on the Service that points at the auth proxy. The auth proxy doesn't handle non-HTTP/non-WebSocket connections, so it doesn't proxy JMX connections, and therefore Cryostat's KubeApi discovery that looks for Endpoints objects will find what looks like a JMX port on 9091 (its own JMX port) but be unable to connect to it because it will actually get routed to the proxy.

We should either remove the JMX port from the Deployment/Pod/Service so that Cryostat doesn't see it, or we need to add a separate Service to advertise this port so that the traffic can bypass the proxy.

tthvo:
> One "bug" I realized affects 3.0 here is that since Cryostat itself is behind the proxy, the 9091 port that it exposes on the Pod is actually now on the Service that points at the auth proxy. The auth proxy doesn't handle non-HTTP/non-WebSocket connections

I have a question here: wouldn't the service already route the connection to `9091` to pod's `9091` which is open on the cryostat container? 

My understanding is that the auth proxy is listening on port `4180` only which it does the auth check. Otherwise, if other containers expose other ports, it essentially won't go through the proxy? I might be misunderstanding tho...

tthvo:
The [doc here](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports) says ports on `0.0.0.0` will be accessible via the network. The issue might just be the jmx host is set to be `localhost`?

andrewazores:
The auth proxy is listening on 4180 for any traffic. Depending on the HTTP path of the incoming request, the auth proxy forwards that request to another container (on another port) - if the path begins with `/grafana` then it sends the request to the Grafana dashboard container listening on port 3000 for example.

If something tries to open a JMX connection to port 9091, even if that port is exposed, the proxy won't know how it's supposed to handle that - it isn't an HTTP connection, so there is no HTTP path, and the proxy doesn't know which container to pass the request to.

tthvo:
Ah yess thank you! That's very insightful!

> Any unauthorized user could just go make requests to 8181 directly and access everything unauthenticated.

I guess my thoughts are that for the k8s service the operator creates, it maps requests to 9091 directly to the pod's 9091. This is cryostat container's rite?

So when cryostat opens a jmx connection to itself, it wont go through the proxy?

andrewazores:
True, the Endpoints object should have port 9091 corresponding to the Cryostat container within the Pod, not the auth proxy container. I need to double check how that is put together currently. In theory this should allow direct communication to the 9091 port on the target (Cryostat) container, but I think this still needs the Service to expose that port and I'm not sure if it currently does.






---- On Tue, 14 May 2024 17:03:27 -0400 ***@***.*** wrote ----



Ah yess thank you! That's very insightful!

Any unauthorized user could just go make requests to 8181 directly and access everything unauthenticated.

I guess my thoughts are that for the k8s service the operator creates, it maps requests to 9091 directly to the pod's 9091. This is cryostat container's rite?

So when cryostat opens a jmx connection to itself, it wont go through the proxy?

tthvo:
> True, the Endpoints object should have port 9091 corresponding to the Cryostat container within the Pod, not the auth proxy container. I need to double check how that is put together currently. In theory this should allow direct communication to the 9091 port on the target (Cryostat) container, but I think this still needs the Service to expose that port and I'm not sure if it currently does.

I think it is currently doing that?

https://github.com/cryostatio/cryostat-operator/blob/9fa1007c976ad7de3ad9c22b0260ee0f0f95481c/internal/controllers/services.go#L50

tthvo:
I guess my questions come up as connection to cryostat on helm side (without jmx auth?) seems to be working fine.

andrewazores:
Hmm. That's a good point. I'm not sure what the difference is then other than auth, which I think is working fine in the podman smoketest...

tthvo:
> I guess my questions come up as connection to cryostat on helm side (without jmx auth?) seems to be working fine.

Ahh nope. Also broken on helm side. I was a bit confused as the recording table shows no error.

tthvo:
> Hmm. That's a good point. I'm not sure what the difference is then other than auth, which I think is working fine in the podman smoketest...

That being said, I think helm chart also has jmx auth enabled for cryostat. I think the password are generated by default in the entrypoint. It's not reported in a k8s secret like on the operator side. Is this intentional?

tthvo:
I wonder if we need to set the below flag to the jvm in the entrypoint? The default JMX port is not `9091` rite?

```
-Dcom.sun.management.jmxremote.port=9091 -Dcom.sun.management.jmxremote.rmi.port=9091
```

andrewazores:
> We use to have it in 2.x entrypoint but seems like 3.0 does not.

Right - it can be added to the `JAVA_OPTS_APPEND` now, which is part of the base image.

https://github.com/cryostatio/cryostat3/blob/94134bd826ff62397f67d1cbac3b3d797876064d/compose/cryostat.yml#L39

I think that's missing from this new setup... that would explain things.

tthvo:
> > We use to have it in 2.x entrypoint but seems like 3.0 does not.
> 
> Right - it can be added to the `JAVA_OPTS_APPEND` now, which is part of the base image.
> 
> https://github.com/cryostatio/cryostat3/blob/94134bd826ff62397f67d1cbac3b3d797876064d/compose/cryostat.yml#L39
> 
> I think that's missing from this new setup... that would explain things.

Yea, i suspect the connection to 9091 is refused since it's not actually open without these flags :d

ebaron:
Since Cryostat no longer handles TLS itself, does that mean this JMX server always allows clear-text connections?

andrewazores:
Right now it doesn't handle TLS and it also doesn't expose JMX. I'm working on enabling the JMX server again but running into some odd behaviour with it and some apparent port number conflicts, somehow. I can make it handle TLS again for the JMX server only and continue relying on the auth proxy to do HTTPS.

Otherwise, it might be OK to just drop the JMX server from Cryostat itself. If the user needs to be able to monitor Cryostat itself they can always define the `localhost:0` custom target.

ebaron:
> FWIW I'm leaning toward removing the JMX server from Cryostat, since `localhost:0` can be used instead and is just about the most secure way that the user can monitor/profile Cryostat itself, while also requiring nothing from the Operator to configure it.

That makes sense, or just restrict the JMX server to listen on localhost only, but that may be more hassle than it's worth

andrewazores:
@mwangggg https://github.com/cryostatio/cryostat-operator/pull/815/commits/c1b194a119d80e50fc8ff21f7bb89ed30ef8f755 looks like an accident - can I remove that?

mwangggg:
I removed it already!

tthvo:
Looks like `controller-gen` is panicking in go `1.22` :((

tthvo:
> Looks like `controller-gen` is panicking in go `1.22` :((

Bumping controller-tools to 0.14.0+ seems to solve the issue: https://github.com/kubernetes-sigs/controller-tools/issues/880.

ebaron:
I think we can just upgrade controller-tools. I did a test run with `make bundle` and it seems like it just made some YAML formatting nicer.

ebaron (approved):

ebaron:
I'd like to do a merge if possible to preserve individual commits

andrewazores:
Sure, makes sense

Issue #475, opened 2022-10-14
[Bug] Credentials DB password should not update


tthvo:
> How about this idea to guard against the user breaking their credentials DB by destroying the password?
> 1. If the CR doesn't have a defined `DatabaseSecretName`, create the default secret with a generated password
> 2. When creating the deployment, check whether the default secret exists.
    - If it does, then the CR initially had no custom secret and the credentials DB would have been created with the generated password.
>     - If it doesn't, then check if `DatabaseSecretName` has been set (it should), and use that.
>     - If for some reason the `DatabaseSecretName` is not set, then the password has been lost. Likely due to the user deleting the default secret. We could emit an Event here.

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/pull/474#pullrequestreview-1142820938_

ebaron:
From https://github.com/cryostatio/cryostat-operator/pull/474#issuecomment-1279347320:
>     Just a question: What about the case where the user switch from custom secret to default one? The default secret will take place and still make Cryostat fail right?
> 
> True, that would be harder to detect. How about this idea instead? We add a `Status.DatabaseSecret` to the CRD. We write to this field once, which will be the canonical database secret for the lifetime of that Cryostat. We could do this by comparing against the existing value of the field and only setting it when nil. Then the secret referenced here would be what the Deployment uses.

PR #831, opened 2024-05-23
fix(crd): restrict database secret name to be immutable

tthvo:
# Welcome to Cryostat! 👋

Fixes: #475 

## Description of the change:

Updated the reconciler logic to emit an Event to warn against mismatched secret name for database and return an approriate error.

## Motivation for the change:

See #475. The secret used to database encryption must not change (i.e. `databaseOptions.secretName` is immutable).

## How to manually test:

Create a secret:

```bash
cat <<EOF | oc apply -f -
apiVersion: v1
stringData:
  CONNECTION_KEY: connection_key
  ENCRYPTION_KEY: encryption_key
kind: Secret
metadata:
  name: db-secret
type: Opaque
EOF
```

Deploy the operator and create a CR:

```bash
export IMAGE_NAMESPACE=quay.io/some-user
export IMAGE_VERSION=test-version
make oci-build bundle bundle-build
podman push $IMAGE_NAMESPACE/cryostat-operator:$IMAGE_VERSION
podman push $IMAGE_NAMESPACE/cryostat-operator-bundle:$IMAGE_VERSION
make deploy_bundle

cat <<EOF | oc apply -f -
kind: Cryostat
metadata:
  name: cryostat-sample
spec:
  enableCertManager: true
  databaseOptions:
    secretName: db-secret
EOF
```
Wait for the controller finishes reconciling. Then, update the CR to remove the databaseOptions.secretName spec. The controller should show an error in reconciler loop and an event emitted. Vice versa, create an empty CR, wait for reconciling, and update the CR to specify the databaseOptions.secretName spec. The same behaviour should occur.

### Controller's log:

```
2024-05-31T05:48:59Z	ERROR	Reconciler error	{"controller": "cryostat", "controllerGroup": "operator.cryostat.io", "controllerKind": "Cryostat", "Cryostat": {"name":"cryostat-sample","namespace":"myns"}, "namespace": "myns", "name": "cryostat-sample", "reconcileID": "f2dcb01e-d6a4-4d92-b949-11727633e32b", "error": "database secret cannot be updated, but another secret is specified"}
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.5/pkg/internal/controller/controller.go:329
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.5/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.16.5/pkg/internal/controller/controller.go:227
2024-05-31T05:48:59Z	DEBUG	events	"databaseOptions.secretName" field cannot be updated, please revert its value or re-create this Cryostat custom resource	{"type": "Warning", "object": {"kind":"Cryostat","namespace":"myns","name":"cryostat-sample","uid":"21527815-9aa8-4196-8bce-0b0625e28596","apiVersion":"operator.cryostat.io/v1beta2","resourceVersion":"165990"}, "reason": "DatabaseSecretMismatched"}
```

### Events

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,9 @@ the JFR datasource for Grafana.
 for the Grafana dashboard.
 
 # Using
+
+Minimum Supported Kubernetes Version: `v1.21`.
+
 Once deployed, the `cryostat` instance can be accessed via web browser
 at the URL provided by:
 ```
--- a/README.md
+++ b/README.md
@@ -27,7 +27,12 @@ for the Grafana dashboard.
 
 # Using
 
-Minimum Supported Kubernetes Version: `v1.21`.
+## Requirements
+
+- `kubernetes` v1.21+ with [`Operator Lifecycle Manager`](https://olm.operatorframework.io/)
+- [`cert-manager`](https://github.com/cert-manager/cert-manager) v1.11.5+ (Recommended)
+
+## Instructions
 
 Once deployed, the `cryostat` instance can be accessed via web browser
 at the URL provided by:
@@ -44,7 +49,6 @@ the need to expose a JMX port over the network.
 ## Requirements
 - `go` v1.21
 - [`operator-sdk`](https://github.com/operator-framework/operator-sdk) v1.31.0
-- [`cert-manager`](https://github.com/cert-manager/cert-manager) v1.11.5+ (Recommended)
 - `podman` or `docker`
 - [`jq`](https://stedolan.github.io/jq/) v1.6+
 - `ginkgo` (Optional)
--- a/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
@@ -53,7 +53,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: Monitoring, Developer Tools
     containerImage: quay.io/cryostat/cryostat-operator:3.0.0-dev
-    createdAt: "2024-05-31T05:26:09Z"
+    createdAt: "2024-05-31T05:27:37Z"
     description: JVM monitoring and profiling tool
     operatorframework.io/initialization-resource: |-
       {
@@ -581,7 +581,8 @@ spec:
           and an ENCRYPTION_KEY secret which is the key used to encrypt sensitive
           data stored within the database, such as the target credentials keyring.
           This field cannot be updated. It is recommended that the secret should be
-          marked as immutable to avoid accidential mutations. More details: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable'
+          marked as immutable to avoid accidental changes to secret''s data. More
+          details: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable'
         displayName: Secret Name
         path: databaseOptions.secretName
         x-descriptors:
--- a/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
@@ -53,7 +53,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: Monitoring, Developer Tools
     containerImage: quay.io/cryostat/cryostat-operator:3.0.0-dev
-    createdAt: "2024-05-28T20:14:14Z"
+    createdAt: "2024-05-31T05:18:57Z"
     description: JVM monitoring and profiling tool
     operatorframework.io/initialization-resource: |-
       {
@@ -580,6 +580,7 @@ spec:
           contain a CONNECTION_KEY secret which is the database connection password,
           and an ENCRYPTION_KEY secret which is the key used to encrypt sensitive
           data stored within the database, such as the target credentials keyring.
+          Cannot be updated.
         displayName: Secret Name
         path: databaseOptions.secretName
         x-descriptors:
--- a/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
@@ -53,7 +53,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: Monitoring, Developer Tools
     containerImage: quay.io/cryostat/cryostat-operator:3.0.0-dev
-    createdAt: "2024-05-31T05:18:57Z"
+    createdAt: "2024-05-31T05:26:09Z"
     description: JVM monitoring and profiling tool
     operatorframework.io/initialization-resource: |-
       {
@@ -576,11 +576,12 @@ spec:
       - description: Options to configure the Cryostat application's database.
         displayName: Database Options
         path: databaseOptions
-      - description: Name of the secret containing database keys. This secret must
+      - description: 'Name of the secret containing database keys. This secret must
           contain a CONNECTION_KEY secret which is the database connection password,
           and an ENCRYPTION_KEY secret which is the key used to encrypt sensitive
           data stored within the database, such as the target credentials keyring.
-          Cannot be updated.
+          This field cannot be updated. It is recommended that the secret should be
+          marked as immutable to avoid accidential mutations. More details: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable'
         displayName: Secret Name
         path: databaseOptions.secretName
         x-descriptors:
--- a/internal/controllers/secrets.go
+++ b/internal/controllers/secrets.go
@@ -73,7 +73,6 @@ func (r *Reconciler) reconcileDatabaseConnectionSecret(ctx context.Context, cr *
 			Name:      cr.Name + databaseSecretNameSuffix,
 			Namespace: cr.InstallNamespace,
 		},
-		Immutable: &[]bool{true}[0],
 	}
 	secretName := secret.Name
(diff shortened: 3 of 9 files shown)


tthvo:
There might be cases where the user simply migrate the content of the custom database secret to another one. This means the keys are the same but the operator will consider it as updated. 

I am not sure if it's good practice or secure to check secret contents or I am overthinking. Let me know what you think :D

andrewazores:
> I am not sure if it's good practice or secure to check secret contents or I am overthinking. Let me know what you think :D

Seems to me like it should be OK to do in general, so long as we are careful with how we handle the secret after we have accessed it. ie don't log the secret contents, don't copy it to another file/secret/configmap/etc., avoid holding it in memory on heap and if this can't be avoided then be sure to write over the char/byte array so that the secret is unlikely to be seen within a heap dump.

tthvo:
> > I am not sure if it's good practice or secure to check secret contents or I am overthinking. Let me know what you think :D
> 
> Seems to me like it should be OK to do in general, so long as we are careful with how we handle the secret after we have accessed it. ie don't log the secret contents, don't copy it to another file/secret/configmap/etc., avoid holding it in memory on heap and if this can't be avoided then be sure to write over the char/byte array so that the secret is unlikely to be seen within a heap dump.

Thanks! Make sense to me! Are we looking to implement the checks at secret field level here? If so, I make this draft and move in that direction. Or is this good?

andrewazores:
I don't think it's necessary to do that, but if we were to implement something like that, those would be the constraints. For example,

```
reconcile() {
  ....
  if len(cr.Status.SecretHash) == 0 {
    databaseSecret := generatePassword()
    defer clearString(databaseSecret)
    cr.Status.SecretHash = sha256(databaseSecret)
    createSecret("database", databaseSecret)
  } else {
    currentSecret = getSecret("database")
    hash = sha256(currentSecret)
    defer clearString(currentSecret)
    if hash != cr.Status.SecretHash {
      return badSecretError
    }
  }
}
```

as a very rough pseudocode example. Here the idea is that the generated secret is hashed and the hash stored in the CR status, so we can use this hash for change detection rather than relying on Secret naming for change detection.

There are probably better k8s-specific ways to do this, but if you were to directly examine the Secret contents to do change detection, I think it would look something like this.

tthvo:
Ahh how about the cases where the secret contains more than just `CONNECTION_KEY` and `ENCRYPTION_KEY`? That should make the hash different right even though the secret values are valid. We would need to save hash for each field. Seem not very ideal...

I guess I am also a bit hesitant to save to secret hash in CR status.

andrewazores:
Making it immutable sounds like another good option. Basically, I think the right options are the ones that are intuitive for the user. So if it's immutable, they will understand that they need to delete and recreate things over again if they want to make any changes. If it is mutable then I think they would expect full change detection and automatic redeploy as needed - having it mutable but only checking for name changes seems like it could be confusing, since a rename with no content change should not actually do anything, whereas keeping the same name with updated content should cause a redeploy.

tthvo:
Ahh thanks make sense! I think, IMO, the more simple and easier way with more restrictions (i.e. immutable field) would fit more since it's common in k8s too.

I can try: https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/ (not working for some reasons but I can check again).

We can also suggest the users to make their secret immutable: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable (k8s 1.21+). That being said, we are still supporting k8s 1.19+?

ebaron:
> We can also suggest the users to make their secret immutable: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable (k8s 1.21+). That being said, we are still supporting k8s 1.19+?

I'd be fine with us bumping the minimum support to 1.25.

tthvo:
> > We can also suggest the users to make their secret immutable: https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable (k8s 1.21+). That being said, we are still supporting k8s 1.19+?
> 
> I'd be fine with us bumping the minimum support to 1.25.

Ah nice, considering pretty much k8s 1.25- are EOL. Another thing, would it make sense to to emit a warning if the secret provided is immutable? For our own generated one, we can add the `immutable: true`.

tthvo:
> A transition rule is a validation rule that references 'oldSelf'. The API server only evaluates transition rules when both an old value and new value exist.

https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/#crd-transition-rules

> Both an old and a new value exist. It remains possible to check if a value has been added or removed by placing a transition rule on the parent node. Transition rules are never applied to custom resource creation. When placed on an optional field, a transition rule will not apply to update operations that set or unset the field.

 https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#transition-rules

Unfortunately, I ran into a bit of a roadblock here. The CRD's validation/transition rules have above restrictions. Indeed, the updates from unset to set `databaseOptions.secretName` (and vice versa) are totally ignored while changing custom secret names are invalidated correctly :(

There are possible workaround that requires k8s 1.30+ with explicit enabling of [CRD validation ratcheting](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-ratcheting) feature gate. IMO, I don't think this is ideal to enforce it.

So, I kept the original implementation that would detect the change in `databaseOptions.secretName` and emit a warning Event. This behaves as if the field is immutable and allows reverting CR specs. I also added `immutable` to the generated database secret to avoid accidental mutations. What are your thoughts @ebaron @andrewazores?

README.md
```diff
@@ -26,6 +26,9 @@ the JFR datasource for Grafana.
 for the Grafana dashboard.
 
 # Using
+
+Minimum Supported Kubernetes Version: `v1.21`.
+
```
andrewazores, line 31: Maybe this should have a requirements list like what is in `Building`.

```markdown
# Using

## Requirements

- `kubernetes` v1.21+
- `operator-lifecycle-manager`
- `cert-manager` (optional)

# Building

## Requirements

- `go` v1.21+
- `operator-sdk` v1.31.0
- `podman` (or `buildah`?) or `docker`
- `jq` v1.6+
- `ginkgo` (optional)
- `cert-manager` (optional)

File: bundle/manifests/cryostat-operator.clusterserviceversion.yaml [hunk 2]
```diff
@@ -576,10 +576,13 @@ spec:
       - description: Options to configure the Cryostat application's database.
         displayName: Database Options
         path: databaseOptions
-      - description: Name of the secret containing database keys. This secret must
+      - description: 'Name of the secret containing database keys. This secret must
           contain a CONNECTION_KEY secret which is the database connection password,
           and an ENCRYPTION_KEY secret which is the key used to encrypt sensitive
           data stored within the database, such as the target credentials keyring.
+          This field cannot be updated. It is recommended that the secret should be
+          marked as immutable to avoid accidental changes to secret''s data. More
```
andrewazores, line 584: Something weird happening here with `''`

andrewazores:
I think the change itself makes sense, just the two documentation notes above.

README.md
tthvo: Good idea, updated now :D

bundle/manifests/cryostat-operator.clusterserviceversion.yaml
tthvo: I guess that single quote `'` is used to escape the `'s` in YAML? If I do, `make install && kubectl explain cryostat.spec.databaseOptions`, it looks alright.

```
GROUP:      operator.cryostat.io
KIND:       Cryostat
VERSION:    v1beta2

FIELD: databaseOptions <Object>

DESCRIPTION:
    Options to configure the Cryostat application's database.
    
FIELDS:
  secretName	<string>
    Name of the secret containing database keys. This secret must contain a
    CONNECTION_KEY secret which is the database connection password, and an
    ENCRYPTION_KEY secret which is the key used to encrypt sensitive data stored
    within the database, such as the target credentials keyring. This field
    cannot be updated. It is recommended that the secret should be marked as
    immutable to avoid accidental changes to secret's data. More details:
    https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable
```

andrewazores (approved):
LGTM, @ebaron?

internal/controllers/secrets.go
```diff
@@ -57,23 +58,36 @@ func (r *Reconciler) reconcileAuthProxyCookieSecret(ctx context.Context, cr *mod
...
+var errDatabaseSecretUpdated = errors.New("database secret cannot be updated, but another secret is specified")
 
 func (r *Reconciler) reconcileDatabaseConnectionSecret(ctx context.Context, cr *model.CryostatInstance) error {
-	var secretName string
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cr.Name + databaseSecretNameSuffix,
+			Namespace: cr.InstallNamespace,
+		},
+		Immutable: &[]bool{true}[0],
```
ebaron, line 76: Should we set Immutable inside the createOrUpdate logic below? This would allow us to make the secret immutable as part of an upgrade.

tthvo: Good catch! Updated now. Once again, I forgot about this haha :D Thanks!!

ebaron (approved):
Looks good!

Issue #901, opened 2024-06-27
[Request] Helm chart for the operator


grzesuav:
### Describe the feature

Hi, I would like to install operator using helm chart - can you provide one ? Or is there any already existing

### Anything other information?

N/A

andrewazores:
There is no such thing existing, we have either the Operator or the Helm Chart available. Can you describe what your requirements are that you want the Operator functionality but installable via Helm Chart? Maybe there is another way to accomplish what you need.

grzesuav:
I wanted to install operator via helm as I do not use OLM - like Prometheus operator have helm chart or manifest for its installation. I did not notice any manifest to setup cryostat operator as well.

My use case is simple, wanted to explore how much easier it would be to install cryostat via CRD however I do not see a way to install operator via helm of kubectl apply

tthvo:
> I wanted to install operator via helm as I do not use OLM - like Prometheus operator have helm chart or manifest for its installation. I did not notice any manifest to setup cryostat operator as well.
> 
> My use case is simple, wanted to explore how much easier it would be to install cryostat via CRD however I do not see a way to install operator via helm of kubectl apply

The operator is built as a Go-based operator. See https://sdk.operatorframework.io/docs/building-operators/golang/. Thus, it can't be installed with helm. I don't know if there is any plan for helm-based operator.

But if you are looking for a single yaml file to use with `kubectl apply`. There is a Make target `deploy` (i.e. run `make deploy`) that builds all necessary resources into a single k8s yaml and `kubectl apply` it. You can also build the yaml yourself with kustomize: `kubectl kustomize config/default`. Note: the namespace is default to: `cryostat-operator-system`. Though, this requires you to checkout the source code.

andrewazores:
There are no current plans for such a "helm chart to install the operator". It could be something we investigate, but adding yet another way to install and configure everything seems like it would be a relatively heavy maintenance burden on top of all the rest of the work to develop features and things. It could maybe be something we maintain on a best-effort basis upstream if someone provides it to begin with.

Distributing a `bundle.yaml` seems simple enough and just uses existing Operator tooling, so I can see that being a pretty realistic goal. We could just have a release action that generates this YAML bundle and publishes it somewhere.

tthvo:
> Distributing a bundle.yaml seems simple enough and just uses existing Operator tooling, so I can see that being a pretty realistic goal. We could just have a release action that generates this YAML bundle and publishes it somewhere.

[...]

Issue #144, opened 2020-12-21
"Installer Script" for setting up Operator


andrewazores:
Does it make sense to create a simpler, more "streamlined" feeling (un)installer script for setting up the Operator? This would be useful for us in the interim before we're published in any OperatorHub or similar registries, as well as useful for end users who want to install without going through these Hubs. This would basically just wrap around the existing various YAML files for CRDs and ServiceAccounts with the Makefile or some shell script that would basically do the same job as `make deploy` or `make undeploy`. The goal simply being to make installation "feel" easier than cloning the repo and running `make` instructions.

ebaron:
Something like a big YAML file with all the resources in one? Then you can just `kubectl/oc apply -f big.yaml`? This is what cert-manager does, and how we install it with `make cert-manager`.

andrewazores:
Yea, something like that - it was initially @jiekang's idea here, but I think the thought is just to have something as close to a "one-click install" as possible directly from GitHub. So one big YAML file that we can reference in the README sounds good to me, too. This should probably be tagged as a release here on GitHub so it's easy to refer to specific versions of it, too.

jiekang:
Yeah, basically and something we can deliver as a tarball. So download off release section, untar, run script. Or at least that was the idea.

jiekang:
A large yaml file in the repo that we instruct users to `kubectl/oc apply -f` is also good too. It's a little more transparent than the current makefile setup.

PR #903, opened 2024-06-27
docs(deploy): add instructions on how to obtain static manifests

tthvo:
# Welcome to Cryostat! 👋

Fixes: #901 
Fixes: #144 

## Description of the change:

- Added instructions on how to obtain static manifests

Note: `kubectl apply -f` would fail due to annotation size limit. Use `kubectl create -f` instead.

## Motivation for the change:

See #901.  A single static manifest that is bundled with all k8s resources allows users to install the operator with `kubectl apply/create -f <manifest-url>`. This aligns with common practice on upstream k8s.

## How to manually test:

You can follow the instructions in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 
 A Kubernetes Operator to automate deployment of
 [Cryostat](https://github.com/cryostatio/cryostat).
+
 ## SEE ALSO
 
 * [cryostat-core](https://github.com/cryostatio/cryostat-core) for
@@ -25,14 +26,14 @@ the JFR datasource for Grafana.
 * [cryostat-grafana-dashboard](https://github.com/cryostatio/cryostat-grafana-dashboard)
 for the Grafana dashboard.
 
-# Using
+## USING
 
-## Requirements
+### Requirements
 
 - `kubernetes` v1.21+ with [`Operator Lifecycle Manager`](https://olm.operatorframework.io/)
 - [`cert-manager`](https://github.com/cert-manager/cert-manager) v1.11.5+ (Recommended)
 
-## Instructions
+### Instructions
 
 Once deployed, the `cryostat` instance can be accessed via web browser
 at the URL provided by:
@@ -45,8 +46,9 @@ you may use the Cryostat web UI to define a Custom Target with the connection UR
 This is a special value which tells Cryostat's JVM that it should connect to itself directly, without
 the need to expose a JMX port over the network.
 
-# Building
-## Requirements
+## BUILDING
+
+### Requirements
 - `go` v1.21+
 - [`operator-sdk`](https://github.com/operator-framework/operator-sdk) v1.31.0
 - `podman` or `docker`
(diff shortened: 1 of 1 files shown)


tthvo:
This is a pretty (maybe too) simple solution. It uses the hard-coded namespace `cryostat-operator-system` that the user must create in-advance. I am thinking 2 possible ways around it:

- Maybe use a commonly available namespace like `default`?
- Or remove the namespace field? I am a bit unsure how to do this with kustomize.

andrewazores (commented):
I think it'd be nice to have this static manifest kept up-to-date automatically by some git commit hook or a GH action. Otherwise it's too easy to forget to update this and have it fall out of date with the rest of the sources it's meant to correspond to.

I think I would also still rather have it published some other way/place than having it checked in to the source repo. It is after all just a generated artifact, which can be reproduced from any given source repo commit. So checking it in to the repository only seems to serve the purpose of making it accessible via a GitHub URL.

andrewazores:
> Or maybe it does not matter much since others (e.g. ingress-nginx, cert-manager) use a specific namespace anyway. That being said, perhaps the YAML for namespace cryostat-operator-system should also be included in the static file?

Being completely honest, I think at some point if more customization and functionality is needed, then the end user needs to either use the Operator or Helm Chart, or they need to be comfortable modifying the static manifest file to suit their needs. It's likely that users have their own particular details to suit their own particular deployment scenarios, so we should make sure that it's possible for them to achieve what they need to, but carrying all of the customizations upstream seems a bit too much. I fear that if we add the namespace customization feature to the static manifest, then we will also be asked to add other features to the static manifest, and then we are just re-inventing the Helm Chart again.

tthvo:
> > Or maybe it does not matter much since others (e.g. ingress-nginx, cert-manager) use a specific namespace anyway. That being said, perhaps the YAML for namespace cryostat-operator-system should also be included in the static file?
> 
> Being completely honest, I think at some point if more customization and functionality is needed, then the end user needs to either use the Operator or Helm Chart, or they need to be comfortable modifying the static manifest file to suit their needs. It's likely that users have their own particular details to suit their own particular deployment scenarios, so we should make sure that it's possible for them to achieve what they need to, but carrying all of the customizations upstream seems a bit too much. I fear that if we add the namespace customization feature to the static manifest, then we will also be asked to add other features to the static manifest, and then we are just re-inventing the Helm Chart again.

Now, looking back, I proposed this solution before discovering we can use URL with kustomize: https://github.com/cryostatio/cryostat-operator/issues/901#issuecomment-2195766199. How about instead of this, I document that it is possible gain static manifests (out of `make deploy` without actually apply them to k8s)?

```bash
# Replace ref with any version tag or branch
kubectl kustomize https://github.com/cryostatio/cryostat-operator.git/config/default?ref=v2.4.0 | kubectl create -f
```

This way, we can avoid it all together to add more maintenance efforts.

andrewazores (approved):

ebaron (approved):

ebaron:
Thanks @tthvo!

Issue #841, opened 2024-06-04
[Bug] Scorecard test can fail with "failed to get Cryostat CR: resource name may not be empty"


ebaron:
I observed this failure once. Perhaps we can guard against it somehow.

```yaml
{
  "kind": "Test",
  "apiVersion": "scorecard.operatorframework.io/v1alpha3",
  "spec": {
    "image": "quay.io/cryostat/cryostat-operator-scorecard:3.0.0-20240511064726",
    "entrypoint": [
      "cryostat-scorecard-tests",
      "--installOpenShiftCertManager",
      "cryostat-config-change"
    ],
    "labels": {
      "suite": "cryostat",
      "test": "cryostat-config-change"
    },
    "storage": {
      "spec": {
        "mountPath": {}
      }
    }
  },
  "status": {
    "results": [
      {
        "name": "cryostat-config-change",
        "log": "OpenShift version is: 4.14.28
          OperatorHub patched to enable default catalog sources
          Created cert-manager-operator namespace
          Created OperatorGroup for cert-manager-operator
          Created Subscription for openshift-cert-manager-operator in cert-manager-operator
          ClusterServiceVersion cert-manager-operator.v1.13.1 successfully installed
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet found
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
          deployment cryostat-config-change is not yet available
[...]

PR #912, opened 2024-07-10
test(scorecard): preserve CR when retrying on conflict

ebaron:
# Welcome to Cryostat! 👋

Fixes: #841 

## Description of the change:
* Avoids overwriting the `cr` variable when calling `Update`. If this fails due to a conflict, the update will be silently retried, but `cr` will now be nil.
* Moves the update and retry to a helper function for convenience

## Motivation for the change:
* Fixes a sporadic scorecard test failure encountered downstream

## How to manually test:
1. `make test-scorecard-local SCORECARD_TEST_SELECTION=cryostat-config-change`
2. In order to reliably reproduce the update conflict, you can use a patch similar to the one listed here: https://github.com/cryostatio/cryostat-operator/issues/841#issuecomment-2221087146

### DIFF
--- a/internal/test/scorecard/common_utils.go
+++ b/internal/test/scorecard/common_utils.go
@@ -516,7 +516,7 @@ func (r *TestResources) updateAndWaitTillCryostatAvailable(cr *operatorv1beta2.C
 }
 
 func (r *TestResources) updateStorageOptions(ctx context.Context, cr *operatorv1beta2.Cryostat) (*operatorv1beta2.Cryostat, error) {
-	result := &operatorv1beta2.Cryostat{}
+	var result *operatorv1beta2.Cryostat
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		cr, err := r.Client.OperatorCRDs().Cryostats(cr.Namespace).Get(ctx, cr.Name)
 		if err != nil {
--- a/internal/test/scorecard/common_utils.go
+++ b/internal/test/scorecard/common_utils.go
@@ -453,29 +453,8 @@ func (r *TestResources) sendHealthRequest(base *url.URL, healthCheck func(resp *
 func (r *TestResources) updateAndWaitTillCryostatAvailable(cr *operatorv1beta2.Cryostat) (*operatorv1beta2.Cryostat, error) {
 	ctx := context.Background()
 
-	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		var err error
-		cr, err = r.Client.OperatorCRDs().Cryostats(cr.Namespace).Get(ctx, cr.Name)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


andrewazores:
https://github.com/cryostatio/cryostat-operator/actions/runs/9878971140

internal/test/scorecard/common_utils.go
```diff
@@ -453,29 +453,8 @@ func (r *TestResources) sendHealthRequest(base *url.URL, healthCheck func(resp *
...
-					Resources: corev1.ResourceRequirements{
-						Requests: corev1.ResourceList{
-							corev1.ResourceStorage: resource.MustParse("1Gi"),
-						},
-					},
-				},
-			},
-		}
-
-		cr, err = r.Client.OperatorCRDs().Cryostats(cr.Namespace).Update(context.Background(), cr)
```
andrewazores, line 476: If I'm understanding correctly, this is the part that was problematic? ie this `Update()` call can fail, which returns `nil, SomeError`. Since the `return err` now has a non-nil return value, the `retry.RetryOnConflict()` fires again, but the `cr` reference has now been set to `nil` and this second attempt fails for that reason?

tthvo: https://github.com/cryostatio/cryostat-operator/blob/f84caf0e9452ae570c9afdfae6bba368ccdf5f6a/internal/test/scorecard/clients.go#L138

https://github.com/cryostatio/cryostat-operator/blob/f84caf0e9452ae570c9afdfae6bba368ccdf5f6a/internal/test/scorecard/clients.go#L171-L181

Oh my, this is it. When there is an error, the empty CR is returned :D

ebaron: That's correct

ebaron: It was setting the `cr` to nil that exists outside the scope of this callback

internal/test/scorecard/common_utils.go
```diff
@@ -536,6 +515,33 @@ func (r *TestResources) updateAndWaitTillCryostatAvailable(cr *operatorv1beta2.C
 	return cr, err
 }
 
+func (r *TestResources) updateStorageOptions(ctx context.Context, cr *operatorv1beta2.Cryostat) (*operatorv1beta2.Cryostat, error) {
+	result := &operatorv1beta2.Cryostat{}
```
tthvo, line 519: I guess we don't need to initialize an empty CR? Perhaps, `var result *operatorv1beta2.Cryostat`?

This way, if any error, `nil` is returned and access to nil can throw nil pointer exception. This way, it is easier to understand other than an empty CR like this issue?

ebaron: Good call

andrewazores (approved):

PR #928, opened 2024-08-08
test(sample): update sample application images and configs

andrewazores:
- **test(sample): update sample application image**
- **test(sample): set up Agent sample application with new image and TLS config**

# Welcome to Cryostat! 👋

See https://github.com/cryostatio/test-applications/issues/1
See https://github.com/cryostatio/test-applications/issues/7
Related to https://github.com/cryostatio/cryostat-agent/issues/139
Depends on https://github.com/cryostatio/cryostat-agent/pull/491

## Description of the change:
Updates the two existing sample app definitions for the newer split TLS truststore/keystore where the Agent's stores are now independent from the host application's stores. Also adds a new sample app definition for an Agent that uses the nginx TLS client auth proxy, rather than the standard OAuth proxy.

## How to manually test:
1. `make sample_app sample_app_agent sample_app_agent_proxy` and ensure all three cases become discovered and are interactable as usual.
2. `make undeploy_sample_app undeploy_sample_app_agent undeploy_sample_app_agent_proxy` and ensure all three cases are correctly torn down and removed from discovery.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -436,11 +436,9 @@ sample_app_agent_proxy: undeploy_sample_app_agent_proxy ## Deploy sample app wit
 			echo "'SAMPLE_APP_NAMESPACE' must be specified."; \
 			exit 1; \
 		fi ;\
-		SECRET_HASH=`echo -n ${DEPLOY_NAMESPACE}/cryostat-sample/${SAMPLE_APP_NAMESPACE} | sha256sum`; \
+		SECRET_HASH=`echo -n ${DEPLOY_NAMESPACE}/cryostat-sample/${SAMPLE_APP_NAMESPACE} | sha256sum | cut -d' ' -f 1`; \
 	fi; \
-	$(CLUSTER_CLIENT) patch -f config/samples/sample-app-agent-tls-proxy.yaml --local=true --type=merge \
-	-p "{\"spec\":{\"template\":{\"spec\":{\"\$setElementOrder/volumes\":[{\"name\":\"agent-tls\"}],\"volumes\":[{\"\$retainKeys\":[\"name\",\"secret\"],\"name\":\"agent-tls\",\"secret\":{\"secretName\":\"cryostat-agent-$${SECRET_HASH}\"}}]}}}}" \
-	-o yaml | oc apply -f -
+	sed "s/REPLACEHASH/$${SECRET_HASH}/" < config/samples/sample-app-agent-tls-proxy.yaml | oc apply -f -
 
 .PHONY: undeploy_sample_app_agent
 undeploy_sample_app_agent: ## Undeploy sample app with Cryostat Agent.
--- a/Makefile
+++ b/Makefile
@@ -424,19 +424,10 @@ undeploy_sample_app_agent_proxy: ## Undeploy sample app with Cryostat Agent conf
 .PHONY: sample_app_agent_proxy
 sample_app_agent_proxy: undeploy_sample_app_agent_proxy ## Deploy sample app with Cryostat Agent configured for TLS client auth on nginx proxy.
 	@if [ -z "${SECRET_HASH}" ]; then \
-		if [ -z "${DEPLOY_NAMESPACE}" ]; then \
-			if [ "${CLUSTER_CLIENT}" = "oc" ]; then \
-				DEPLOY_NAMESPACE=`oc project -q`; \
-			else \
-				echo "'DEPLOY_NAMESPACE' must be specified."; \
-				exit 1; \
-			fi; \
-		fi; \
 		if [ -z "${SAMPLE_APP_NAMESPACE}" ]; then \
-			echo "'SAMPLE_APP_NAMESPACE' must be specified."; \
-			exit 1; \
+			SAMPLE_APP_NAMESPACE=`$(CLUSTER_CLIENT) config view --minify -o 'jsonpath={.contexts[0].context.namespace}'`; \
 		fi ;\
-		SECRET_HASH=`echo -n ${DEPLOY_NAMESPACE}/cryostat-sample/${SAMPLE_APP_NAMESPACE} | sha256sum | cut -d' ' -f 1`; \
+		SECRET_HASH=`echo -n ${DEPLOY_NAMESPACE}/cryostat-sample/${namespace} | sha256sum | cut -d' ' -f 1`; \
 	fi; \
 	sed "s/REPLACEHASH/$${SECRET_HASH}/" < config/samples/sample-app-agent-tls-proxy.yaml | oc apply -f -
 
--- a/Makefile
+++ b/Makefile
@@ -423,7 +423,24 @@ undeploy_sample_app_agent_proxy: ## Undeploy sample app with Cryostat Agent conf
 
 .PHONY: sample_app_agent_proxy
 sample_app_agent_proxy: undeploy_sample_app_agent_proxy ## Deploy sample app with Cryostat Agent configured for TLS client auth on nginx proxy.
-	$(CLUSTER_CLIENT) apply $(SAMPLE_APP_FLAGS) -f config/samples/sample-app-agent-tls-proxy.yaml
+	@if [ -z "${SECRET_HASH}" ]; then \
+		if [ -z "${DEPLOY_NAMESPACE}" ]; then \
+			if [ "${CLUSTER_CLIENT}" = "oc" ]; then \
+				DEPLOY_NAMESPACE=`oc project -q`; \
+			else \
+				echo "'DEPLOY_NAMESPACE' must be specified."; \
+				exit 1; \
+			fi; \
+		fi; \
+		if [ -z "${SAMPLE_APP_NAMESPACE}" ]; then \
+			echo "'SAMPLE_APP_NAMESPACE' must be specified."; \
+			exit 1; \
+		fi ;\
+		SECRET_HASH=`echo -n ${DEPLOY_NAMESPACE}/cryostat-sample/${SAMPLE_APP_NAMESPACE} | sha256sum`; \
+	fi; \
+	$(CLUSTER_CLIENT) patch -f config/samples/sample-app-agent-tls-proxy.yaml --local=true --type=merge \
+	-p "{\"spec\":{\"template\":{\"spec\":{\"\$setElementOrder/volumes\":[{\"name\":\"agent-tls\"}],\"volumes\":[{\"\$retainKeys\":[\"name\",\"secret\"],\"name\":\"agent-tls\",\"secret\":{\"secretName\":\"cryostat-agent-$${SECRET_HASH}\"}}]}}}}" \
+	-o yaml | oc apply -f -
 
 .PHONY: undeploy_sample_app_agent
 undeploy_sample_app_agent: ## Undeploy sample app with Cryostat Agent.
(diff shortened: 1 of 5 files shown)


andrewazores:
@mwangggg could you help figure out the new TLS setup to use here? Using the current configuration in this PR:

```yaml
    env:
        - name: KEYSTORE_PASS
          valueFrom:
            secretKeyRef:
              key: KEYSTORE_PASS
              name: cryostat-sample-keystore
        - name: JAVA_OPTS_APPEND
          value: |-
            -Dquarkus.http.host=0.0.0.0
            -Djava.util.logging.manager=org.jboss.logmanager.LogManager
            -Dcom.sun.management.jmxremote.port=9097
            -Dcom.sun.management.jmxremote.ssl=false
            -Dcom.sun.management.jmxremote.authenticate=false
            -javaagent:/deployments/app/cryostat-agent.jar
            -Dcryostat.agent.webclient.tls.truststore.cert[0].path=/var/run/secrets/myapp/truststore.p12
            -Dcryostat.agent.webclient.tls.truststore.cert[0].type=X.509
            -Dcryostat.agent.webclient.tls.truststore.cert[0].alias=cryostat-sample
        image: quay.io/redhat-java-monitoring/quarkus-cryostat-agent:latest
...
        volumeMounts:
        - mountPath: /var/run/secrets/myapp/truststore.p12
          name: truststore
          subPath: truststore.p12
      volumes:
      - name: truststore
        secret:
          secretName: cryostat-sample-tls
```

I get an Agent startup failure with logs like this:

```
Starting the Java application using /opt/jboss/container/java/run/run-java.sh ...
INFO exec -a "java" java -XX:MaxRAMPercentage=80.0 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+ExitOnOutOfMemoryError -Dquarkus.http.host=0.0.0.0 -cp "." -jar /deployments/quarkus-run.jar 
INFO running in /deployments
2024-08-08 18:59:56:263 +0000 [cryostat-agent-main] INFO io.cryostat.agent.Agent - Cryostat Agent starting...
2024-08-08 18:59:57:961 +0000 [cryostat-agent-main] ERROR io.cryostat.agent.Agent - Agent startup failure
java.lang.RuntimeException: java.security.cert.CertificateParsingException: signed fields invalid
	at io.cryostat.agent.MainModule.provideClientSslContext(MainModule.java:249)
	at io.cryostat.agent.MainModule_ProvideClientSslContextFactory.provideClientSslContext(MainModule_ProvideClientSslContextFactory.java:54)
	at io.cryostat.agent.MainModule_ProvideClientSslContextFactory.get(MainModule_ProvideClientSslContextFactory.java:43)
	at io.cryostat.agent.MainModule_ProvideClientSslContextFactory.get(MainModule_ProvideClientSslContextFactory.java:13)
	at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
	at io.cryostat.agent.MainModule_ProvideHttpClientFactory.get(MainModule_ProvideHttpClientFactory.java:55)
	at io.cryostat.agent.MainModule_ProvideHttpClientFactory.get(MainModule_ProvideHttpClientFactory.java:14)
	at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
	at io.cryostat.agent.MainModule_ProvideCryostatClientFactory.get(MainModule_ProvideCryostatClientFactory.java:64)
	at io.cryostat.agent.MainModule_ProvideCryostatClientFactory.get(MainModule_ProvideCryostatClientFactory.java:15)
	at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
	at io.cryostat.agent.MainModule_ProvideRegistrationFactory.get(MainModule_ProvideRegistrationFactory.java:83)
	at io.cryostat.agent.MainModule_ProvideRegistrationFactory.get(MainModule_ProvideRegistrationFactory.java:13)
	at dagger.internal.DoubleCheck.get(DoubleCheck.java:47)
	at io.cryostat.agent.DaggerAgent_Client$ClientImpl.registration(DaggerAgent_Client.java:299)
	at io.cryostat.agent.Agent.accept(Agent.java:225)
	at io.cryostat.agent.Agent.lambda$agentmain$0(Agent.java:159)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.security.cert.CertificateParsingException: signed fields invalid
	at java.base/sun.security.x509.X509CertImpl.parse(X509CertImpl.java:1774)
	at java.base/sun.security.x509.X509CertImpl.<init>(X509CertImpl.java:178)
	at java.base/sun.security.provider.X509Factory.engineGenerateCertificate(X509Factory.java:102)
	at java.base/java.security.cert.CertificateFactory.generateCertificate(CertificateFactory.java:355)
	at io.cryostat.agent.MainModule.provideClientSslContext(MainModule.java:187)
```

I wonder if I just have the wrong alias, or something.

andrewazores:
^ maybe the Agent should have some other config property to allow it to import a whole truststore as well? Like `cryostat.agent.webclient.tls.truststore.path=/path/to/init/truststore`, `cryostat.agent.webclient.tls.truststore.pass=changeit`. I think this could be used in conjunction with the new configs from https://github.com/cryostatio/cryostat-agent/pull/448  - if these properties aren't specified then it does the same thing as https://github.com/cryostatio/cryostat-agent/pull/448, but if they are specified then probably just [this bit of code](https://github.com/cryostatio/cryostat-agent/blob/938e8f1032415606f5a65cfae20fb377d1cd0b89/src/main/java/io/cryostat/agent/MainModule.java#L180) needs to be adjusted to start with the provided truststore?

mwangggg:
We could do that, or we could add the cryostat's cert to the quarkus-cryostat-agent container with a volume mount instead of the "keystore" truststore volume mount

mwangggg:
^I'll test that method out and see if it works- although giving users the option of importing a whole truststore is also a good idea.

andrewazores:
Ah, right, there's already a Secret that contains the bare cert we need, isn't there. That sounds like it should work well with your existing Agent work.

I think the init-truststore idea is still something worth considering - I can see that being useful for some users in the future. It'd be nice to allow specifying the whole truststore as a starting point instead of requiring users to "unwrap" it into its individual certs, if that isn't the form they already have available.

config/samples/sample-app-agent-tls-proxy.yaml
```diff
@@ -0,0 +1,95 @@
...
+        - mountPath: /var/run/secrets/io.cryostat/cryostat-agent
+          name: agent-tls
+      restartPolicy: Always
+      securityContext:
+        runAsNonRoot: true
+      volumes:
+      - name: agent-tls
+        secret:
+          # FIXME determine the secretName at deployment time in the Makefile and patch it here
+          secretName: cryostat-agent-f46ed1c40de4d61ac533fef337f7705ac39c8690f91a9cdca5185140f0455563
```
andrewazores, line 75: @ebaron any suggestion for how to do this better? The secret doesn't seem to have any useful labels that I could use to query for it in the Makefile. I thought maybe I could just list all Secrets in the current namespace and filter by name pattern, but that doesn't seem ideal.

ebaron: We could add labels to the secrets generated from cert-manager. It may be easier than that here though. The suffix here can be computed ahead of time. It's the SHA256 hash of `x/y/z` where:
* _x_ is the CR's namespace
* _y_ is the CR's name
* _z_ is the target namespace (e.g. this pod's namespace)

andrewazores: Done in the last updates.

Makefile
```diff
@@ -397,25 +397,48 @@ SAMPLE_APP_FLAGS += -n $(SAMPLE_APP_NAMESPACE)
...
+				DEPLOY_NAMESPACE=`oc project -q`; \
+			else \
+				echo "'DEPLOY_NAMESPACE' must be specified."; \
+				exit 1; \
+			fi; \
+		fi; \
+		if [ -z "${SAMPLE_APP_NAMESPACE}" ]; then \
+			echo "'SAMPLE_APP_NAMESPACE' must be specified."; \
+			exit 1; \
+		fi ;\
```
ebaron, line 438: `DEPLOY_NAMESPACE` should always be defined. It's defaulted to cryostat-operator-system if not explicitly set. I think the sample app's namespace should instead use whatever the current namespace is in your kubeconfig. There's an implementation-independent way to do this with `$(CLUSTER_CLIENT) config view --minify -o 'jsonpath={.contexts[0].context.namespace}'`.

Makefile
```diff
@@ -397,25 +397,48 @@ SAMPLE_APP_FLAGS += -n $(SAMPLE_APP_NAMESPACE)
...
 undeploy_sample_app: ## Undeploy sample app.
-	$(CLUSTER_CLIENT) delete $(SAMPLE_APP_FLAGS) --ignore-not-found=$(ignore-not-found) -f config/samples/sample-app.yaml
+	- $(CLUSTER_CLIENT) delete $(SAMPLE_APP_FLAGS) --ignore-not-found=$(ignore-not-found) -f config/samples/sample-app.yaml
 
 .PHONY: sample_app_agent
 sample_app_agent: undeploy_sample_app_agent ## Deploy sample app with Cryostat Agent.
 	@if [ -z "${AUTH_TOKEN}" ]; then \
 		if [ "${CLUSTER_CLIENT}" = "oc" ]; then\
-			AUTH_TOKEN=`oc whoami -t | base64`; \
+			AUTH_TOKEN=`oc whoami -t`; \
```
ebaron, line 411: Since the sample app is using the client cert now, we don't need this to be mandatory anymore.

andrewazores: The sample app using the client cert and the new proxy is a separate makefile target, so I think the token should still be required for this older one that is set up to still go through the OAuth proxy?

Makefile
```diff
@@ -397,25 +397,39 @@ SAMPLE_APP_FLAGS += -n $(SAMPLE_APP_NAMESPACE)
...
+undeploy_sample_app_agent_proxy: ## Undeploy sample app with Cryostat Agent configured for TLS client auth on nginx proxy.
+	- $(CLUSTER_CLIENT) delete $(SAMPLE_APP_FLAGS) --ignore-not-found=$(ignore-not-found) -f config/samples/sample-app-agent-tls-proxy.yaml
+
+.PHONY: sample_app_agent_proxy
+sample_app_agent_proxy: undeploy_sample_app_agent_proxy ## Deploy sample app with Cryostat Agent configured for TLS client auth on nginx proxy.
+	@if [ -z "${SECRET_HASH}" ]; then \
+		if [ -z "${SAMPLE_APP_NAMESPACE}" ]; then \
+			SAMPLE_APP_NAMESPACE=`$(CLUSTER_CLIENT) config view --minify -o 'jsonpath={.contexts[0].context.namespace}'`; \
+		fi ;\
+		SECRET_HASH=`echo -n ${DEPLOY_NAMESPACE}/cryostat-sample/${namespace} | sha256sum | cut -d' ' -f 1`; \
```
ebaron, line 430: This should be `${SAMPLE_APP_NAMESPACE}` instead of `${namespace}`

ebaron: I think `${SAMPLE_APP_NAMESPACE}` also needs an extra `$` in front on both line 427 and 430.

ebaron (approved):

Issue #936, opened 2024-09-05
[Story] Generate self-signed certificates for agents


ebaron:
### Describe the feature

The operator should generate certificates and keys using cert-manager for use by Cryostat agents in each target namespace. In conjunction with agent auto-configuration, the operator could use these certificates and keys to:
* Configure TLS on its HTTP server
* Authenticate with Cryostat using a form of client certificate authentication

Issue #784, opened 2024-04-03
[Epic] Application annotations for Cryostat Agent integrations


andrewazores:
### Describe the feature

There should be some defined annotation prefix (like `io.cryostat/`) and keys like `inject-agent`, `inject-tls-cert`. Users should be able to add these annotations to their applications (on the Deployment or Pod) within any Namespace associated with a Cryostat instance. When the Operator observes that an application has such an annotation, an integration feature should be enabled:

- Cryostat Agent injection. The Operator mounts a volume to the annotated application, containing the Cryostat Agent JAR.
  - if the Operator can detect the base image or framework that the application is using, and if we determine some safe method to do so (maybe with another optional annotation), the Operator will also modify the application's Deployment to automatically instrument the application with the Agent JAR, ie. by setting the `-javaagent` JVM flag as well as setting environment variables to point the Agent at the correct Cryostat instance
- TLS cert injection. This would be used primarily for https://github.com/cryostatio/cryostat-agent/issues/138 . If the user has requested Cryostat Agent injection to their application, they can use this second annotation to also request a TLS cert. This requires cert-manager integration. The Operator uses cert-manager to generate a TLS cert and places that into a ConfigMap or Secret and mounts that to the application. The Operator can also automatically add configuration to the application so that the injected Agent instance uses this new cert for its webserver.
  - the same TLS cert ConfigMap/Secret should also be mounted to the associated Cryostat server instance and included [in its TLS truststore](https://github.com/cryostatio/cryostat3/pull/330)
- the associated Cryostat instance's TLS certificate should also be mounted to the application so that the Agent's HTTP client can trust the certificate

This way we can automate end-to-end TLS in both directions between Cryostats and their Agent instances, with the user doing as little work as possible to enable this security feature.

Tasks:
- [x] #936
- [x] #941 
- [x] #952
- [x] https://github.com/cryostatio/cryostat-agent/issues/525
- [x] https://github.com/cryostatio/cryostat-agent/issues/533
- [x] https://github.com/cryostatio/cryostat-operator/issues/970
- [x] #971 
- [x] https://github.com/cryostatio/cryostat/issues/707
- [x] #992 
- [x] #993
- [x] #994
- [x] #995
- [x] #996
- [x] #1001

andrewazores:
Some more thoughts about the Agent injection/integration.

1. We want our Operator to be able to supply the Agent JAR to the application - the user should be able but not required to rebuild their application image containing our Agent.
[...]

PR #938, opened 2024-09-06
feat(tls): generate client certificates for agents

ebaron:
# Welcome to Cryostat! 👋

Fixes: #936 

## Description of the change:
* Creates a uniquely-named certificate for each target namespace for use by Cryostat agents within it. The certificate has a wildcard hostname within the [namespace subdomain](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records-1), and supports client and server usages.
   1. Create a Certificate in the install namespace with a cluster unique name (hashed `install_namespace/cr_name/target_namespace`)
   2. Once cert-manager creates the secret in the install namespace, copy it to the target namespace
   3. If the target namespace is removed, the Certificate and its secret will be deleted, along with the copy in the target namespace
   4. If the CR is deleted, the finalizer allows the operator to delete the secret copies from all target namespaces
* Adds a test for finalizing CA cert secret copies in target namespaces
* Makes use of the `TARGET_NAMESPACES` variable in the Makefile, previously used for ClusterCryostat

## Motivation for the change:
* This PR is part of the work to allow auto-configuration of the agent (#784) using client certificates as an alternative authentication mechanism. The certificate and key can also be used to set up TLS for the agents' HTTP servers.

## How to manually test:
1. Deploy PR
2. Create multi-namespace Cryostat
3. Add/remove namespaces
4. Verify certificates/secrets are created/deleted in expected namespaces

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -546,7 +546,9 @@ undeploy_bundle: operator-sdk ## Undeploy the controller in the bundle format wi
 
 .PHONY: create_cryostat_cr
 create_cryostat_cr: destroy_cryostat_cr ## Create a namespaced Cryostat instance.
-	$(CLUSTER_CLIENT) create -f config/samples/operator_v1beta2_cryostat.yaml
+	target_ns_json=$$(jq -nc '$$ARGS.positional' --args -- $(TARGET_NAMESPACES)) && \
+	$(CLUSTER_CLIENT) patch -f config/samples/operator_v1beta2_cryostat.yaml --local=true --type=merge \
+	-p "{\"spec\": {\"targetNamespaces\": $$target_ns_json}}" -o yaml | oc apply -f -
 
 .PHONY: destroy_cryostat_cr
 destroy_cryostat_cr: ## Delete a namespaced Cryostat instance.
--- a/internal/controllers/certmanager.go
+++ b/internal/controllers/certmanager.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	certv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	"github.com/cryostatio/cryostat-operator/internal/controllers/common"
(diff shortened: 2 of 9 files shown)


andrewazores:
I tried creating a Cryostat with `targetNamespaces: ['cryostat', 'apps1', 'apps2']`, where `cryostat` is also the installation namespace. Then I used `oc get secrets -n $ns` to check that the cert secret was created in each, which it was. Then I tried `oc delete project apps1` - `oc get secrets -n apps1` shows nothing (obviously), `oc get secrets -n cryostat` shows all three secrets still present. Then I did `oc new-project apps1` and `oc get secrets -n apps1`, `oc get secrets -n cryostat` - the secret still exists in the installation namespace, but did not get (re-)copied into `apps1`.

I guess for this to work the Operator would also need to be watching for any namespaces to be created/deleted in the cluster, so that it can reconcile by deleting the original secrets if namespaces are deleted, or recreating/re-copying them if a matching namespace is recreated too?

Otherwise, the change itself looks good, though it seems that scorecard tests are failing.

ebaron:
> I guess for this to work the Operator would also need to be watching for any namespaces to be created/deleted in the cluster, so that it can reconcile by deleting the original secrets if namespaces are deleted, or recreating/re-copying them if a matching namespace is recreated too?

I suppose other objects that are placed in target namespaces would suffer from the same issue (e.g. role bindings, CA secrets). This might be a bit tricky to handle. Since we can't have cross-namespace owner references, maybe we could do something like this:
1. Apply a label to all cross-namespace objects pointing to the namespace/name of the CR
2. Add a custom controller watch on cross-namepsace object types (e.g. secret, rolebinding) that checks for the label(s)
3. If the label exists, enqueue a reconcile request for that CR's namespace/name
4. Controller recreates any missing objects

As for deleting the certificate object from the install namespace for deleted namespaces, we could issue a get request for the namespace and, if it doesn't exist, delete the certificate and secret. I think the namespace deletion event should be captured with the above custom controller watch.

andrewazores (approved):

ebaron:
I filed #941 for the target namespace issue. I'll work on that next in a follow-up PR.

PR #974, opened 2024-11-21
feat(service): add part-of, instance selector labels and appProtocol

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits): `git commit -S -m "YOUR_COMMIT_MESSAGE"`
_______________________________________________

Fixes: #<issue number>
See https://github.com/cryostatio/cryostat-helm/pull/211
Related to https://github.com/cryostatio/cryostat-openshift-console-plugin/issues/2

## Description of the change:
*This change adds allows the users to provide...*

## Motivation for the change:
*This change is helpful because users may want to...*

## How to manually test:
1. *Insert steps here...*
2. *...*

PR #985, opened 2024-12-17
test(scorecard): adjust scorecard report generation expectation

andrewazores:
# Welcome to Cryostat! 👋

See https://github.com/cryostatio/cryostat/pull/698
See #974

## Description of the change:
Adjusts the test to not expect a full report JSON response body, on request, but only a job ID string response.

Since https://github.com/cryostatio/cryostat/pull/698 the response is expected to be an async job ID on first request, then later the server will emit a websocket notification containing the same ID to notify clients that the request has been completed. Then when the client makes an identical request again, the report is ready and cached, so the server responds with it this second time around.

In the old system the server would always respond with the report response, even though this could sometimes take considerable time to complete. Because of this long delay, with the server writing no response bytes at all in the interim, clients (especially web browsers) could time out waiting while the server was still processing.

Ideally this scorecard test should be adjusted to account for the entire async job API cycle, but this would mean establishing a websocket connection to the server to listen for the async job ID. I think this is worth looking into but I will do it separately, and just get this minimal fix in ASAP to get CI passing again.

## Motivation for the change:
Get scorecards passing again.

## How to manually test:
1. *Insert steps here...*
2. *...*

### DIFF
--- a/internal/test/scorecard/clients.go
+++ b/internal/test/scorecard/clients.go
@@ -422,7 +422,7 @@ func (client *RecordingClient) Delete(ctx context.Context, target *Target, recor
 	return nil
 }
 
-func (client *RecordingClient) GenerateReport(ctx context.Context, target *Target, recording *Recording) (map[string]interface{}, error) {
+func (client *RecordingClient) RequestReportGeneration(ctx context.Context, target *Target, recording *Recording) (*string, error) {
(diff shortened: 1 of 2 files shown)


ebaron (approved):
Looks good, would you mind filing an issue to test the full async report generation?

andrewazores:
#986

Issue #970, opened 2024-11-14
[Story] Mutate pods to inject and configure agent


ebaron:
We can use a [Mutating Admission Webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#mutatingadmissionwebhook) to modify pods as they're being created to inject the Cryostat agent using our init container, and configure it so that it connects to a selected Cryostat instance automatically. This feature can be opted-in on a specific pod by adding labels that identify which Cryostat instance the pod should work with.

PR #991, opened 2024-12-20
feat(agent): add pod mutation webhook to inject agent

ebaron:
# Welcome to Cryostat! 👋

Fixes: #970 

## Description of the change:
* Adds a mutating admission webhook for pods that responds to a pair of labels by injecting the Cryostat agent and configuring it to work with the specified Cryostat instance.

## Motivation for the change:
* Makes it much easier to set up the Cryostat agent with arbitrary pods.

## How to manually test:
1. `make deploy`
2. `make create_cryostat_cr`
3. `make sample_app`
4. Modify the sample app deployment as follows:
    1. Override `JAVA_OPTS_APPEND` to not inject the built-in Cryostat agent.
        ```
        - env:
          - name: JAVA_OPTS_APPEND
            value: -Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -Dio.cryostat.agent.shaded.org.slf4j.simpleLogger.defaultLogLevel=debug
        ```
    2. Add the following labels:
        ```
        cryostat.io/name: cryostat-sample
        cryostat.io/namespace: cryostat-operator-system
        ```
    3. I had to also increase the memory limit to prevent OOM.
5. The sample application should appear as an agent target in Cryostat

I've filed issues for the TODOs in this PR. There are a few of them, but they're all pretty small compared to this PR:
#992, #993, #994, #995, #996


andrewazores:
I added a config sample to do `DEPLOY_NAMESPACE=$(oc project -q) make sample_app_agent_injected` for test convenience. Seems to work nicely!

internal/controllers/common/common_utils.go
```diff
@@ -77,21 +89,46 @@ func ClusterUniqueNameWithPrefix(gvk *schema.GroupVersionKind, prefix string, na
...
 	targetNS string) string {
+	return clusterUniqueName(gvk, prefix, name, namespace, targetNS, false)
+}
+
+func clusterUniqueName(gvk *schema.GroupVersionKind, prefix string, name string, namespace string, targetNS string,
+	short bool) string {
 	prefixWithKind := strings.ToLower(gvk.Kind)
+	if short {
+		// Prefix is 5 characters: 4 from kind, and hyphen
+		prefixWithKind = prefixWithKind[:4] // TODO can we use a weaker and shorter hash and keep the full kind?
```
ebaron, line 112: @andrewazores what do you think about using a non-cryptographic hash function for this instead?
With https://pkg.go.dev/hash/fnv, we could use:
`cryostat-<128-bit hash>` instead of `cryo-<224-bit hash>`

andrewazores: Do you mean just swapping out the hash functions, or would this also include changing the prefix before the hash or changing the hash inputs?

internal/controllers/services.go
```diff
@@ -138,6 +141,70 @@ func (r *Reconciler) reconcileAgentService(ctx context.Context, cr *model.Cryost
 	})
 }
 
+func (r *Reconciler) newAgentHeadlessService(cr *model.CryostatInstance, namespace string) *corev1.Service {
+	return &corev1.Service{
```
andrewazores, line 145: If I understand correctly, this Service is made headless by having a `ClusterIP: nil` in its spec. I guess here that is done implicitly, but would it be worth setting it explicitly for readability?

internal/webhooks/agent/pod_defaulter.go
```diff
@@ -0,0 +1,363 @@
...
+	existing, err := findJavaToolOptions(envs)
+	if err != nil {
+		return nil, err
+	}
+
+	if existing != nil {
+		existing.Value += " " + agentArg
+	} else {
+		envs = append(envs, corev1.EnvVar{
+			Name:  "JAVA_TOOL_OPTIONS",
```
andrewazores, line 334: Maybe this could be another label-based extension point? Just in case the application is already setting this to some other value - ex in its Deployment, for the purpose of launching another Agent. The user might have another environment variable they can use for a similar purpose.

And also, maybe here we can check if the `JAVA_TOOL_OPTIONS` is already set in the container's env vars? If it is then I think at least emitting some warning would be good, but I think it's also likely to be OK to simply append our arg line to it rather than overwriting it.

andrewazores: As I continue thinking on this I can't really come up with a scenario where this won't work. Maybe if the application has some strange entrypoint script and bootup logic that would cause this environment variable setting to get overridden on that side?

internal/webhooks/agent/pod_defaulter.go
```diff
@@ -0,0 +1,363 @@
...
+			if env.ValueFrom != nil {
+				return nil, errJavaToolOptionsValueFrom
+			}
+			return &envs[i], nil
+		}
+	}
+	return nil, nil
+}
+
+func isTargetNamespace(targetNamespaces []string, podNamespace string) bool {
```
andrewazores, line 356: Maybe this can be replaced with https://pkg.go.dev/slices#Contains ?

andrewazores:
Other than the minor comments above this seems to be working quite well. Exciting to see!

internal/controllers/common/common_utils.go
ebaron: My goal was to fit the entire `cryostat` prefix since it's more meaningful to a user than just `cryo`, but with SHA224 the name is still too long.

ebaron: And I'm talking specifically just for this new ClusterUniqueShortName introduced in this PR, so it wouldn't affect any existing released objects.

andrewazores: Okay, I think that sounds fine then.

internal/webhooks/agent/pod_defaulter.go
ebaron: You mean a label to use a different environment variable than JAVA_TOOL_OPTIONS? That would be a good idea, and allow an easy way to use `MODULE_OPTS` for Wildfly.

internal/controllers/services.go
ebaron: Instead of `nil`, it's the string `"None"` this is set a bit further down in the file:
https://github.com/cryostatio/cryostat-operator/blob/c5bb6b94966a02ccec94e984431fe34ed9541cd8/internal/controllers/services.go#L177

internal/webhooks/agent/pod_defaulter.go
andrewazores: Yea, something like that. I'm just cautious about making any hardcoded assumptions about what environment variables are usable for the end user's workload.

internal/webhooks/agent/pod_defaulter.go
ebaron: Cool! I wasn't aware of the slices package.

internal/controllers/common/common_utils.go
ebaron: I've added a new commit that switches the agent headless service to use 128-bit FNV-1 for its name, with the prefix of `cryostat-agent`. This should make a lot more sense to users looking at their services. I also added some labels to help select it, since it's not easy (I think) to get an FNV-1 hash on the command line:
```
oc get svc -lapp=cryostat-sample,component=agent
NAME                                              TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)    AGE
cryostat-agent-80edadf28f94839214cd32275600ca7c   ClusterIP   None         <none>        9977/TCP   3m6s
```

internal/webhooks/agent/pod_defaulter.go
ebaron: I've filed #1001 for this.

andrewazores (approved):
Just needs rebasing. I'm relieved to see that there are no conflicts with #923 :-)

Issue #1004, opened 2025-01-14
[Task] Split PVCs


andrewazores:
Could we extract this to a helper method and replace all 3 instances?

On that note, I suppose we can't get away with one PVC since these pods may be scheduled on other nodes and we default to the `ReadWriteOnce` mode. Maybe we could force all the pods to be on the same node using [PodAffinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity)? We would need to merge this with whatever config the user provides in Scheduling Options, and that could get a bit tricky.

Assuming we stick with 3 PVCs, do you think it would make sense to have separate configurations for each PVC? I could see object storage requiring quite a bit larger size for recordings than the Postgres database. What is the main Cryostat pod's PVC used for now?

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-operator/pull/923#discussion_r1911097023_

andrewazores:
See also https://github.com/cryostatio/cryostat-helm/pull/224

andrewazores:
Done in #1010

PR #1010, opened 2025-01-14
feat(pvc): split PVCs for database and storage

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1004

## Description of the change:

## Motivation for the change:

## How to manually test:
1. *Insert steps here...*
2. *...*


andrewazores:
https://github.com/cryostatio/cryostat-operator/actions/runs/12776809750

andrewazores:
```
	Name: olm-spec-descriptors
	State: fail

	Suggestions:
		Add a spec descriptor for storageOptions
	Errors:
		storageOptions does not have a spec descriptor
```
----
```
 	Name: olm-crds-have-validation
	State: fail

	Suggestions:
		Add CRD validation for spec field `storageOptions` in Cryostat/v1beta2
```

@ebaron is this because I renamed the CRD property in v1beta2 compared to what v1beta1 had? `storageOptions` no longer appears in v1beta2 so I'm not really sure what I'm seeing here.

ebaron:
I'm not sure if this is causing the test failure, but I don't think we can break API by renaming an existing property in v1beta2:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md

I think the proper way would be to deprecate the old property, but still use it as a fallback if the new properties aren't given.
Something like this in terms of precedence:
1. Database:
    i. `spec.storageConfig.Database`
    ii. `spec.storageOptions`
    iii. Operator defaults 
3. Object Storage:
    i. `spec.storageConfig.ObjectStorage`
    ii. `spec.storageOptions`
    iii. Operator defaults

andrewazores:
@ebaron any idea what's going on with the envtest failure? I don't think I really understand it.

ebaron:
> @ebaron any idea what's going on with the envtest failure? I don't think I really understand it.

I should have fixed it. Before this test was sensitive to the order things were created in the reconciler, it should be more robust now.

ebaron (approved):

Issue #814, opened 2024-05-08
[Story] Deploy database and storage containers separately


andrewazores:
Initial round of this is done now. In the future the database and storage containers should be pulled out into their own Pods as well.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-operator/issues/710#issuecomment-2101361753_

- [x] #923 
- [ ] #1004 
- [ ] #1012
- [ ] #1005
- [x] #1006
- [x] #1007
- [x] #806 
- [ ] #1008

andrewazores:
Related: #806

andrewazores:
#923

PR #1017, opened 2025-01-16
feat(network): implement ingress network policies

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1008
See #814
See also https://github.com/cryostatio/cryostat-helm/pull/208

## Description of the change:
Mirrors the `NetworkPolicy` implementation for ingress traffic from the Helm PR above.

## Motivation for the change:
Adds network-level isolation to various Pods created by the Operator, so that traffic to each is only allowed from expected origins (assuming the cluster supports this feature). This prevents unexpected connections to the database or storage, which could result in data being leaked if authentication is not configured or is somehow bypassed.

## How to manually test:
1. Check out and build PR
2. Deploy, then create a Cryostat CR with 1 reports replica
3. Wait for everything to be reconciled
4. `oc describe networkpolicy`
5. See other testing steps in Helm PR, ex. use `oc run` and try to `curl` requests to `http://cryostat-sample-reports:10000` or `http://cryostat-sample-storage:8333`.


andrewazores:
https://github.com/cryostatio/cryostat-operator/actions/runs/12813863402

internal/controllers/networkpolicy.go
```diff
@@ -0,0 +1,215 @@
...
+			},
+			Ingress: []networkingv1.NetworkPolicyIngressRule{
+				networkingv1.NetworkPolicyIngressRule{
+					From: []networkingv1.NetworkPolicyPeer{
+						AllNamespacesSelector,
+						RouteSelector,
+					},
+					Ports: []networkingv1.NetworkPolicyPort{
+						networkingv1.NetworkPolicyPort{
+							Port: &intstr.IntOrString{IntVal: constants.AuthProxyHttpContainerPort},
```
ebaron, line 72: Do we need a policy for the agent gateway (nginx) port as well? This could theoretically be limited to inbound traffic from the target namespaces of the CR

andrewazores: Makes sense, I'll add that in

andrewazores: Opened as a separate PR: #1022

ebaron (approved):
Seems to work well. I added a commit with a couple of changes:
* Delete the NetworkPolicy object when the `disabled` property is true
* Update the NetworkPolicy spec by placing it in the body of the `CreateOrUpdate...` method

What do you think?

andrewazores:
> Seems to work well. I added a commit with a couple of changes:
> 
>     * Delete the NetworkPolicy object when the `disabled` property is true
> 
>     * Update the NetworkPolicy spec by placing it in the body of the `CreateOrUpdate...` method
> 
> 
> What do you think?

Nice, thanks. That looks good.

PR #1019, opened 2025-01-17
feat(db): enable JDBC TLS configuration

andrewazores:
# Welcome to Cryostat! 👋
## Before contributing, make sure you have:
* [x] Read the [contributing guidelines](https://github.com/cryostatio/cryostat/blob/main/CONTRIBUTING.md)
* [x] Linked a relevant issue which this PR resolves
* [x] Linked any other relevant issues, PR's, or documentation, if any
* [x] Resolved all conflicts, if any
* [x] Rebased your branch PR on top of the latest upstream `main` branch
* [x] Attached at least one of the following labels to the PR: `[chore, ci, docs, feat, fix, test]`
* [x] [Signed all commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits): `git commit -S -m "YOUR_COMMIT_MESSAGE"`
_______________________________________________

Related to #1005

## Description of the change:
Enables JDBC TLS configuration for the connection between Cryostat and the database container. This sets up TLS on the database's server side, and sets up Cryostat's database driver to expect to find a TLS port there and to require that the connection passes all checks (cert from trusted CA, hostname matches, SSL/TLS must be enabled). Authentication from the client (Cryostat) to the database server is still using normal Basic-style authentication, not TLS client certificate authentication.

If TLS is disabled, ie cert-manager is not available, then this should fall back to the same previous unencrypted JDBC behaviour.

## Motivation for the change:
Enhances security.

## How to manually test:
1. Check out and build PR
2. Install cert-manager
3. Deploy Operator
4. Create Cryostat CR
5. Everything should continue to work as expected. Verify that the Cryostat container's environment variables specify the JDBC URL with SSL parameters: `jdbc:postgresql://cryostat-sample-database.cryostat.svc.cluster.local:5432/cryostat?ssl=true&sslmode=verify-full&sslcert=&sslrootcert=/var/run/secrets/operator.cryostat.io/cryostat-sample-database-tls/ca.crt`. In particular `ssl=true` and `sslmode=verify-full` should be present, and the `sslrootcert=` should point to the database TLS secret and the `ca.crt` corresponding to the cert used by the database container.
6. Edit Cryostat CR and set `.spec.enableCertManager: false`. After everything gets redeployed, ensure it's all still working, but now the JDBC URL has no SSL parameters, the database container has no SSL arguments, etc.

Issue #1005, opened 2025-01-14
[Task] Add TLS configurations for split Deployments


andrewazores:

andrewazores:
Done in #1019 and #1020

PR #1020, opened 2025-01-20
feat(storage): enable storage TLS configuration

andrewazores:
# Welcome to Cryostat! 👋

Related to #1005
See also #1019

## Description of the change:
Enables TLS (HTTPS) for the connection between Cryostat and the storage container. This sets up TLS on the storage server side using the existing storage TLS secret (key and cert) so that the SeaweedFS S3 service is listening on both HTTP and HTTPS (ports 8333 and 8334 respectively). If TLS is not enabled then it listens only on HTTP 8333 as before. Cryostat will use the HTTPS port if TLS is enabled, or the HTTP port if not. Client authentication is done by the HTTP Basic style as before, not TLS client certificate authentication.

Currently, Cryostat gets configured to trust all certificates encountered by the S3 client. This should be enhanced to trust specifically the certificate within the storage TLS secret. See https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-s3.html#quarkus-amazon-s3_section_quarkus-s3-sync-client and the `quarkus.s3.sync-client.tls-trust-managers-provider.type` property.

## Motivation for the change:
Enhances security.

## How to manually test:
1. Check out and build PR
2. Install cert-manager
3. Deploy Operator
4. Create Cryostat CR
5. Everything should continue to work as expected. Verify that the Cryostat container's environment variables specify the S3 storage URL with SSL markers: `https` scheme and `8334` port number. Try listing custom event templates, or all archives, or try creating a recording on Cryostat itself (`localhost:0`) or a sample application and then archiving the recording to verify that Cryostat is able to talk to the storage container.
6. Edit Cryostat CR and set `.spec.enableCertManager: false`. After everything gets redeployed, ensure it's all still working, but now the storage URL uses `http` and `8333`, the storage container has no HTTPS arguments, etc.


api/v1beta2/cryostat_types.go
```diff
@@ -344,7 +344,12 @@ type StorageServiceConfig struct {
 	// HTTP port number for the storage to be created by the operator.
 	// Defaults to 8333.
 	// +optional
-	HTTPPort      *int32 `json:"httpPort,omitempty"`
+	HTTPPort *int32 `json:"httpPort,omitempty"`
+	// HTTPS port number for the storage to be created by the operator.
+	// Only used if TLS is enabled.
+	// Defaults to 8334.
+	// +optional
+	HTTPSPort     *int32 `json:"httpsPort,omitempty"`
```
ebaron, line 352: What about using `HTTPPort` whether TLS is enabled or not? For example, with Reports, we have just the one HTTPPort property.

andrewazores: I thought about that, but it doesn't seem that SeaweedFS' S3 implementation allows for completely turning off the HTTP port. HTTPS can only be turned on as an additional port. So, I figured I would reflect that here. I can remove it from the API and just "hide"/ignore the HTTP port internally when HTTPS is enabled, though, if you think that'd be better.

ebaron: Hiding it might be better. We could remove it from the Service, and also block traffic to it using the network policy?

ebaron: I might be misreading this, but it looks like if the HTTPS port isn't provided, but the private key is, it will serve HTTPS over the configured HTTP port:
https://github.com/seaweedfs/seaweedfs/blob/95357df8dda57baba53b15dad6efcb6be9a2c123/weed/command/s3.go#L338-L367

andrewazores: Nice find! I tried something like that with just `podman run` and the container image, but I think I only tried setting the HTTPS port to a blank value. This way seems to work well.

internal/controllers/networkpolicy.go
```diff
@@ -176,7 +178,9 @@ func (r *Reconciler) reconcileStorageNetworkPolicy(ctx context.Context, cr *mode
 					},
 					Ports: []networkingv1.NetworkPolicyPort{
 						{
-							Port: &intstr.IntOrString{IntVal: constants.StoragePort},
+							Port: &intstr.IntOrString{
+								IntVal: getServicePortOrDefault(*serviceSpecs.StorageURL, constants.StoragePort),
```
ebaron, line 182: The network policies work using the container ports, right? If so, referencing the service port here may not be what we want.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -1763,7 +1763,8 @@ func NewStorageContainer(cr *model.CryostatInstance, imageTag string, tls *TLSCo
 	args := []string{}
 	if tls != nil {
 		args = append(args,
-			fmt.Sprintf("-s3.port.https=%d", constants.StoragePortTLS),
+			fmt.Sprintf("-s3.port=%d", constants.StoragePort),
+			"-s3.port.https=0",
```
ebaron, line 1767: This might be a bit confusing to come across later down the road. Maybe we could either remove the explicit `s3.port.https` option, or give a comment that explains that the `s3.port` value will be used in this case?

internal/controllers/services.go
```diff
@@ -191,37 +191,31 @@ func (r *Reconciler) reconcileStorageService(ctx context.Context, cr *model.Cryo
...
+	}
 	err := r.createOrUpdateService(ctx, svc, cr.Object, &config.ServiceConfig, func() error {
 		svc.Spec.Selector = map[string]string{
 			"app":       cr.Name,
 			"component": "storage",
 		}
 		svc.Spec.Ports = []corev1.ServicePort{
 			{
-				Name:       "http",
+				Name:       name,
```
ebaron, line 206: There's a `constants.HttpPortName` that might be better here.

ebaron (approved):
Looks good! Just removed an unused parameter.

PR #1043, opened 2025-02-05
build(versions): bump versions for 4.1.0 development

ebaron:


.mergify.yml
andrewazores, line 1: The rule name at line 2 should be updated as well

andrewazores (commented):
```
config/overlays/insights/insights_patch.yaml
18:          value: "cryostat-operator/4.0.0-dev"
```

Does that need to be bumped too?

ebaron:
I've changed the Makefile to unconditionally regenerate the Insights patch, and also bumped the APICast version

andrewazores (approved):

PR #1076, opened 2025-03-13
feat(agent): add toggle to allow insecure connections

andrewazores:
# Welcome to Cryostat! 👋

See #1011

## Description of the change:
Adds a simple toggle to set `cryostat.agent.tls.required=false`.

## Motivation for the change:
For manual (non-autoconfig) Agent setups, in testing or prototyping scenarios, it can be difficult or complicated to set up TLS fully. This toggle can be used to allow insecure HTTP Agent callback URLs again, since 4.0 enforces this requirement by default whereas prior Cryostat versions did not. This would also be required for agent autoconfiguration to work if cert-manager integration is disabled, though I haven't specifically tested that scenario yet and more might need to be done.

## How to manually test:
1. *Insert steps here...*
2. *...*

### DIFF
--- a/api/v1beta2/cryostat_types.go
+++ b/api/v1beta2/cryostat_types.go
@@ -744,6 +744,10 @@ type AgentOptions struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Disable Hostname Verification",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
 	DisableHostnameVerification bool `json:"disableHostnameVerification,omitempty"`
+	// Allow insecure (non-TLS) HTTP connections to Cryostat Agents.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Allow Insecure Connections",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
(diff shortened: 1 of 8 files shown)


andrewazores:
https://github.com/cryostatio/cryostat-operator/actions/runs/13841168945

ebaron (commented):
Should we tie this into the `enableCertManager` toggle as well? Does it make sense to have one imply the other?

andrewazores:
They're definitely related, and maybe if `enableCertManager: false` should imply `allowInsecure: true`, or at least should imply setting the same `CRYOSTAT_AGENT_TLS_REQUIRED` value. Would it make sense to put that implication into the defaulter webhook?

ebaron:
> They're definitely related, and maybe if `enableCertManager: false` should imply `allowInsecure: true`, or at least should imply setting the same `CRYOSTAT_AGENT_TLS_REQUIRED` value. Would it make sense to put that implication into the defaulter webhook?

On second thought, it does seem valid to have TLS agent connections and cert-manager disabled. Users can add certs to Cryostat's truststore even if cert-manager is disabled:
https://github.com/cryostatio/cryostat-operator/blob/ecfc6f3170ef37dfb818d200115ea72255d4d982/internal/controllers/common/resource_definitions/resource_definitions.go#L440-L463

ebaron (approved):

Issue #1100, opened 2025-04-24
[Bug] Reports Sidecars do not receive requests from Cryostat


andrewazores:
### Current Behavior

Cryostat was refactored some time ago and the HTTP client used within the Report Sidecar client implementation was changed. Along with that, the configuration property key for setting the sidecar Service URL was changed.

Required format:
`QUARKUS_REST_CLIENT_REPORTS_URL=http://reports:10001`

The Operator uses:
`CRYOSTAT_SERVICES_REPORTS_URL=https://cryostat-sample-reports:10000`
(https://github.com/cryostatio/cryostat-operator/blob/5b7f2a9c674eb212fe4d52eee1224cf396357992/internal/controllers/common/resource_definitions/resource_definitions.go#L1400)

### Environment

```Markdown
- OS: 
- Environment: 
- Version:
```

andrewazores:
Shockingly, and very unfortunately, this has been broken for a _long_ time in the Operator and we just never detected it, since in usual testing Cryostat is able to handle report generation just fine.

This configuration key was changed in https://github.com/cryostatio/cryostat/pull/383 .

PR #1101, opened 2025-04-25
fix(reports): correct reports service URL configuration key

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #1100

## Description of the change:
Corrects the configuration key (environment variable name) that the Operator sets to inform Cryostat about the URL for the report generator sidecar service.

## Motivation for the change:
The key currently used is outdated and no longer used, so Cryostat ignores it and handles report generation on its own rather than shifting it onto the report generator instance(s). This means that report generator configuration is currently useless.

## How to manually test:
1. Check out and build PR
2. Deploy Operator, create Cryostat CR with `reportOptions: { replicas: 1 }`
3. `make sample_app` in a target namespace
4. Wait for everything to come up, then open the Cryostat UI and request a target analysis on the sample app, or request analysis of an archived recording.
5. Check the reports Pod logs and ensure that an HTTP access log line appears corresponding to a `POST` request to handle a report.
6. Currently, the requests will fail with an HTTP 502 due to https://github.com/cryostatio/cryostat/pull/856 . ~~Separate fix incoming. This one needs to be backported, that one will not.~~ Separate fix: https://github.com/cryostatio/cryostat-reports/pull/345

### DIFF
--- a/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/cryostat-operator.clusterserviceversion.yaml
@@ -24,7 +24,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: Monitoring, Developer Tools
     containerImage: quay.io/cryostat/cryostat-operator:4.1.0-dev
-    createdAt: "2025-03-13T17:39:42Z"
+    createdAt: "2025-04-25T12:49:59Z"
     description: JVM monitoring and profiling tool
     operatorframework.io/initialization-resource: |-
       {
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -1397,7 +1397,7 @@ func NewCoreContainer(cr *model.CryostatInstance, specs *ServiceSpecs, imageTag 
 	if specs.ReportsURL != nil {
 		reportsEnvs := []corev1.EnvVar{
 			{
-				Name:  "CRYOSTAT_SERVICES_REPORTS_URL",
+				Name:  "QUARKUS_REST_CLIENT_REPORTS_URL",
 				Value: specs.ReportsURL.String(),
 			},
 		}
--- a/internal/test/resources.go
+++ b/internal/test/resources.go
@@ -2445,7 +2445,7 @@ func (r *TestResources) NewCoreEnvironmentVariables(reportsUrl string, ingress b
 	if reportsUrl != "" {
 		envs = append(envs,
 			corev1.EnvVar{
-				Name:  "CRYOSTAT_SERVICES_REPORTS_URL",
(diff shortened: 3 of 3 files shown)


ebaron (approved):
Works as described :+1:

andrewazores:
Seeing a scorecard test failure after push:

https://github.com/cryostatio/cryostat-operator/actions/runs/14735761920/job/41361322262

This looks the same or very similar to the failure I've been seeing over at https://github.com/cryostatio/cryostat-operator/pull/1102 , which seems unrelated to either changeset. I guess something else in `main` broke this at some point?

ebaron:
> Seeing a scorecard test failure after push:
> 
> https://github.com/cryostatio/cryostat-operator/actions/runs/14735761920/job/41361322262
> 
> This looks the same or very similar to the failure I've been seeing over at #1102 , which seems unrelated to either changeset. I guess something else in `main` broke this at some point?

Nothing is jumping out as an obvious cause of this. It doesn't look like we've touched the `pods` permissions in any Role/ClusterRoles recently.

andrewazores:
It also seems like this failure happens in CI consistently, but I can't seem to reproduce it locally.

ebaron:
I seem to be able to reproduce it. Looks like only the `cryostat-recording` test fails. I'm thinking the RBAC exceptions are actually a red herring, and the exception we're seeing in the logs is after the RBAC has been deleted for the test. The error message is an HTTP 500 when trying to generate a report. You said the main is broken at the moment. Could that be the cause?
```
Image:      quay.io/cryostat/cryostat-operator-scorecard:4.1.0-20250226193258
Labels:
    "suite":"cryostat"
    "test":"cryostat-recording"
Results:
    Name: cryostat-recording
    State: fail
    
    Errors:
        failed to generate report for the recording: API request failed with status code: 500, response body: {"details":"Error id a870f33f-9a20-4595-a864-897b59b9abce-1","stack":""}, and headers:
Cache-Control: no-cache
Content-Length: 72
Content-Type: application/json; charset=utf-8
Gap-Auth: system:serviceaccount:cryostat-operator-scorecard:cryostat-scorecard@cluster.local
Gap-Upstream-Address: localhost:8181
Date: Wed, 30 Apr 2025 16:35:19 GMT
Set-Cookie: c85e25706896056aeed77e60677084df=09bbeb9d528cfecd0e70ce315ce5165b; path=/; HttpOnly; Secure; SameSite=None
```

andrewazores:
Ah, yes, that does sound like it would make sense. I got too hung up on the RBAC message and didn't even consider that it could just be a red herring from after cleanup. The report generation will fail right now and that HTTP 500 response matches what I would expect, which should be fixed by https://github.com/cryostatio/cryostat-operator/pull/1102 . But over in that PR the latest scorecard run also looks like it's failing in the same way during `cryostat-recording`, which is unexpected.

PR #1105, opened 2025-04-30
chore(scorecard): add '-list' flag and implement wildcard test selection

andrewazores:
- **chore(scorecard): add '-list' flag and implement wildcard test selection**
- **refactor, mapping of test names to functions happens in one place only**
- **allow scorecard to handle multiple test names**


andrewazores:
https://github.com/cryostatio/cryostat-operator/pull/1101#issuecomment-2842691272

This PR is also having the same apparent RBAC issue.

ebaron (approved):

PR #1137, opened 2025-07-25
fix(cryostat): set QUARKUS_HTTP_HOST to 0.0.0.0

andrewazores:
# Welcome to Cryostat! 👋

See cryostatio/cryostat-helm#255
See cryostatio/cryostat-helm#256

## Description of the change:

## Motivation for the change:

## How to manually test:
1. *Insert steps here...*
2. *...*

### DIFF
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -1339,7 +1339,7 @@ func NewCoreContainer(cr *model.CryostatInstance, specs *ServiceSpecs, imageTag 
 	envs := []corev1.EnvVar{
 		{
 			Name:  "QUARKUS_HTTP_HOST",
-			Value: "localhost",
+			Value: "0.0.0.0",
 		},
 		{
 			Name:  "QUARKUS_HTTP_PORT",
--- a/internal/test/resources.go
+++ b/internal/test/resources.go
@@ -2453,7 +2453,7 @@ func (r *TestResources) NewCoreEnvironmentVariables(reportsUrl string, ingress b
 	envs := []corev1.EnvVar{
 		{
 			Name:  "QUARKUS_HTTP_HOST",
-			Value: "localhost",
+			Value: "0.0.0.0",
 		},
 		{
 			Name:  "QUARKUS_HTTP_PORT",


andrewazores:
https://github.com/cryostatio/cryostat-operator/actions/runs/16529291699

ebaron (commented):
Strange that this would suddenly become a problem. Do you know if opening up the host allows for bypassing the OAuth Proxy (with network policies disabled)?

andrewazores:
> Strange that this would suddenly become a problem.

It's not so sudden, I made a change to Cryostat's `Dockerfile` (explained in https://github.com/cryostatio/cryostat-helm/pull/256) which surfaced this as an accidental side-effect.

> Do you know if opening up the host allows for bypassing the OAuth Proxy (with network policies disabled)?

Cryostat's own HTTP port still shouldn't be exposed by any Service, so I don't think so. And, if it does, then unfortunately that was already a problem, because the `JAVA_OPTS_APPEND=-Dquarkus.http.host=0.0.0.0` default value that was previously in the Cryostat Dockerfile was already causing Quarkus to bind on `0.0.0.0` instead of `localhost` anyway :-( the change I made removed that override, so the Operator's setting of `localhost` is now actually taking effect, which is what leads to the probe failure.

> If, so, supposedly it's possible to bind the health checks to a separate host:
https://quarkus.io/guides/management-interface-reference

This seems to talking about the Quarkus `/q` management endpoint in particular, so it would need either that `smallrye-health` Quarkus extension to be added, or else [other source-level changes in Cryostat](https://quarkus.io/guides/management-interface-reference#management-endpoint-application) to work, and needs to have the management interface enabled. I don't think we want to expose the `/q` management endpoint in real deployments, even if it is supposed to be locked down to `localhost` - it would still be possible that a misconfiguration exposes it, or there is some way to bypass the proxy (or just trick the proxy into allowing access to that path) and gain access to the management interface.

ebaron:
> Cryostat's own HTTP port still shouldn't be exposed by any Service

It could still be accessed using the pod IP though, couldn't it (excluding Network Policies)?

ebaron:
Another thing we could do is change the liveness probe port to the auth proxy. I think this is configured to pass through health check requests with no authentication. Unfortunately this means we don't get a view of the Cryostat container's health in complete isolation since it'll depend on the auth proxy container, but might be a worthwhile tradeoff.

andrewazores:
> Another thing we could do is change the liveness probe port to the auth proxy. 

That's an interesting option. Another way could be to change the probe from an `httpGet` to something else, like a command executed within the container:

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command

The Cryostat container has `curl`, so the command could just do the same thing as the `httpGet` probe but from within the container so that it can connect to `localhost`.

Issue #959, opened 2024-10-07
[Request] User configuration options to plug in their own S3 object storage provider


andrewazores:
See also https://github.com/cryostatio/cryostat-operator/issues/167#issuecomment-2035049343_

steve-todorov:
It would be very useful if people could use their own AWS S3 buckets. We are evaluating this in AWS so we could directly use call S3 instead of running a side-container.

andrewazores:
@steve-todorov I have this implemented in a PR on the Cryostat Helm Chart side (https://github.com/cryostatio/cryostat-helm/issues/246) and I intend to follow up with an Operator implementation once that change is reviewed and merged. Those enhancements should land in Cryostat 4.1, which would come out in a few months' time toward the end of this year.

steve-todorov:
Awesome. I'd be interested in testing this out and giving feedback so if you ping me I'll give it a spin. :)

steve-todorov:
Helm chart works perfect -- we have a staging environment so we can easily deploy it. Let me know if there's a ticket I should subscribe to.

PR #1146, opened 2025-08-21
feat(storage): add configuration for external object storage provider

andrewazores:
# Welcome to Cryostat! 👋

Fixes: #959
See https://github.com/cryostatio/cryostat-helm/pull/247

## Description of the change:

## Motivation for the change:

## How to manually test:
1. *Insert steps here...*
2. *...*


api/v1beta2/cryostat_types.go
```diff
@@ -758,6 +762,46 @@ type DatabaseOptions struct {
...
+	Region *string `json:"region,omitempty"`
+	// Whether Cryostat should trust all TLS certificates presented by the external object storage provider. Defaults to false.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="TLS Trust All",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:booleanSwitch"}
+	TLSTrustAll *bool `json:"tlsTrustAll,omitempty"`
+	// The strategy Cryostat will use for storing files' metadata. The default 'tagging' strategy stores all metadata as object Tags.
+	// The 'metadata' strategy stores metadata as object Metadata, which is immutable but allows for more entries than Tags.
+	// The 'bucket' strategy stores metadata as separate files (ex. JSON object maps) in a dedicated bucket,
+	// with prefixes to differentiate the kind of object the metadata belongs to.
+	// +optional
```
ebaron, line 801:
```suggestion
	// +optional
	// +kubebuilder:validation:Enum=tagging;metadata;bucket
	// +operator-sdk:csv:customresourcedefinitions:type=spec,xDescriptors={"urn:alm:descriptor:com.tectonic.ui:select:tagging","urn:alm:descriptor:com.tectonic.ui:select:metadata","urn:alm:descriptor:com.tectonic.ui:select:bucket"}
```

The kubebuilder marker adds server-side validation:
`The Cryostat "cryostat-sample" is invalid: spec.objectStorageOptions.provider.metadataMode: Unsupported value: "pineapple": supported values: "tagging", "metadata", "bucket"`

The operator-sdk marker adds a dropdown in the OpenShift UI:
Screenshot 2025-09-24 at 13-46-44 Create Cryostat · Red Hat OpenShift Service on AWS

internal/controllers/secrets.go
```diff
@@ -145,7 +140,6 @@ func (r *Reconciler) reconcileStorageSecret(ctx context.Context, cr *model.Cryos
 			passwd := r.GenPasswd(32)
 			secret.StringData[storageSecretAccessKey] = accessKey
 			secret.StringData[storageSecretPassKey] = passwd
-			secret.StringData[storageSecretBasicAuthKey] = b64.StdEncoding.EncodeToString(fmt.Appendf(nil, "%s:%s", accessKey, passwd))
```
ebaron, line 148: Do we no longer use basic auth for the SeaweedFS container?

andrewazores: Hmm, I don't quite recall what the original reason for this was.

I think it's because the ACCESS_KEY/SECRET_KEY are what's actually needed for the S3 SDK client, which is how Cryostat interacts with the SeaweedFS storage, so nothing was using this Base64 basic encoded form of those credentials since the S3 client picks those up from environment variables.

The presigned recording transfers where Cryostat instructs the report generator or jfr-datasource to retrieve an archived recording file also shouldn't need these basic auth credentials, because the URL generated by Cryostat is presigned with Cryostat's own credentials and allows the holder to access that one specific resource (recording file) for a limited duration using that identity - so these components actually don't need to bother with anything authn/authz-related when interacting with storage.

andrewazores:
Hmm... having new issues with this PR now when also used together with reports sidecars - the external S3 (Backblaze B2) service is rejecting the presigned requests with 403. Seems to work fine when it's just Cryostat alone talking to the external S3. This might be something that I just missed in previous development/testing, maybe I never tried this combination of configurations.

andrewazores:
I get the same kind of response when I try to configure a local smoketest setup similarly (report sidecar + external S3):

```
2025-09-24 19:38:51,238 ERROR [io.cry.rep.ReportResource] (executor-thread-1) java.io.IOException: Server returned HTTP response code: 403 for URL: https://s3.us-east-005.backblazeb2.com/NBQrozhmX9XPYm9bXlTJ1LdB9v6FE-dQS_NlGVSnSnc=/-deployments-quarkus-run-jar_startup_20250924T193847Z.jfr?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20250924T193850Z&X-Amz-SignedHeaders=host&X-Amz-Credential=005c367ad28c3fa0000000002/20250924/us-east-1/s3/aws4_request&X-Amz-Expires=60&X-Amz-Signature=f14cc002cf5523d94bfe9de682bcdf378ba16d5ede5e0b127d8ebdc9bdf631d1
```

(this from the reports container)

Presigned URLs are obviously quite restricted and locked down to prevent any kind of forgery or other unauthorized accesses, so it's easy to misconfigure something and invalidate things. I wonder if this just normally works with our cryostat-storage container because SeaweedFS might be more lenient on some of the restrictions than the B2 commercial provider is, and there might be something that we've just been doing wrong silently all this time.

andrewazores:
See linked PRs on other components above ^

I think I figured out what was going wrong with the HTTP 403s on presigned URLs - for some reason, even though Cryostat's S3 SDK client is configured to use path-style accesses, the presigned URLs it generated were using the more recent virtual host/subdomain access style. The reports and jfr-datasource implementation of receiving presigned URLs assumed that the URI's "base" (authority) would be static, but when using S3 virtual host access each separate storage bucket has its own distinct subdomain name. So, the presigned URL transformation that was applied to append the path and query onto the base resulted in an invalid URL which was missing any information about which bucket the request object should be located in.

I originally implemented this URI base configuration on the reports and jfr-datasource containers as a (weak) security measure, to ensure that they could not be told to download resources from URLs on unexpected origins. It's more complex to deal with this when using S3 virtual hosts, but that is the higher performance option and may be the only supported option for users using commercial storage providers, so we need to support it for this feature to make sense. Since both reports and jfr-datasource are now placed behind Ingress NetworkPolicies by default, and since jfr-datasource is also hidden with the Cryostat Pod and behind an auth proxy, the restriction on presigned URL base does not really add much security but does break compatibility. So, in the PRs linked above, I have simply removed that restriction and allowed the reports and jfr-datasource to simply retrieve the asset from the presigned URL verbatim, and have Cryostat send that verbatim URL as generated by its S3 SDK client.

In local Podman smoketesting against my Backblaze B2 storage account this now seems to work just fine, with both reports and jfr-datasource successfully able to pull presigned files out. I will need to do some more work on -helm and -operator to clean up some configurations and to ensure that path-style access is not enforced when external storage is configured, but once I do that then I think this PR will be ready to go.

andrewazores:
The equivalent -helm PR is working fine with my B2 account and presigned recordings transfers to both -reports and jfr-datasource. This PR is having some issues with jfr-datasource, but is working fine with B2 and presigned -reports transfers. Digging into why.

andrewazores:
Fixed, it was just a silly mistake with TLS setup when using an external S3 service. I forgot to remove the configuration that overrides the datasource's truststore with one just containing the generated storage cert (which would be applied to the Cryostat instance's managed cryostat-storage container, if any). Removing the override so that the datasource just uses its default system truststore works, so that it's able to trust B2's cert.

I used a Cryostat CR like this to test:

```yaml
kind: Cryostat
metadata:
  name: cryostat-sample
  namespace: cryostat
spec:
  enableCertManager: true
  networkPolicies: {}
  objectStorageOptions:
    provider:
      metadataMode: bucket
      region: us-east-1
      url: https://s3.us-east-005.backblazeb2.com
      usePathStyleAccess: false
    secretName: s3cred
    storageBucketNameOptions:
      archivedRecordings: archivedrecordings
      archivedReports: archivedreports
      eventTemplates: eventtemplates
      heapDumps: heapdumps
      jmcAgentProbeTemplates: probes
      metadata: cryostatmeta
      threadDumps: threaddumps
  reportOptions:
    replicas: 1
    resources: {}
  targetNamespaces:
  - cryostat
  - apps1
```

With an `s3cred` Secret containing my B2 account credentials.

internal/controllers/common/resource_definitions/resource_definitions.go
```diff
@@ -1806,8 +1886,9 @@ func NewStorageContainer(cr *model.CryostatInstance, imageTag string, tls *TLSCo
 	var containerSc *corev1.SecurityContext
 	envs := []corev1.EnvVar{
 		{
-			Name:  "CRYOSTAT_BUCKETS",
-			Value: "archivedrecordings,archivedreports,eventtemplates,probes",
+			Name: "CRYOSTAT_BUCKETS",
+			// TODO use cr.Spec.ObjectStorageOptions.StorageBucketNameOptions if specified, with the list below as default values
+			Value: "archivedrecordings,archivedreports,eventtemplates,probes,heapdumps,threaddumps",
```
ebaron, line 1891: Should this be implemented before merging?

api/v1beta2/cryostat_types.go
```diff
@@ -762,6 +766,83 @@ type DatabaseOptions struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	JMCAgentProbeTemplates *string `json:"jmcAgentProbeTemplates,omitempty"`
+	// The name of the bucket used to store JVM heap dumps.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	HeapDumps *string `json:"heapDumps,omitempty"`
+	// The name of the bucket used to storage JVM thread dumps.
+	// +optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	ThreadDumps *string `json:"threadDumps,omitempty"`
```
ebaron, line 839: Do you think it's worth noting what values these bucket names default to?

ebaron (commented):
I'm glad the fix for the pre-signed URLs wasn't too bad. Just a couple comments below.

andrewazores:
Just noticed that there's a missing cleanup step if you go from a CR without external storage, then edit the same CR to use external storage. The cryostat-storage Deployment gets left running, unused, along with its Service, PVC, etc.

internal/controllers/common/resource_definitions/resource_definitions.go
ebaron: Would it make sense to update the CRD description to mention that the StorageBucketNameOptions only apply to external providers, and remove this TODO?

PR #1205, opened 2025-11-20
feat(agent): Add smart triggers to agent autoconfig

### LINKED ISSUE #1184: [Request] Allow defining smart triggers for target applications
### Describe the feature

See https://github.com/cryostatio/cryostat-agent/issues/227

"This would end up tying in similarly to the various Declarative configurations being implemented in the Operator and Helm Chart right now, and would also be useful for the Operator's Autoconfiguration feature. The user could use the Operator to specify a ConfigMap name, for example, and the Operator would mount that ConfigMap as a volume to the target application, where the Agent would then pick up the Smart Triggers contained within the ConfigMap and load them."

### Anything other information?

_No response_

Josh-Matsuoka:
# Welcome to Cryostat! 👋

Fixes: [#1184](https://github.com/cryostatio/cryostat-operator/issues/1184)

## Description of the change:

Expands the agent autoconfig feature to automatically configure agent smart triggers when the user specifies a list of config map names containing smart triggers. 

Does the following:
- Mounts the contents of each config map name under /tmp/smart-triggers
- Defines the CRYOSTAT_AGENT_SMART_TRIGGER_CONFIG_PATH environment variable to point the agent at /tmp/smart-triggers

## Motivation for the change:

## How to manually test:
1. Build and Deploy this PR on a crc cluster or similar environment
2. make sample_app_agent
3. Create a configMap triggers containing a smart trigger definition (e.g. "[ProcessCpuLoad > 0.2 ; TargetDuration > duration(\"30s\")]~profile" ) in the same namespace as the sample app.
4. Add the following labels to the pod spec in the deployment:

        cryostat.io/name: cryostat-sample
        cryostat.io/namespace: cryostat-operator-system
        cryostat.io/smart-triggers: triggers

5. Check that the smart triggers configMap gets mounted under /tmp/smart-triggers

sh-4.4$ ls /tmp/smart-triggers/
trigger1
sh-4.4$ cat /tmp/smart-triggers/trigger1
"[ProcessCpuLoad > 0.2 ; TargetDuration > duration(\"30s\")]~profile"

6. Check that the Environment variable CRYOSTAT_AGENT_SMART_TRIGGER_CONFIG_PATH is set.


internal/webhooks/agent/pod_defaulter.go
```diff
@@ -163,6 +164,41 @@ func (r *podMutator) Default(ctx context.Context, obj runtime.Object) error {
...
+					},
+				},
+			})
+		}
+
+		// Mount the triggers specified in the pod labels under /tmp/smart-triggers
+		for _, triggerMap := range getSmartTriggersConfigMapNames(pod.Labels) {
+			container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{
+				Name:      "trigger-" + triggerMap,
+				MountPath: "/tmp/smart-triggers",
```
andrewazores: I think this path should be configurable too, just to be safe. It's all unlikely but maybe there is some conflict with something else at this path, or maybe `/tmp` doesn't exist or isn't writable. Best to make sure there's a way to configure around these potential issues.

internal/webhooks/agent/pod_defaulter.go
```diff
@@ -163,6 +164,41 @@ func (r *podMutator) Default(ctx context.Context, obj runtime.Object) error {
 		},
 	})
 
+	if metav1.HasLabel(pod.ObjectMeta, constants.AgentLabelSmartTriggersConfigMaps) {
+		// Mount the Smart Triggers colume
```
andrewazores:
```suggestion
		// Mount the Smart Triggers volume
```

internal/webhooks/agent/pod_defaulter.go
ebaron: Maybe we could mount it under the empty dir volume we create? We'd just have to account for the file size of things we put there and our resource limits we defined for the volume.

internal/webhooks/agent/pod_defaulter.go
```diff
@@ -60,7 +60,7 @@ const (
 	agentInitMemoryLimit        = "64Mi"
 	defaultLogLevel             = "off"
 	defaultJavaOptsVar          = "JAVA_TOOL_OPTIONS"
-	defaultSmartTriggersMount   = "/tmp/smart-triggers"
+	defaultSmartTriggersMount   = "/tmp/cryostat-agent/smart-triggers"
```
andrewazores: Could you pull out the `/tmp/cryostat-agent` path prefix as a separate constant? This is hardcoded in several places and probably should have already been extracted, but now that there are two different purposes for it it should definitely be done.

andrewazores (commented):
Code looks good and this looks like it's working great in a manual test - could some envtests be added?

ebaron (commented):
Looks good to me, just agree with Andrew's comment about adding test cases.

andrewazores (approved):
LGTM, anything else @ebaron?

ebaron (approved):
Looks good to me too

Issue #1068, opened 2025-03-10
[Request] Agent autoconfig Downward API properties


andrewazores:
### Describe the feature

See https://github.com/cryostatio/cryostat-agent/issues/477

PR #1218, opened 2026-01-16
feat(autoconfig): inject Agent discovery augmentations

andrewazores:
# Welcome to Cryostat! 👋

**Disclaimer**: This PR was written with LLM assistance

Fixes: https://github.com/cryostatio/cryostat-operator/issues/1068
See https://github.com/cryostatio/cryostat/issues/635#issuecomment-2332435489
Depends on https://github.com/cryostatio/cryostat-agent/pull/779

## Description of the change:
Adds additional capability to the Agent autoconfig pod mutation webhook so that the Pod's ownership lineage up to the Namespace is determined (following the same algorithm as https://github.com/cryostatio/cryostat/blob/main/src/main/java/io/cryostat/discovery/KubeEndpointSlicesDiscovery.java) and written into a JSON object `hierarchy.json` in a new ConfigMap. The Pod labels are also written into a `metadata.json` in the same ConfigMap. This ConfigMap is then mounted to the autoconfigured Pod so that the Agent can pick up this augmented Discovery information.

Since the ConfigMap is created before the Pod it cannot have an OwnerReference on the Pod initially, but it only makes sense that the Pod's lifecycle defines the ConfigMap's lifecycle too, so a new controller watch is installed that watches for ConfigMaps created by the webhook and attempts to assign the owner reference to the associated Pod once it's ready.

## Motivation for the change:
See cryostatio/cryostat-agent#779

## How to manually test:
1. Check out and build Agent PR and init image, Cryostat server w/ web frontend, and Operator locally, or use prebuilt images:
```bash
$ export OPERATOR_IMG=quay.io/andrewazores/cryostat-operator:4.2.0-k8s-ergo-16
$ export CORE_IMG=quay.io/andrewazores/cryostat:4.2.0-k8s-ergo-5
$ export AGENT_INIT_IMG=quay.io/andrewazores/cryostat-agent-init:0.7.0-k8s-ergo-7
```
2. Build Operator locally, optionally with OCP bundle mode
3. Deploy Operator, create a Cryostat CR, deploy a sample application
4. Edit sample application Pod template metadata labels for Agent autoconfig, or in OCP bundle mode on OpenShift use the Console Plugin UI
5. Verify results, should look like screenshots in cryostatio/cryostat#1249

### DIFF
--- a/internal/controllers/common/resource_definitions/resource_definitions.go
+++ b/internal/controllers/common/resource_definitions/resource_definitions.go
@@ -113,8 +113,11 @@ const (
 	SecretMountPrefix                 string = "/var/run/secrets/operator.cryostat.io"
 )
 
-func createMapCopy(in map[string]string) map[string]string {
-	copy := make(map[string]string)
+func CreateMapCopy(in map[string]string) map[string]string {
+	if in == nil {
+		return make(map[string]string)
+	}
+	copy := make(map[string]string, len(in))
 	for k, v := range in {
 		copy[k] = v
 	}
@@ -126,8 +129,8 @@ func createMetadataCopy(in *operatorv1beta2.ResourceMetadata) operatorv1beta2.Re
 		return operatorv1beta2.ResourceMetadata{}
 	}
 	return operatorv1beta2.ResourceMetadata{
-		Labels:      createMapCopy(in.Labels),
-		Annotations: createMapCopy(in.Annotations),
+		Labels:      CreateMapCopy(in.Labels),
+		Annotations: CreateMapCopy(in.Annotations),
 	}
 }
 
--- a/internal/webhooks/agent/discovery_hierarchy.go
+++ b/internal/webhooks/agent/discovery_hierarchy.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/cryostatio/cryostat-operator/internal/controllers/common/resource_definitions"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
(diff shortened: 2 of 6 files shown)


andrewazores:
> At the very least this can simplify the autoconfig metadata, since the existing metadata can just be mounted as a file in the Pod without needing to marshal the metadata.json and put that into the ConfigMap.

Actually this might not be so simple - the Agent needs it to be JSON because it doesn't have the Jackson module for parsing YAML. Mounting the Pod metadata labels/annotations via Downward API is going to result in YAML files. So it looks like the Operator would need to read the metadata labels/annotations and marshal that to JSON to supply to the Agent.

But then we have a circular problem where we want to read the Pod's metadata YAML, marshal that to JSON, then store that back into the Pod metadata to mount as a file via Downward API. I think we'll end up in an infinite loop here of embedding the updated metadata back into a new update.

Hmm...

Or else we can add the dependency to the Agent so that it can parse YAML, and it should become possible to just use the existing Downward API to mount the Pod metadata as files that the Agent can read. If the hierarchy information is included in the Pod annotations (set there by the Operator webhook) then this would all be in one metadata YAML file for the Agent to process. Do you think that sounds about right @ebaron ?

andrewazores:
Here's another idea - what if the Operator created one ConfigMap per Namespace (or one per NS which contains at least one autoconfig'd Deployment), and then updated that ConfigMap's contents whenever the set of autoconfig'd Pods changed? Then it could mount the ConfigMap to the Pods, but only the particular relevant files within the ConfigMap - so maybe the files would have a naming scheme including the Pod's name to make them unique. The data within the ConfigMap is probably not that large per Pod, so this one ConfigMap itself shouldn't end up being too large unless there are really a lot of replicas or some node in the hierarchy has a lot of metadata (ex. very large annotations).

ebaron:
I was thinking the pod name and namespace could be an extra bit of information sent in the registration data. Cryostat would know that that information corresponds to a pod name and namespace and then should have the information it needs to query the ownership/metadata.

PR #1260, opened 2026-03-11
feat: add enableAudit CR property for audit logging

jtolentino1:
# Welcome to Cryostat! 👋

Fixes: #1257

## Description of the change:
Adds an `enableAudit` boolean property to the Cryostat CR spec that configures the `CRYOSTAT_AUDIT_ENABLED` environment variable on the Cryostat container. New CRs default to `true` via the defaulting webhook; existing CRs created before this field was introduced are left unchanged to preserve upgrade safety.

## Motivation for the change:
Audit logging enables Cryostat to retain historical information about lost target JVMs and their Kubernetes lineages, which is valuable for correlating with other observability signals after unexpected shutdowns. Users should be able to opt in or out through the CR rather than requiring manual env var configuration.

## How to manually test:
1. Deploy the operator and create a new Cryostat custom resource without setting `enableAudit`. Verify the defaulter sets it to `true` and `CRYOSTAT_AUDIT_ENABLED=true` appears in the core container env.
2. Create a custom resource with `enableAudit: false`, `CRYOSTAT_AUDIT_ENABLED=false` should appear.
3. Create a custom resource that omits `enableAudit` on an existing (pre-upgrade) instance. The env var should not exist.

### DIFF
--- a/api/v1beta2/cryostat_types.go
+++ b/api/v1beta2/cryostat_types.go
@@ -78,6 +78,12 @@ type CryostatSpec struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	ReportOptions *ReportConfiguration `json:"reportOptions,omitempty"`
+	// Enable Cryostat audit logging.
+	// New Cryostat CRs default to true, but existing CRs created before this field was
+	// introduced remain unchanged until this property is explicitly set.
+	// +optional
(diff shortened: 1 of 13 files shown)


andrewazores (commented):
LGTM from a reading. I'll give it a manual test drive later today.

ebaron (commented):
This looks really good. I'll give the upgrade path a test, but I expect it to work fine given the way the defaulting webhook is used.

ebaron (approved):
Upgrade works as expected. No change to existing CRs, new CRs receive the default `enableAudit: true` value.

PR #1273, opened 2026-04-13
feat(autoconfig): agent autoconfig from deployment labels

### LINKED ISSUE #1088: [Request] Agent auto-configuration from deployment labels
### Describe the feature

It would more user-friendly if the agent auto-configuration labels could be applied to workload objects like Deployment, Replica Set, Stateful Set, in addition to the existing pod template. This way users could apply the labels using `kubectl  label ...`.

### Anything other information?

_No response_

Josh-Matsuoka:
# Welcome to Cryostat! 👋

Fixes: [#1088](https://github.com/cryostatio/cryostat-operator/issues/1088)

## Description of the change:

Adds a new mutating webhook to intercept deployment creation/updating with cryostat autoconfig labels, propagating them to the Pod Template labels accordingly. From there the existing pod webhook will mutate the pods and perform the agent autoconfig.

## Motivation for the change:

## How to manually test:
1. pull and build this PR
2. deploy the operator to a cluster (make deploy_bundle BUNDLE_IMG=quay.io/jmatsuok/cryostat-operator-bundle:4.2.0-snapshot)
3. Create a Cryostat CR
4. Create an application deployment (I used the quarkus getting-started quickstart)
5. Add the agent autoconfig labels to the created deployment (cryostat.io/name=cryostat-sample cryostat.io/namespace=cryostat)
6. Check the deployment YAML after updating, verify that the autoconfig labels were propagated to the pod template spec
7. Verify that the agent is attached to the application pods


internal/webhooks/agent/deployment_defaulter.go
```diff
@@ -0,0 +1,99 @@
...
+	if !slices.Contains(cr.Status.TargetNamespaces, deployment.Namespace) {
+		return fmt.Errorf("deployment's namespace \"%s\" is not a target namespace of Cryostat \"%s\" in \"%s\"",
+			deployment.Namespace, cr.Name, cr.Namespace)
+	}
+
+	template := &deployment.Spec.Template
+	targetCryostat := getCryostatName(deployment.Labels)
+	targetNamespace := getCryostatNamespace(deployment.Labels)
+	template.Labels[constants.AgentLabelCryostatName] = targetCryostat
+	template.Labels[constants.AgentLabelCryostatNamespace] = targetNamespace
```
ebaron: Would it make sense to pass through the other agent auto-config labels to the pod template? Maybe just filter on labels that begin with [agentLabelPrefix](https://github.com/cryostatio/cryostat-operator/blob/6c56a8653ae47239d5058b358c07cb0369a272d4/internal/controllers/constants/constants.go#L71), and add those to the template?

internal/webhooks/agent/webhooks.go
```diff
@@ -19,6 +19,7 @@ import (
 
 	operatorv1beta2 "github.com/cryostatio/cryostat-operator/api/v1beta2"
 	"github.com/cryostatio/cryostat-operator/internal/controllers/common"
+	appsv1 "k8s.io/api/apps/v1"
```
ebaron, line 22: Maybe rename this file to just webhooks.go since it is no longer pod-specific.

ebaron (commented):
Looks good overall. I like the approach of adding a simple new webhook for the deployment that implicitly triggers the pod webhook. Few comments below.

internal/webhooks/agent/deployment_defaulter.go
andrewazores: I think Elliott's idea here makes sense - instead of having a hardcoded list of the supported labels which this webhook will copy down, it would be nice if it iterated over all of the Deployment's labels and copied down any that begin with the label prefix (ie `cryostat.io/`). That way when support for new labels is added (ex. #1277) the support only needs to be added in one place, not two.

ebaron: Agreed. Feel free to rename `agentLabelPrefix` to `AgentLabelPrefix` to make it visible outside of the constants package.

internal/webhooks/agent/deployment_defaulter.go
```diff
@@ -93,17 +94,11 @@ func (r *deploymentMutator) Default(ctx context.Context, obj runtime.Object) err
...
-	copyLabelIfExists(template, deployment, constants.AgentLabelCallbackPort)
-	copyLabelIfExists(template, deployment, constants.AgentLabelReadOnly)
-	copyLabelIfExists(template, deployment, constants.AgentLabelContainer)
-	copyLabelIfExists(template, deployment, constants.AgentLabelJavaOptionsVar)
-	copyLabelIfExists(template, deployment, constants.AgentLabelHarvesterTemplate)
-	copyLabelIfExists(template, deployment, constants.AgentLabelHarvesterExitMaxAge)
-	copyLabelIfExists(template, deployment, constants.AgentLabelHarvesterExitMaxSize)
-	copyLabelIfExists(template, deployment, constants.AgentLabelSmartTriggersConfigMaps)
+	for label := range deployment.Labels {
+		if strings.HasPrefix(label, "cryostat.io") {
```
andrewazores: https://github.com/cryostatio/cryostat-operator/blob/6c56a8653ae47239d5058b358c07cb0369a272d4/internal/controllers/constants/constants.go#L71

Should use this pre-existing value. As Elliott suggested, just rename it to capitalize the leading `a` to an `A` - this is (unfortunately...) how Go handles visibility.

internal/webhooks/agent/webhook_utils.go
andrewazores, line 1: Should be `webhook_utils`, not `webook` :-)

andrewazores (commented):
I think this is looking good, just needs one file rename and a rebasing. Any other review feedback @ebaron?

internal/webhooks/agent/deployment_defaulter.go
```diff
@@ -0,0 +1,118 @@
...
+	if len(deploymentName) == 0 {
+		deploymentName = deployment.GenerateName
+	}
+	r.log.Info("Configured deployment ", "name", deploymentName, "namespace", deployment.Namespace)
+	return nil
+}
+
+// Pod defaulter will handle setting default values for missing labels
+func copyLabelIfExists(spec *v1.PodTemplateSpec, deployment *appsv1.Deployment, key string) {
+	_, exists := deployment.Labels[key]
```
andrewazores, line 114: I guess one more tiny nitpick is that since we're only calling this function from within a `range deployment.Labels` loop, we can safely assume that the label does exist. But it's fine to be defensive in this function, and it might be useful as a utility elsewhere in the future, so I'm fine with leaving it as-is.

andrewazores:
Tested manually and it's working well.

ebaron (approved):

andrewazores:
@Josh-Matsuoka rebase please

